Issue #885, opened 2020-03-05
Create ET:L on/for Raspberry Pi


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2015-11-06
Redmine Issue: 885, https://dev.etlegacy.com/issues/885

---

Adjust the scripts & code for Raspbian + -rendererGLES- openGL.

---

Related commit(s):
96d8b88985bee87527668de4db4823315c7840e4 - misc: easybuild.sh update for RPI refs 885
 8236b583019a53e5e93d08140336911a4766d10f - misc/rendererGLES: cmake FindGLES added refs 885
 125ee59888daee748771742096244b65b7eaa915 - RPI: update easybuild - let RPI builds no longer u..
 5d703bfde86294ca3e2f5e1f7ae849c9fcfc9473 - RPI: let etl compile with RPI system sdl2 lib refs..

ETLegacyTeam:
From: @morsik (morsik)
Date: 2015-11-07T09:41:49Z

---

Actually, even etlded worked really slow on RPi

ETLegacyTeam:
From: @rafal1137 (RaFaL)
Date: 2015-11-07T21:11:41Z

---

morsik which version of rpi were you trying etl on ?

ETLegacyTeam:
From: @adanski (Dragonji)
Date: 2015-11-08T10:29:00Z

---

As far as ET:L is single threaded I guess it will heavily lag on any ARM. RPi2's got a quad-core Cortex-A7 CPU and single core speed is only 900MHz. I guess it would run fine if it could use the whole power of RPi.

ETLegacyTeam:
From: @IR4T4 (IR4T4)
Date: 2015-12-30T19:00:39Z

---

About performance we'll figure it out. Afaik genuine sytem requirements of vanilla ET are Pentium III 600 MHz, 128 MB RAM. It also depends on used instruction sets and GLES but RPI2 should deal with that. Time will tell.

Current status for raspbian:
```
./easybuild.sh -RPI -j4

```
Should build the current master on RPI. Some libs/features are disabled.

{FIXME}
-sound system isn't found- 
- client doesn't start (fix GLES linking)
- fix bundled libs not compiling (see  easybuild.sh)
- optimize SDL2 build for RPI (our SDL lib/sdl/cmakeList.txt isn't prepared for arm - see architecture in script)
- use armv8, armv7 or armv6 in file names

Interesting links:
[...]

PR #1421, opened 2020-03-23
Raspberry Pi support

techyian:
This work is linked to #885. Required GLES renderer changes and CMake script updates. Also added section to README for install instructions.


cmake/ETLSetupFeatures.cmake
```diff
@@ -37,18 +65,18 @@ if(BUILD_CLIENT)
...
 		find_package(GLES REQUIRED)
 		list(APPEND RENDERER_LIBRARIES ${GLES_LIBRARY})
 		include_directories(SYSTEM ${GLES_INCLUDE_DIR})
 	endif()
 
 	if(NOT BUNDLED_SDL)
-		# specials for ARM - currently set for RPI raspdian/ubuntu mate
-		# system sdl lib isn't latest & sndio lib is required
+		# specials for ARM - currently set for RPI Raspbian/ubuntu mate
+		# system sdl lib isn't latest & sndio lib is required
```
rmarquis, line 76: Is this comment still up-to-date?
It seems the `if(ARM)` condition below isn't required anymore, since both branch are identical now.

src/rendererGLES/tr_image.c
```diff
@@ -793,8 +777,8 @@ static void Upload32(unsigned *data,
 				internalFormat = GL_RGB5;
 			}
 			else
-			{
-				internalFormat = GL_RGB;
+			{
```
rmarquis, line 780: For next commits, beware of trailing characters and tab/space settings. I took care of these for this time.

src/sdl/sdl_glimp.c
```diff
@@ -1034,7 +1042,9 @@ void GLimp_Init(glconfig_t *glConfig, windowContext_t *context)
 	Cvar_Get("r_availableModes", "", CVAR_ROM);
 
 	// Display splash screen
+#ifndef FEATURE_RENDERER_GLES
```
rmarquis: What is the issue here? We might find a way to enable the splash on RPi too.

rmarquis (commented):
2 things:
* identical branches in CMake config
* what is the problem with the splash screen?

cmake/ETLSetupFeatures.cmake
techyian: Well spotted. Doesn't seem to be required. Will update.

src/sdl/sdl_glimp.c
```diff
@@ -973,10 +974,17 @@ void GLimp_Splash(glconfig_t *glConfig)
...
 	// apply image on surface
-	SDL_BlitSurface(splashImage, NULL, SDL_GetWindowSurface(main_window), &dstRect);
-	SDL_UpdateWindowSurface(main_window);
+	blitResult = SDL_BlitSurface(splashImage, NULL, SDL_GetWindowSurface(main_window), &dstRect);
 
-	SDL_FreeSurface(splashImage);
+	if (blitResult < 0)
+	{		
+		Com_Printf(S_COLOR_YELLOW "SDL_BlitSurface failed - %s\n", SDL_GetError());
+		return;
```
rmarquis, line 982: I suppose you need `SDL_FreeSurface(splashImage)` before the return here, but otherwise LGTM!

rmarquis: Actually, might be even better to write something like:
```
if (blitResult >= 0)
{
    SDL_UpdateWindowSurface(main_window);
}
```

so `SDL_FreeSurface` will be called thereafter.

rmarquis:
Merged!

PR #1437, opened 2020-04-24
Slow/fast reload sound is now determined based on the player that's reloading

madsboddum:
Steps to reproduce the problem:

1. Have two players. Player one has Light Weapons >= 2 and player two has Light Weapons < 2.
2. Player one reloads while player two hears it. Player one reloads fast, but player two hears the slow reload sound.
3. The opposite also happens. Player two reloads while player one hears it. Player two reloads slow, but player one hears the fast sound.

Explained simply, the reload sound is based on *your* Light Weapons skill instead of the skill of the player that is actually reloading.

This happens when:
* Spectating
* Playing


rmarquis:
Nice catch! Thank you, merged.

Issue #1276, opened 2020-03-05
Referee commands buggy


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2019-05-04
Redmine Issue: 1276, https://dev.etlegacy.com/issues/1276

---

Some referee commands seem buggy under certain conditions.
Check them all (from the in-game menu).

Instead of working as aexpected, the seem to trigger the *Additional console commands* output from G_refHelp_cmd.

---

Related issue(s): #701

---

Related commit(s):
dc6c301b61020f19910f3fa91ffadc326db550a9 - ui: fix misc votes referenced incorrect vote cast ..

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-06-14T08:15:25Z

---

{TODO}
* bot commands buttons from ref menu don't work (`unknown cmd bot`)
* check help menu formatting (/ref without argument)
* changing time limit doesn't work
* speclock only lock one team if one is empty, not both
* ...

rmarquis:
See existing list in #701.

PR #1441, opened 2020-04-25
Fix being unable to spec lock an empty team

madsboddum:
Spec lock is also no longer removed from empty teams.

### DIFF
--- a/src/game/g_referee.c
+++ b/src/game/g_referee.c
@@ -450,9 +450,9 @@ void G_refSpeclockTeams_cmd(gentity_t *ent, qboolean fLock)
 {
 	char *status;
 
-	// Ensure proper locking
-	G_updateSpecLock(TEAM_AXIS, (TeamCount(-1, TEAM_AXIS)) ? fLock : qfalse);
-	G_updateSpecLock(TEAM_ALLIES, (TeamCount(-1, TEAM_ALLIES)) ? fLock : qfalse);
+	// Both teams are spec locked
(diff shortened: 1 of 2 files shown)


rmarquis:
I am not really familiar with how speclock is used in competition, so bear with me: as far as I can see, speclock is used to prevent spec to look at either team (can see black screen only) unless they're invited to spec one or both team. Team are usually locked in comp as to not allow spec to join.

As I understand it, an empty team means there is nobody to watch, and no spec can join the (now) empty team, which means the speclock is indeed useless and the reason it is removed automatically. In that context, how useful is your change here? What is the use case?

madsboddum:
That's how I understand speclocking as well.

Similar to speclock, the teamlock is removed for empty teams. That means players would be able to join the now empty team.

rmarquis:
You'll excuse me, but I still don't get what is the issue you're trying to solve here. Isn't the current behaviour better?

madsboddum:
I don't know if it's better, I just saw a TODO here that suggested it should be changed to behave this way instead: https://github.com/etlegacy/etlegacy/issues/1276#issuecomment-595500975

rmarquis:
I see. Yes, it's more of a quick, non exhaustive list of things to check. I am not sure what is the exact issue that has been referenced, but it wouldn't be the first time I'd drop some todo after a more careful check.

I'll close this ticket. The feature itself might still have some corner case to fix, but observed behaviour isn't faulty.

PR #1452, opened 2020-05-08
Changed/Added OSX build checks

zedar187:
The previous check in cmake wasn't working, as OSX_NOX86 hasn't been initialized/set. So it has used 32+64 on Catalina, exiting in an error while compiling. I've adjusted Jacker's if statement in the first commit.

The second commit adds an additional check to easybuild.sh. The procedure has to end pretty early, otherwise easybuild would add `-m32` to the compiler flags. This would break cmake's compile tests again.


rmarquis:
Ty!

PR #1459, opened 2020-05-11
Load autoexec_<class>.cfg instead of autoexec_<Class>.cfg

codethief:
This small patch will have ET:L load `autoexec_<class>.cfg` (in lower case) whenever the player changes class, so as to mimic ETPro behavior.


rmarquis:
Thx!

PR #1468, opened 2020-05-13
[client] Hide renderer warnings behind developer cvar (#1407)

### LINKED ISSUE #1407: move warning prints in console to developer mode only
image
Those warning messages are not really helpful to the regular user as they can't fix these anyway. Fixing these is responsibility of the author of the map/mod pk3 which can't be done by the player due to sv_pure.

Therefore, I suggest to show these only with activated `/developer 1`

Exagone313:
Fixes #1407 

`Com_DPrintf` here is the one defined in `qcommon/common.c` which checks for developer cvar.

`CL_RefPrintf` is passed to the renderer which calls it via `Ren_Warning`
```
$ rg CL_RefPrintf src
src/client/cl_main.c
2462: * @brief CL_RefPrintf
2469:static __attribute__ ((format(printf, 2, 3))) void QDECL CL_RefPrintf(int print_level, const char *fmt, ...)
2673:	ri.Printf                  = CL_RefPrintf;
```
```
$ rg Ren_Warning src/renderercommon
(...)
src/renderercommon/tr_common.h
84:#define Ren_Warning(...) ri.Printf(PRINT_WARNING, __VA_ARGS__)
(...)
```

### DIFF
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -2481,7 +2481,7 @@ static __attribute__ ((format(printf, 2, 3))) void QDECL CL_RefPrintf(int print_
 	}
 	else if (print_level == PRINT_WARNING)
 	{
-		Com_Printf(S_COLOR_YELLOW "%s", msg); // yellow
+		Com_DPrintf(S_COLOR_YELLOW "%s", msg); // yellow
 	}
 	else if (print_level == PRINT_DEVELOPER)
 	{


ensiform:
You do know there's a separate developer print flag for ri.Printf right below that one? PRINT_DEVELOPER 

IMHO hiding all renderer warnings behind developer is not the best idea.

Exagone313:
Then I would suggest adding a cvar for displaying these warnings, off by default?

Exagone313:
Let's close this for now.

PR #1469, opened 2020-05-13
WIP game: Decrease ammo when dropping ticking grenade upon getting killed

### LINKED ISSUE #1467: Grenade duping when revived.
When player that primed a grenade gets killed, drops the grenade that will explode and gets revived he will have the same grenade count as if he never dropped it allowing him to use it again and go above the limit a specific class has.

Not sure if this is intended or left on purpose in Legacy due to balancing but it is present in ET since forever or atleast in ETpro 3.2.6. I also couldn't find any relevant information about it here.

It is pretty easy to stumble upon when one of the popular mechanics in ET is to teamkill to revive, which could be annoying to lose your grenade to. It can also be abused to some degree when defending objectives or in general, because the revived player can't die due to shield and can immediately prime another grenade which he should not have if not for this interaction.

I was hesitant to report it here for a long time because it has been in the game for so long, that I'm not sure if I would count it as a bug or a game mechanic at this point, but this could be said about many things that were left unfixed for many years and its not really for me to decide.

Exagone313:
Fixes #1467

Needs testing.

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -472,6 +472,7 @@ void player_die(gentity_t *self, gentity_t *inflictor, gentity_t *attacker, int 
 	if (self->client->ps.grenadeTimeLeft)
 	{
 		vec3_t launchvel, launchspot;
+		weapon_t ammoIndex;
 
 		launchvel[0] = crandom();
 		launchvel[1] = crandom();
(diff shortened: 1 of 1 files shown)


rmarquis:
Pushed in 8373ffeabc2bcc1e97142e2d65ade2e00f88c99a with some modifications:
* check weapon type `WEAPON_TYPE_GRENADE` to differentiate with smoke grenade, which doesn't have ammo count
* use `self->s.weapon` instead of `self->client->ps.weapon`. They're apparently copies of one another, but since `s.weapon` is already used above reusing it improves readability.

PR #1482, opened 2020-05-22
gcc 10 portability

Bonpournickel:
default is -fno-common. To  avoid multiple definitions you need to explicitly initialize conflicting variables.


.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "libs"]
 	path = libs
-	url = git://github.com/etlegacy/etlegacy-libs.git
+	url = https://github.com/Bonpournickel/etlegacy-libs.git
```
isRyven: is this a mistake?

cmake/ETLInstallOmniBot.cmake
```diff
@@ -27,6 +27,6 @@ LEG_DOWNLOAD(
 
 message(STATUS "Adding Omni-Bot to installer scripts")
 install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${MODNAME}/omni-bot/"
-	DESTINATION "${INSTALL_DEFAULT_MODDIR}/${MODNAME}/omni-bot"
-	DIRECTORY_PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ WORLD_READ
+	DESTINATION "/var/lib/etlegacy/omni-bot"
```
rmarquis, line 30: This breaks our standalone shipped package.

src/server/sv_main.c
```diff
@@ -84,7 +84,7 @@ cvar_t *sv_wwwFallbackURL; // URL to send to if an http/ftp fails or is refused
 
 cvar_t *sv_cheats;
 cvar_t *sv_packetloss;
-cvar_t *sv_packetdelay;
+//cvar_t *sv_packetdelay;
```
rmarquis, line 87: Why? What happened to that cvar?

.gitmodules
rmarquis: Instead of patching any project in `libs`, you  need to send your patch upstream (in this case, OpenAL). We'll not maintain any forked code downstream.

src/server/sv_main.c
Bonpournickel: It is declared in src/qcommon/common.c line 131 and conflict with etlegacy src/server/sv_main.c if I remember well.

cmake/ETLInstallOmniBot.cmake
Bonpournickel: Its purpose was to ease the use of omni-bot with a linux dedicated server. /var is rw and /usr is read-only for the server dedicated user in my case. 
An other variable  as ${INSTALL_DEFAULT_MODDIR} would be helpful.

.gitmodules
Bonpournickel: Sorry my mistake.
To make etlegacy-libs openal compatible with gcc-10 openal/Alc/bformatdec.h needs to be patched. It was the only way i figure out to point to it. I cannot use easily openal 32bit on my host. It dragged too much dependencies with it.

cmake/ETLInstallOmniBot.cmake
rmarquis: I am not sure what you are trying to solve here.
The shipped omnibot works fine when it isn't moved outside of `INSTALL_DEFAULT_MODDIR`.
If you want to sue another package that installs OB globally in your system path (with pre-compiled packages, like the `etlegacy-omnibot ` package in Arch Linux AUR), you surely don't need ETL to erase these packages again.

Requesting another variable does not make sense to me.

rmarquis:
> ETL 32bit is playable and doesn't crash any more when opengl2 is called,

That's great and all, but your PR actually breaks the build for _everyone else_. In particular:
* Upgrading OpenAL breaks mingw-w64 on windows
* Upgrading Freetype/OpenAL seems to break msvc on release mode (debug seems fine): See AppVeyor error `LIBCMT.lib(ftol2.obj) : error LNK2005: __ftol2_sse already defined in SDL2.lib(SDL_stdlib.obj)`
* Your "better Linux integration" breaks our automated Linux builds. Nobody knows what you're trying to achieve in the first place.

Some GCC 10 fixes (somewhat based on your changes, but not really) have been pushed in 210e3effd4bf21862da31afef18ea967d0e456bb. Renderer2 code has not been adjusted yet.

rmarquis:
Closing this PR: GCC 10 compatibility has been added in a somewhat cleaner way. OpenAL update to latest version is more problematic (due to mingw and the C to C++ rewrite), but as this isn't directly related to GCC this can be handled later.

PR #1483, opened 2020-05-23
renderer: Allow setting multisample to 8x for increased visual quality

Calinou:
We could go as far as to allow 16×, but it's even slower for little visual benefit (the difference between 4× and 8× isn't that large already). Also, graphics drivers tend to implement 16× MSAA in unusual ways, such as combining 8× MSAA with 2× SSAA. This could cause various issues if not accounted for in the renderer.


rmarquis:
Merged in 8d2f4d29605042282adbcf2ad2d53070659d8cb7. Thanks!

PR #1511, opened 2020-09-21
Improve reinforcement timer colors with a yellow-red gradient

Calinou:
This changes the number's color progressively towards red when the deployment is close. This helps bring the player's attention to the respawn counter.

PS: Do I need to update the PO template myself, or is it done automatically/by a core developer on a reguler basis?


Aranud:
Work fine ! Thank you !

PR #1514, opened 2020-10-09
Add instructions for a 64-bit easybuild on Linux.

jeannekamikaze:
The OSX build instructions mention the `-64` flag, but the Linux ones don't.


Aranud:
Indeed ! Thanks you !

PR #1519, opened 2020-11-02
Install mod description.txt as non-executable

pemensik:
There is nothing to execute inside it, make sure it stays just readable.


ensiform:
It's pointless. All mods ship with this file anyway and the game never does anything but read it.

pemensik:
It is not pointless. It had to be fixed by my [rpmfusion package](https://bugzilla.rpmfusion.org/show_bug.cgi?id=5824), because automated tools scan all installed executable files. Executable file without #!/bin/sh fails whole build.

It is small change, but that file should not be executable. It is made executable during install for no reason. It is correctly without executable bit in source repository and should stay it after install.

ensiform:
Understood. Didn't realize it was a mistake on that part.

rmarquis:
Thanks, merged!

Issue #1535, opened 2020-12-28
2560x1440 Resolution 'ResampleTexture: max width' error


Raisum:
Hi, I'm having a problem of trying to launch ETLegacy in 1440p resolution. I get an error: recursive error  'ResampleTexture: max width' after ResampleTexture: max width.

I'm running it on Windows 10 with a Vega 64 GPU.

<details>
  <summary>Here's the console log</summary>
ET Legacy v2.76 win-x86 Jan  6 2019
Zone megs: 64
Info: fs_game now defaults to 'legacy' mod instead of 'etmain'
----- FS_Startup -----
Current search path:
    C:\Users\Raisum\Documents\ETLegacy\legacy
    C:\Users\Raisum\Documents\ETLegacy\legacy\pak3_v2.76-794-g7c95af1.pk3 (1164 files)
    C:\Users\Raisum\Documents\ETLegacy\legacy\etl_bin_v2.76-794-g7c95af1.pk3 (8 files)
    .\legacy
    .\legacy\pak3_v2.76.pk3 (1107 files)
    .\legacy\etl_bin_v2.76.pk3 (8 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain
    C:\Users\Raisum\Documents\ETLegacy\etmain\venice_ne4.pk3 (325 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\UJE_VIP_b3.pk3 (242 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\trmfght_beta2.pk3 (197 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\tc_base.pk3 (64 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\supply.pk3 (44 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\sp_delivery_te.pk3 (51 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\snatch3.pk3 (34 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\school.pk3 (115 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\saberpeak_final.pk3 (243 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\praetoria_m2.pk3 (560 files)
    C:\Users\Raisum\Documents\ETLegacy\etmain\pirates.pk3 (154 files)
[...]

PR #1539, opened 2020-12-29
renderer: fix "ResampleTexture: max width" occurs with high resolution.

JumptoHistory:
Fixed an error occurs with higher resolutions.


rmarquis:
See #1535 for a discussion about a better fix. Closing now.

PR #1540, opened 2020-12-30
cgame: revert fireteam health color logic to vanilla

mittermichal:
This is reverting a1f7b8c0efbc52ea49375fcff238b5bf98c8d23e commit, whose message doesn't mention it's changing health color to the same way its used for crosshair health indication.
That intruduced issue that its much harder to tell between low HP alive and dead teammates, because both of theirs hp color is red.
I also removed unnecessary trap_R_SetColor call which is called at the end of CG_Text_Paint_Ext

### DIFF
--- a/src/cgame/cg_fireteamoverlay.c
+++ b/src/cgame/cg_fireteamoverlay.c
@@ -396,7 +396,6 @@ void CG_DrawFireTeamOverlay(rectDef_t *rect)
 	fireteamData_t *f  = NULL;
 	char           *locStr[MAX_FIRETEAM_MEMBERS];
 	int            curWeap;
-	vec4_t         hcolor;
 
 	// assign fireteam data, and early out if not on one
 	if (!(f = CG_IsOnFireteam(cg.clientNum)))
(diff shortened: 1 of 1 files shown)


mittermichal:
Actually a1f7b8c is also broken, because CG_ColorForHealth only uses playerstate health so when playerhealth<100 all fireteam healths are colored same: 
2020-12-30-191354-frostbite

mittermichal:
@isRyven suggested on discord to calculate color in separate function. There is `void CG_GetColorForHealth(int health, vec4_t hcolor)` https://github.com/etlegacy/etlegacy/blob/81fe3bc079f76d8290016f11a55fc2474ee7169b/src/cgame/cg_drawtools.c#L638
which is not used anywhere. Should I rewrite it and use it?

Using some different color logic could be done later, but should be discussed as separate issue I guess.

isRyven:
Yeah, didn't know something like that already exists, when i suggested implementation from different codebase.

mittermichal:
Is there something more to be done to get this merged?

Aranud:
Work fine, merged !

rmarquis:
Ah, I actually thought you wanted to do some more work before merging. Maybe open a ticket for the colour logic aspect, if appropriate.

mittermichal:
yeah, will do. Just wanted to play with vanilla logic until its properly done. :)

Issue #1564, opened 2021-01-21
Add option to set player position and pov


realkemon:
For comparing performance with `/r_speeds 1` it would help if we could save a certain player position including pov so we can be sure that the metrics should be identical. It would also help with taking before/after comparison shots for map overhauls. I think etjump saves the position, not sure about pov?
Saving the position is only one side though as it would need to be necessary to also set the position on a different map, possibly different game instance. 

Maybe sth like `/playerPosition` that prints the current position and pov in the console if it's called without arguments and sets the player to that position if it's called with arguments. Is that possible?

Aciz:
Note: there is already `viewpos` for printing out X Y Z Yaw (lacking pitch and roll), and `freecamgetpos` & `freecamsetpos` which unfortunately are only available in demo playback in freecam mode.

Aciz:
...and theres aleady also `setviewpos`... But this command along with `viewpos` should probably be extended to include pitch & roll.

realkemon:
They could've gone for `viewPos` and `setPos`, but no....
But yes, those two sound perfect. Roll is probably unnecessary, but I guess it  wouldn't hurt either.

PR #1565, opened 2021-01-21
Extend viewpos and setviewpos to include all angles

Aciz:
Extends `viewpos` and `setviewpos` commands to take 6 arguments instead of 4, so you can individually view and set all your view angles. The first argument is now pitch instead of yaw.

refs #1564

### DIFF
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -41,9 +41,9 @@
  */
 static void CG_Viewpos_f(void)
 {
-	CG_Printf("(%i %i %i) : %i\n", (int)cg.refdef.vieworg[0],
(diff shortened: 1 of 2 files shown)


rmarquis:
LGTM, but I wonder if backward compatibility is necessary here. Are these commands used in scripts, for example?
If so, I would make the extra angles optional, so the command either accept position or position+angle parameters.

Aciz:
I actually just noticed that some bot waypointing stuff uses `setviewpos` command and probably breaks now, will handle once I figure out what exactly it is. As for other backwards compatiblity, I'm not sure if legacy supports freecam scripts, but those don't use these commands, they use `setfreecampos` to set position + angles.

Aciz:
Okay so now `setviewpos` accepts both the old format `(X Y Z Yaw)` for backwards compatibility as well as new format for setting all angles `(X Y Z Pitch Yaw Roll)`

PR #1569, opened 2021-02-01
game: Don't show health of invulnerable entities, fixes #1563

### LINKED ISSUE #1563: Hide vehicle healthbar when setdamagable 0
In the new version of 'Supply Depot, by ginc' I added all necessary entities to allow server admins to toggle the truck between vulnerable and invulnerable. The only downside of that addition is that for the case where the truck remains invulnerable, there is now a healthbar displayed. In that case it can be considered as intrusive clutter, because the healthbar yields no benefit for the player.

Can the healthbar of script_movers be hidden when said script_mover is set to `setdamagable [script_mover] 0`? This is not a one time setting, but can change throughout the map if that is important for the implementation.

M-RBDR:
game: changed G_ScriptAction_SetDamagable so as to not show health of invulnerable entities, fixes #1563

### DIFF
--- a/src/game/g_script_actions.c
+++ b/src/game/g_script_actions.c
@@ -3969,6 +3969,12 @@ qboolean G_ScriptAction_SetDamagable(gentity_t *ent, char *params)
 	while ((target = G_FindByTargetname(target, name)))
 	{
 		target->takedamage = canDamage;
+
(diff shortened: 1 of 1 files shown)


rmarquis:
Thx!

Issue #1095, opened 2020-03-05
Slicks are bugged on 32bit build


ETLegacyTeam:
Author: @isRyven (ryven)
Date: 2018-06-16
Redmine Issue: 1095, https://dev.etlegacy.com/issues/1095

---

Slick surface merged together with solid brush, ie. solid brush + slick brush on top of each other(slick brushes are also solid btw), breaks slick functionality on 45 degree angled plane, this isn't reproducible on 64bit build, it only happens on 32bit linux builds, windows builds don't get bugged. 

The bug in action, on real map, 45 degree slope just becomes non-slick:
https://streamable.com/jus8p

Now, this obviously doesn't happen with 2.60b, BUT the same bug appears  also in et300, which could mean, that any GPL based engine modification should inherit the very same problem!
Makes me wonder if compilation flags are sort of messed up? cmake shows that fast-math is enabled, so i don't know.
I didn't test master builds(2.77), since i don't have linux installed, capable to run 32bit builds atm, so i hope someone can test that for me, but the bug should be reproducible on 2.76.

!https://media.discordapp.net/attachments/353604235926962179/457545574724665355/unknown.png(45icetest)!
The attached map to test bug on:
* right slope - slick + brush on top of each other, 
* center - just slick, 
* left slope - brush + slick, that was moved 1 unit away, so they don't overlap together. 

The right one should bug, the rest are fine.

Aciz:
This issue seems to have resolved itself at some point, I can no longer reproduce this on a server running 32-bit 2.76 release on Linux.

PR #1571, opened 2021-02-04
qcommon: unhardcode patch plane optimization

Aciz:
Unhardcode the patch optimization revert done in [this commit](https://github.com/etlegacy/etlegacy/commit/abd2d4d910a6376c3edc79b20db1b687c677c369?branch=abd2d4d910a6376c3edc79b20db1b687c677c369) and tie the optimization to `cm_optimizePatchPlanes` cvar. This is mainly meant for trickjump servers, where some maps require usage of this bug in order to be completed.

### DIFF
--- a/src/qcommon/cm_load.c
+++ b/src/qcommon/cm_load.c
@@ -53,6 +53,7 @@ cvar_t *cm_noAreas;
 cvar_t *cm_noCurves;
 cvar_t *cm_playerCurveClip;
 cvar_t *cm_optimize;
+cvar_t *cm_optimizePatchPlanes;
 
 cmodel_t box_model;
 cplane_t *box_planes;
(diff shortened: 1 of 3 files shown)


Aciz:
Note to self: don't code while you're tired....

The cvar logic is actually reversed here, it's the optimization in VET that causes the bug, so the cvar should default to 0 and that be the default, unoptimized behavior that fixes the original issue. Will fix later today.

rmarquis:
Thx! With #1095 now out of the way, I think that was the last missing piece to have full compatibility with etjump.

Aciz:
I do hope so, I have started running my own private server with eltded in order to possibly find more bugs. Once first snapshot is out, I'll see if I can persuade some server owners to swap for etlded too, for more widespread testing 👍

isRyven:
There is one long standing bug that happens on the gamma mill map, `MAX_PATCH_PLANES reached`, seems to be there is no ticket for this one yet.

image

Issue #1414, opened 2020-03-19
set spawntimer period by game


mittermichal:
to use spawntimer one has to first set its period with timerSet command
It is possible to set it with `autoexec_<team>.cfg` and `autoexec_<map>.cfg` scripts. 
Some use script that would cycle through common values like 15, 20, 25, 30
In etpro some use openTimerInput command that lets player input the period by input dialog

I propose to set this period by game and let player only to decide when to reset the timer
the spawn times are already parsed by client from config string in CG_ParseServerinfo and are accesible by g_red(blue)limbotime cvars
I personally use bind KP_END "g_redlimbotime;g_bluelimbotime" to get the spawn periods

PR #1572, opened 2021-02-06
cgame: draw limbo times instead of warmup in reinforcement time HUD element

mittermichal:
During warmup the reinforcement time HUD element only displays blinking WARMUP string, this information is redundant because its already displayed in the middle of screen:
image

So instead of WARMUP display limbo times of both teams with following logic:
- If player or following player is axis => red allies limbo time and blue axis limbo time
- For allies or in free cam => red axis limbo time and blue allies limbo time
- If the game type is Last Man Standing => WARMUP

I also sneaked in change to not blink local time for alternative timers.
I kept the blinking for the limbo times, can't decide whether that should be disabled too.

Preview:
- Normal timer:
image

- Alternative timers position (cg_althudFlags & 1) + non-blinking local time:
image

refs #1414

### DIFF
--- a/src/cgame/cg_draw_hud.c
+++ b/src/cgame/cg_draw_hud.c
@@ -2267,18 +2267,18 @@ static const char *CG_TimerWarmupString()
 	}
 	else                                              // draw limbotime periods otherwise 
 	{
-		int limbotime_own, limbotime_enemy;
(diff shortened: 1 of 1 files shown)


Aranud:
Work fine, merged !

PR #1573, opened 2021-02-07
Auto spawn timer period refs: #1414

mittermichal:
Use spawnTimer period from `cg_red(blue)limbotime`, this relieves burden of setting the value manually or with `autoexec_<team>.cfg` and `autoexec_<mapname>.cfg` scripts.
Player now only has to execute `timerReset` or `resetTimer`.

timerSet behaviour:
- `timerSet <number>` - custom period
- `timerSet` - resets the stamp which disables the spawn timer
- `timerSet 0` to use the cg_red(blue)limbotime again - this is changed in this PR, behavior before would be same as `timerSet`

I extracted the logic of drawing spawn timer text into function that is used by both normal and alternative timer.

Note that `cycle cg_spawnTimer_set 0 60000 1000` and `cycle cg_spawnTimer_set 60000 0 1000` can be used to adjust the timer forward or backward in time. Could be added as bind in advanced controls menu? Or at least as a hint.

If spawn time periods are changed on server then the spawn timer isn't accurate anymore, but I think it could done too. 

This pull request contains also fix for timerReset/resetTimer command.
Also contains invalidation of spawn timer stamp when cgs.levelStartTime is changed - this fixes spawn timer being broken on map reset or map change.

Including preview because I did some adjustments when adding space character between spawn timer and reinforcement timer:

- before `timerReset` + positive timelimit + alternative timer:
2021-02-07-174914-frostbite

- before `timerReset` + positive timelimit + normal timer:
2021-02-07-174919-frostbite

- after `timerReset` + positive timelimit + alternative timer:
2021-02-07-174941-frostbite

- after `timerReset` + positive timelimit + normal timer:
2021-02-07-175011-frostbite

- before `timerReset` + non-positive timelimit + normal timer:
2021-02-07-175028-frostbite

- before `timerReset` + non-positive timelimit + alternative timer:
2021-02-07-175033-frostbite

- after `timerReset` + non-positive timelimit + normal timer:
2021-02-07-175114-frostbite

- after `timerReset` + non-positive timelimit + alternative timer:
2021-02-07-175046-frostbite

refs: #1414

### DIFF
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -1274,16 +1274,13 @@ static void CG_TimerSet_f(void)
  */
 static void CG_TimerReset_f(void)
 {
-	int msec;
-
 	if (cgs.gamestate != GS_PLAYING)
 	{
 		CG_Printf("You may only use this command during the match.\n");
 		return;
 	}
 
-	msec = (int)(cgs.timelimit * 60000.f) - (cg.time - cgs.levelStartTime); // 60.f * 1000.f
-	trap_Cvar_Set("cg_spawnTimer_set", va("%d", msec / 1000));
+	trap_Cvar_Set("cg_spawnTimer_set", va("%d", cg.time - cgs.levelStartTime));
 }
 
 /**
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -1251,7 +1251,7 @@ static void CG_TimerSet_f(void)
 
 		if (spawnPeriod == 0)
 		{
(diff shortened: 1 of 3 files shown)


Aranud:
Work fine, merged !

BystryPL:
It's bugged, sometimes the spawntimes are incorrect when using resetTimer command.

Aciz:
From what I understood, you need to use `timerSet` once before using `resetTimer` after a map has changed, to load the current active spawntimes on the map? I've never experienced issues with this by using it this way (although my experience is pretty limited)

mittermichal:
@Aciz no you don't have to. You only need to use `resetTimer` or `timerReset`

Issue #1415, opened 2020-03-20
show cvar name and bind command in tooltips


mittermichal:
If you customize game through menu, you don't see the cvar you are setting or command you assign to key. If you want to do it in console or config you have to know them.
I used to look for them in *.menu files.

![image](https://user-images.githubusercontent.com/4035800/77152392-2f569c80-6a98-11ea-9cb1-d666ae73c2b6.png)

mittermichal:
played around with it here: https://github.com/mittermichal/etlegacy/commit/fc312dd582ca20a0efeb783751449fc0b514601a
Should I make pull request in case someone wants to improve it?

PR #1577, opened 2021-02-08
ui: Add cvar name tooltip refs: #1415

mittermichal:
Adds grey `cmd:` / `bind:` `<cvarname>` at the end of tooltip text.
Preview:
ui_r
not sure what to do with the ui_ prefix here. Remove it or color it differently?

bind_cmd
bind command

profile

maybe it would be better to not show in profile creation since its the 1st thing new player sees?

refs: #1415


rmarquis:
There is a few minor issues:

```
In file included from ..\etlegacy\src\ui\ui_shared.h:39,
                 from ..\etlegacy\src\ui\ui_parse.c:36:
..\etlegacy\src\ui\ui_parse.c: In function 'ItemParse_tooltip':
../etlegacy/src/qcommon/q_shared.h:1933:16: warning: initialization of 'const char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  #define __(x) CG_TranslateString(x)
                ^~~~~~~~~~~~~~~~~~
..\etlegacy\src\ui\ui_parse.c:2110:37: note: in expansion of macro '__'
  const char *translatedParsedText = __(token.string);
                                     ^~
```

Somehow, the `__()` macro looks for `CG_TranslateString` rather than `UI_TranslateString`. The cast issue might be related or independent.

Also, in case the cvar requires some vid_restart/snd_restart, the cvar displayed in tooltips is the UI cvar used as proxy (f.e. `ui_r_intensity` while the user would rather manually modify `r_intensity` instead). Not really an issue though, just something to be aware off.

I otherwise *really* like the end result. Well done here.

mittermichal:
I tried to fix the warning here https://github.com/etlegacy/etlegacy/pull/1577/commits/a4730c59b0d39b1e1cdbc7a9c2fac189dabe1ef9 but it fails to build for windows:
```
ui_parse.obj : error LNK2019: unresolved external symbol _UI_TranslateString referenced in function _ItemParse_tooltip [C:\projects\etlegacy\build\cgame.vcxproj]
C:\projects\etlegacy\build\legacy\cgame_mp_x86.dll : fatal error LNK1120: 1 unresolved externals [C:\projects\etlegacy\build\cgame.vcxproj]
```
https://ci.appveyor.com/project/rmarquis/etlegacy/build/4298

I don't know what to do with this, hopefully @Aranud can fix this :D

mittermichal:
- in certain menus like profile creation the cvar name is not displayed. 
- ui_ prefix common in system settings is removed

Some cvars have their text too long and with cvar name it overflows the screen so I split the text with new line. Unfortunately if tooltip has multiple lines, the bottom overflows a bit.
shot0001

shot0000

mittermichal:
build should be fixed by using `DC->translateString()` instead of `__()`
thanks ensiform for tip

mittermichal:
changed status from WIP to READY :sunglasses:

PR #1580, opened 2021-02-11
cgame: fix spawn timer disappearing on pause

mittermichal:
This is reverting invalidation of spawn timer stamp when cgs.levelStartTime is changed.
I thought levelStartTime changes only on map restart or map change, but it changes every 500 ms when match is paused too. This is vanilla behavior which was marked as FIXME in G_RunFrame.
So now timer is enabled once its been set, even after map change, which is not a big deal.
I don't have any idea how to detect map restart or change, but one ugly way around this could be to check whether the levelStartTime change is exactly 500.

### DIFF
--- a/src/cgame/cg_servercmds.c
+++ b/src/cgame/cg_servercmds.c
@@ -961,7 +961,6 @@ static void CG_ConfigStringModified(void)
 		break;
 	case CS_LEVEL_START_TIME:
 		cgs.levelStartTime = atoi(CG_ConfigString(num));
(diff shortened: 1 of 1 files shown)


mittermichal:
its reverting this: https://github.com/etlegacy/etlegacy/pull/1573/commits/81cbd774d72789837a65c2efec0f6b722dcc5b36#diff-6a3b4eeea34e5b40f283961caef85c6143ed00a0a9c3202ece626166b9fe074cR964

PR #1593, opened 2021-03-01
Add PowerPC64 Little Endian ARCH_STRING

runlevel5:
When I run my game on Linux PPC64LE, the game refuses to start with a warning that failing to find libs with suffix `.ppc64.so`.

The ARCH_STRING of PowerPC64 LE should be `ppc64le` instead of `ppc64`

Good to see that the game now run ok on Linux ppc64le :)

Screenshot from 2021-03-01 21-36-30
Screenshot from 2021-03-01 23-44-30

### DIFF
--- a/src/qcommon/q_platform.h
+++ b/src/qcommon/q_platform.h
@@ -203,7 +203,11 @@
 #define idx64 1
 #define ARCH_STRING "x86_64"
 #elif defined __powerpc64__
+#if BYTE_ORDER == BIG_ENDIAN
 #define ARCH_STRING "ppc64"
+#else
+#define ARCH_STRING "ppc64le"
+#endif
 #elif defined __powerpc__
 #define ARCH_STRING "ppc"
 #elif defined __s390__


rmarquis:
Closing. We'll most likely never add PPC support back nor distribute these binaries, so there is no point in adding this.

runlevel5:
Thanks for @rmarquis review. Let's hope one day ET Legacy would support PPC

rmarquis:
Some additional information: I understand you might be disappointed here, but adding support for a new arch requires the PPC binaries to be shipped in the official .pk3, which would then be installed on every server. It's much more than adding a few lines of code, but ensuring the package compiles on the platform and automating the compilation (like we do with the provided Windows/Linux/MacIntel binaries).

This makes very little sense for PPC, since that arch has basically been replaced for the past 15 years, maybe apart in very exotic scenario.

runlevel5:
I hope one day PPC would gain back its share of the market then we could revisit this PR.

> This makes very little sense for PPC, since that arch has basically been replaced for the past 15 years, maybe apart in very exotic scenario.

Not entirely true, yes we all know Apple has moved away from PPC since the G5 but the PPC is still living on. Blackbird or Talos II workstations are some choices if users want to have a desktop Linux running POWER9 CPU.

aryoda:
I also would like to see ppc64le support for at least compiling it myself without thinking about required/missing patches.

Perhaps modfiying the CI/CD pipeline to exclude certain stages could be way (even though it causes work to be done at your side).

Regarding CI/CD for ppc64le: This is now supported at gitlab:

https://about.gitlab.com/releases/2021/09/22/gitlab-14-3-released/#gitlab-runner-on-ibm-power9-linux-os

CloudyProton:
ET Legacy does not build without this patch. I would also like to see ppc64le support revisited.

madscientist159:
That's a real shame, my daily driver is a POWER9 Talos II so I'd have to patch and compile before I could play.

Aciz:
Even if you compile it yourself, you won't be able to basically do anything. Mods need to support the architecture, therefore you won't be able to play on any server, and you cannot even play with bots locally, as omni-bots have no support for PPC either. So unless support is officially added (which it very likely won't be because that's yet another niche arch to maintain, we barely have contributors maintaining and testing other arches besides linux/windows), there really isn't even point in doing it yourself.

runlevel5:
> Even if you compile it yourself, you won't be able to basically do anything. Mods need to support the architecture, therefore you won't be able to play on any server, and you cannot even play with bots locally, as omni-bots have no support for PPC either. So unless support is officially added (which it very likely won't be because that's yet another niche arch to maintain, we barely have contributors maintaining and testing other arches besides linux/windows), there really isn't even point in doing it yourself.

I am fully aware of such limitations. However if the team accepts my PR, this would open up opportunities for the PPC community to contribute to the omni-bots and related ecosystems. Just my 5cent

runlevel5:
@Aciz 

>  and you cannot even play with bots locally, as omni-bots have no support for PPC either.

it is actually possible. The omnibot can be compiled and worked nicely.

PR #1601, opened 2021-03-01
Remove icon referernce from metainfo.xml

Eonfge:
The Icon-tag is not part of the metainfo spec, and a strict parser will fail on it.


rmarquis:
The release tag should probably be added through a cmake template. There is no point in doing it manually over and over again.

Eonfge:
Depends, how many releases are you planning in the coming month?

jackeri:
This should definately be a cmake configurable file, _no_one_ will remember nor should need to update files for tag versions.

PR #1608, opened 2021-03-03
client: fix compatibility with ETJump shared cvar

Aciz:
Temporary fix to allow ETJump `shared` cvar to be modifed on client side by server. `shared` contains information about various settings set on a map level via worldspawn keys, and is sent from server to client upon loading a map to allow client to predict these settings. This can probably be reverted at some point, if adding `CVAR_SYSTEMINFO` to cgame side of this cvar in ETJump is okay and doesn't cause any issues.

Appearently there's also an issue here where `CVAR_SERVER_CREATED` gets dropped along the way, hence the TODO comment.

### DIFF
--- a/src/client/cl_parse.c
+++ b/src/client/cl_parse.c
@@ -283,7 +283,7 @@ void CL_ParsePacketEntities(msg_t *msg, clSnapshot_t *oldframe, clSnapshot_t *ne
 		else
 		{
 			oldstate = &cl.parseEntities[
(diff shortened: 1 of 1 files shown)


Aciz:
`shared` cvar has been now fixed in ETJump (https://github.com/etjump/etjump/commit/eb8b56005b9678795ef036a9944b583a67891754) so once a new version of ETJump is released, this compatibility fix can be removed.

isRyven:
unless we want to support older versions

Aciz:
> unless we want to support older versions

we don't

PR #1629, opened 2021-04-01
game: Fixing pmove_fixed integration with antiwarp

ryzyk-krzysiek:
In the current master a game with `g_antiwarp` and `pmove_fixed` enabled means that `pmove_fixed` doesn't work at all, this is an attempt to fix it. From my tests the fix works and I haven't encountered any problems with physics.

@isRyven Maybe you can take a look?

### DIFF
--- a/src/game/et-antiwarp.c
+++ b/src/game/et-antiwarp.c
@@ -187,9 +187,10 @@ void DoClientThinks(gentity_t *ent)
 		serverTime = cmd->serverTime;
 		totalDelta = latestTime - cmd->serverTime;
 
-		if (ent->client->pers.pmoveFixed)
+		if (ent->client->pers.pmoveFixed || pmove_fixed.integer)
 		{
 			serverTime = ((serverTime + pmove_msec.integer - 1) / pmove_msec.integer) * pmove_msec.integer;
+			deltahax   = qtrue;
 		}
 
 		timeDelta = serverTime - lastTime;


isRyven:
It didn't work in the first place because players don't transmit pmove fixed state to the server as you would normally encounter in other mods (incl. etjump), hence it can only be forced by the server, but doing so would mess up the spread globally (disabling it on client side wouldn't have any effect, probably would just screw up the prediction). I wanted to enable per client pmove fixed, but i couldn't really find the reason to do it for the legacy, it's better to invest time and effort in making fixed physics working, and enforcing this for everyone, well, as it already basically is. Pmove fixed doesn't really bring any value to the legacy. And this fix only would work if pmove fixed is enabled on the server side, if that was the implication, then alright.

ryzyk-krzysiek:
I'm trying to figure out the whole pmove stuff and physics and encountered this and tried to fix it, which gave me a bit better understanding of things I think. I realize that it doesn't matter much since noone really uses pmove_fixed in legacy I think. But not sure right now about the fact that it would mess spread for everyone with this change...

Anyway what about applying the same for your patch? I'm trying to test it right now, but not sure as I just started. Basically: 
```
if (g_fixedphysics.integer)
		{
			int msec = 1000 / g_fixedphysicsfps.integer;
			serverTime = ((serverTime + msec - 1) / msec) * msec;
			deltahax = qtrue;
		}
		else if (ent->client->pers.pmoveFixed || pmove_fixed.integer)
		{
			serverTime = ((serverTime + pmove_msec.integer - 1) / pmove_msec.integer) * pmove_msec.integer;
                        deltahax = qtrue;
		}
```

Edit: nevermind, there are some problems with the code above (g_fixedphysics)
Edit2: oops I was trying to test 30 fps fixedphysics but it was capped to 60fps that's why I had problem

isRyven:
> But not sure right now about the fact that it would mess spread for everyone with this change...

I meant if the server would enable `pmove_fixed`, there would be no way to disable it on the client side, similar to fixed physics, hence it would be enforced for everyone, as a result everyone would suffer from bigger spread scale issue. But if you want to make pmove_fixed working, it's better then first make it client togglable (which would populate the session field). This is usually done by transmitting it through `cg_uinfo` cvar, you can follow the trail if you are interested: https://github.com/etjump/etjump/blob/master/src/cgame/cg_main.cpp#L1206

About the patch, i tried to do similar thing, but i didn't really get the right results, but maybe you will succeed, more over, i would suggest to also enable the `cg_showmiss` (it can be configured further with bitmask, just check the cg_predict.c code) when you testing the things.

ryzyk-krzysiek:
> I meant if the server would enable `pmove_fixed`, there would be no way to disable it on the client side, similar to fixed physics, hence it would be enforced for everyone, as a result everyone would suffer from bigger spread scale issue. But if you want to make pmove_fixed working, it's better then first make it client togglable (which would populate the session field). This is usually done by transmitting it through `cg_uinfo` cvar, you can follow the trail if you are interested: https://github.com/etjump/etjump/blob/master/src/cgame/cg_main.cpp#L1206

Yes, I meant that players can't control it. I didn't really thought about it much, but I guess it should be made like other mods. Thanks for further clarification, I will take a look if I won't manage to make your fixed physics patch work with antiwarp, since it would make pmove_fixed obsolete anyway I think, but on the other hand fixed physics is server side only too, whatever :p.

> About the patch, i tried to do similar thing, but i didn't really get the right results, but maybe you will succeed, more over, i would suggest to also enable the `cg_showmiss` (it can be configured further with bitmask, just check the cg_predict.c code) when you testing the things.

Will do for sure, I think I tried it before but it didin't work or something, I know I tried it on ETPro. Thanks!

PR #1635, opened 2021-04-09
mod: fix adrenaline animations, fixes #1634

### LINKED ISSUE #1634: Adrenaline animation/usage is broken
Using adrenaline in legacy is broken in several ways.

* There is no usage animation, the syringe will just disappears. In etmain, you see yourself injecting the syringe.
* After using, weapon is autoswitched (though I think this is made intentionally looking at previous commits)
* Swapping back to adrenaline once it's been used seemingly tries to play the injection animation again

https://streamable.com/jq22tm

Aciz:
Fixes adrenaline not playing animation, and getting stuck for 1500ms on a non-existing animation after switching back to it. Apparently all weapons go through a reload now, so just setting adrenaline reload time to 0 seem to be a correct way to handle it, same is done for syringe (both have 1500ms reload time in VET).


rmarquis:
Thx!

PR #1640, opened 2021-04-22
Updated brew command for installing xquartz

Samuel-Martin23:
I’m on macOS Catalina Version 10.15.5. Maybe it’s different for each version of macOS, but the original cask install didn’t work. I found the new command here: https://formulae.brew.sh/cask/xquartz


rmarquis:
Thanks!

Issue #1408, opened 2020-03-15
Issues with hit detection


jackeri:
The "feeling" when shooting is somehow delayed. The way it manifests is just simply: you press the shooting button -> the hit event may or may not register at all or it registers delayed.

Things to look for: antiwarp code, antilag code and the "updated" client & server download code made in ~2015 ish.

rmarquis:
Related UDP download speed change was done in 108ca8503a141f996148b9c5b4dfed6e9ed78f51. See #330.

rmarquis:
Some initial changes done in f73b9c85c95f89644bfcf5bed984505afc9f838c.

Early testing indicates better results already.

rmarquis:
See also #1201 to potentially improve hit "feeling" by having hitsounds generated server side.

rafal1137:
Point of interest
[ETPub](https://github.com/BulldogDrummond/etpub/blob/d6e81ec7d8c62a4fab5c970a850aad049e2ed9ee/trunk/src/game/g_antilag.c#L624) Replace servertime value with attackTime and test the results.

rmarquis:
For the record: some heavy testing was done back in March, with various combination of etded/etlded and Jaymod/Legacy mod (since Jaymod is compatible with etlded, has same hitbox, etc.). The issue has been pigeonholed to the mod code, especially antilag part.


Server | Mod | Client |   |   | Jacker's notes | Spyhawk's notes
-- | -- | -- | -- | -- | -- | --
etded | Jaymod | 2.60b |   |   | With both antilag and antiwarp disabled I was consistenly still hitting the players where I expected to hit. | I'm not used to jaymod, so I played a bit to get the feel of it. No noticeable difference between 2.60b and ETL clients. (lag 0, warp 0): There is little lag noticeable. Aiming at or very close to the target pretty much hits most of the time.
  |   | Legacy master |   |   |   |  
etlded | Jaymod | 2.60b |   |   | Pretty much the same as the vanilla server | (lag 0, warp 0) can't see diff between etlded vs etded (both Jaymod). I'm less sure about the client, but that might be placebo.
  |   | Legacy master |   |   |   |  
[...]

PR #1644, opened 2021-05-11
game: Fixing a rare case where headshot would register as bodyshot, refs #1408

ryzyk-krzysiek:
In a rare occasion headshot would register as bodyshot because the `tr.endpos` gets modified before it is used as a starting point for another trace that checks again for headshot hit. Also the knockback code was affecting this second trace causing it to fail too so I moved it after the trace.

https://streamable.com/x10ni3

refs #1408


rmarquis:
That fix is very interesting. Looks like we're finally onto something tangible to fix the main hitbox issue!

PR #1649, opened 2021-05-19
Support build also on 64 arm - aarch64

pemensik:


ensiform:
I wouldn't reuse idx64 as that is generally accepted for x86_64 not 64 bits generic

pemensik:
Okay, but I found it on ``defined(__APPLE__) && #elif defined __arm64__`` set too, whatever that means. So I copied it over. It seems it is not used for anything serious even in qcommon/q_math.h. Would update it.

rmarquis:
Which platform is ARM 64 based that we might support in the future, and which isn't Android (it has a define already)? I can only think of Apple M1 chips, but the future of this game on that platform is kinda uncertain (OpenGL deprecation).

sebdanielsson:
Besides Apple Silicon we have lots of ARM64 devices like the Raspberry Pi 4, Pinebook Pro. Eventually you should be able to run it on ARM64 Chromebooks when Linux apps are can utilise hardware accelerated graphics on ChromeOS.

sebdanielsson:
> Which platform is ARM 64 based that we might support in the future, and which isn't Android (it has a define already)? I can only think of Apple M1 chips, but the future of this game on that platform is kinda uncertain (OpenGL deprecation).

I have no experience in development but could this keep ET: Legacy on macOS after it's OpenGL deprecation?
https://moltengl.com/moltengl/

rmarquis:
We already have support for Raspberry Pi 4B, as of last year.
To be honest, I am personally not really keen on adding support for exotic hardware just for the sake of it, since we had to remove badly maintained arch in the past. 

Apple Silicon might make sense since it is much more mainstream, with the drawback of not knowing if (or rather when) they'll remove support for OpenGL. The "good" news is that a deprecated OpenGL still work - as long as they keep it.
I don't think using MoltenGL is possible due to licensing reason. It doesn't seem free nor open source, unlike MoltenVK.

sebdanielsson:
> We already have support for Raspberry Pi 4B, as of last year.
> To be honest, I am personally not really keen on adding support for exotic hardware just for the sake of it, since we had to remove badly maintained arch in the past.
> 
> Apple Silicon might make sense since it is much more mainstream, with the drawback of not knowing if (or rather when) they'll remove support for OpenGL. The "good" news is that a deprecated OpenGL still work - as long as they keep it.
> I don't think using MoltenGL is possible due to licensing reason. It doesn't seem free nor open source, unlike MoltenVK.

I found the following on the website. Can this armv7 version still be installed on an aarch64/arm64 version of Linux on a RPI4?
Raspberry Pi RPI armv7 compatible RPI 2/3
Raspberry Pi 32-bit installer (size 22.531 MB)
Raspberry Pi 32-bit binaries (size 22.49 MB)

rmarquis:
Looks like a typo to me. The tested versions were 3 and 4B+ (See our README.md, and #885 #1421 for details).

pemensik:
I would target on Raspberry Pi 3-4 64b. Even if those platforms are not officialy supported, why it shouldn't be easy to compile it on/for them?

rmarquis:
I think the real question is why is 64bit not provided on RPi at all. Asking @techyian, he might have some useful input here.

sebdanielsson:
> unlike

Could this be useful?
https://www.phoronix.com/scan.php?page=news_item&px=Mesa-22.0-Build-macOS-Zink

PR #1651, opened 2021-05-24
qcommon: Fix const correctness in absence of curl

suburbski:


Aranud:
Fix of the year ! Thanks you !

PR #1676, opened 2021-07-25
game: add sudden death

### LINKED ISSUE #1498: Add dynamite sudden death
On maps where the final objective is to be dynamited (such as oasis anti tank guns), dynamites have to be planted before the timer reaches 30 seconds. Sometimes, dynamites are successfully planted, but everyone knows the game is lost because it's not been done early enough.

A nice solution is to implement some "sudden death" mechanics: if the dyna is planted after the 30 sec limits, then the game will not end until the dynamite either explodes or is disarmed - allowing both team to fight until the very end.

This is a feature that has been implemented in etpub, but also in Brink later on and works quite well.

> (g_dyno 1): If there is 30 seconds or less on the clock and dynamite is planted on an objective, the clock will continue to run past zero. The match will not end until the dynamite either explodes, or is disarmed. You are not allowed to plant additional dynamite during sudden death.

Exagone313:
Fixes #1498 

Could be changed/improved:
* remove the cvar
* free the entity when trying to arm a dynamite durring sudden death (like planting near a friendly objective)
* codestyle?
* stopwatch-specific behavior?

### DIFF
--- a/src/game/g_main.c
+++ b/src/game/g_main.c
@@ -4204,7 +4204,7 @@ qboolean DynamiteOnObjective(void)
 	for (e = 0; e < MAX_GENTITIES; e++)
 	{
 		ent = &g_entities[e];
-		if (ent->s.weapon == WP_DYNAMITE && ent->onobjective != NULL)
+		if (ent->s.weapon == WP_DYNAMITE && ent->onobjective)
 		{
 			return qtrue;
 		}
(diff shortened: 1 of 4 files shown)


src/game/g_main.c
```diff
@@ -4298,6 +4322,23 @@ void CheckExitRules(void)
...
 					return;
 				}
+
+				if (level.suddenDeath)
+				{
+					if (DynamiteOnObjective())
+					{
+						return;
+					}
+					level.suddenDeath = 0;
```
Exagone313, line 4332: Not completely sure about this line, in regard to the change in `g_script_actions.c`.

Issue #284, opened 2020-03-05
Make all weapons recoil FPS independent and random


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2013-04-07
Redmine Issue: 284, https://dev.etlegacy.com/issues/284

---

Current recoil value is highly FPS dependent (specially the 71 FPS value).
Recoil should be made (slightly) random and FPS independent.

Note that some mode removed +lookup, -lookup, +lookdown, and -lookdown commands from the client mod to deal with anti-recoil cheats.

Forum discussion link: http://dev.etlegacy.com/boards/3/topics/326

ETLegacyTeam:
From: @realkemon (keMoN)
Date: 2019-05-03T12:28:13Z

---

https://github.com/etlegacy/etlegacy/blob/f0bf85d7e1b1675b9e69ce6b47d3c12604406560/src/qcommon/common.c#L2923

What's going on here?

ryzyk-krzysiek:
The mentioned issue with recoil in PR was just on my end as the servers using snapshot that has this change have no issues, so everything looks to be fine.

PR #1680, opened 2021-08-07
mod: Fix framerate dependence issues

ryzyk-krzysiek:
1. Fixed stamina recovery framerate independence.
2. Fixed weapons overheating framerate independence.
a) Adjusted mg42 firerate and overheating to match etpro. (overall buffed)
b) Adjusted sten overheating to match etpro. Bullets 32->15->Overheating 15->0->Overheating. (overall buffed)
3. Made weapon recoil framerate independent, refs #284.
a) Adjusted pistols recoil to match 125fps recoil.
b) Scoped garand and k43 have lower recoil than before (it's in the middle of 125fps and 71fps) to compensate for the fact that many players used com_maxfps 71 for lower recoil since forever.


ryzyk-krzysiek:
4. Fixed bobcycle framerate independence (weapon swaying, footsteps)

There seems to be a slight issue with weapon recoil on dedicated server with `com_maxfps` 200 and 250. The kickback is a bit higher but then returns to correct position, making the animation not as smooth as 125fps and lower. This is still much better than before where on 200fps the recoil was way too big and it's probably something that players can get used to too. I will be looking a for solution for that, but in the meantime the change should be ok anyway.

Issue #508, opened 2020-03-05
Can't shoot picked up gun while your previous gun is overheated


ETLegacyTeam:
Author: @saukko28 (Saukko)
Date: 2014-02-14
Redmine Issue: 508, https://dev.etlegacy.com/issues/508

---

As title says. For example you shoot with your Sten until it overheats and you pick up a new gun, you can't shoot with it straight away since it appears to be "overheated".
There's a demo to show the issue clearly.

-*S

---

Related commit(s):
6af2532cdad96c56a0aeffed785e83b321b4bb1a - game: Prevent picking up another gun while previou..
 b63e0837c2b6d58392d2070abb70ae4076268a71 - game: prevent picking up while reloading or overhe..

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2014-02-16T12:44:37Z

---

-I would also like to mention that the overheated gun doesn't cooldown while it's on the ground. This has been fixed in other mods, at least in NQ.-

-*S

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2014-02-16T14:44:02Z

---

Saukko wrote:
> I would also like to mention that the overheated gun doesn't cooldown while it's on the ground. This has been fixed in other mods, at least in NQ.

I cannot reproduce this specific issue. As far as I can see, weapons on ground are also cooled down.

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2014-08-06T01:34:13Z

---

Ok it has been a while ago I tried this out but I can still reproduce this and here are the steps to do it:

1. Shoot with Sten until it overheats
2. Pick up a new weapon from the ground

So this causes that you can't shoot with your new weapon instantly, it shows like it's overheated. And the other bug it also causes that if you pick the Sten back up even after 10 seconds, it still haven't cooled down.

-*S

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2015-08-29T14:31:38Z

---

Current fix seems to do a bit too much. Keep in mind the "heat" of the weapon is actually stored on the player, not the weapon. One way to fix that is to check the weapon used and the "heat" status before shooting.

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2017-03-04T22:56:27Z

---

[...]

PR #1686, opened 2021-08-16
game: fix prevent picking up weapon when overheating

ryzyk-krzysiek:
This should fix players being unable to pick up weapons while doing other actions like throwing medpacks. 

The weapons that have heat mechanic are unchanged by this, but there is a bug with them anyway. When you switch a weapon the one on the ground doesn't get its heat reduced. Imo if this bug was to be fixed this whole check could be removed as I don't really see a reason why a player should be prevented from picking up a weapon, it's simply throwing a weapon on the ground.

EDIT: Unless the restriction on heat weapons it's done for balance reasons.

### DIFF
--- a/src/game/g_items.c
+++ b/src/game/g_items.c
@@ -419,7 +419,8 @@ qboolean G_CanPickupWeapon(weapon_t weapon, gentity_t *ent)
 	}
 
 	// prevent picking up when overheating
-	if (ent->client->ps.weaponTime > 0)
+	// FIXME: heat of dropped weapon isn't reduced overtime
+	if (ent->client->ps.curWeapHeat > 0 && ent->client->ps.weaponTime > 0)
 	{
 		return qfalse;
 	}


rmarquis:
See comments in #508.

rmarquis:
I'm closing this issue if that's ok, since the current status has been discussed previously (balance reason).

ryzyk-krzysiek:
But this doesn't change overheating mechanic, you can't pick up a weapon if your sten overheated, so I don't understand what is the problem?

ryzyk-krzysiek:
This fixes player being unable to pick up weapons while doing other actions like medpacks, ammopacks etc. You still can't pick up a weapon if your sten overheated. 

In etpro there is no restrictions even for weapons like sten. My second part of the comment was about when you swap a weapon (sten) with another one, then the one on the ground doesn't get its heat reduced, which imo is a bug. The rest of the message was about maybe lifting the restriction altogheter for such weapons if the bug was to be fixed in the future (etpro allows weapon switching no matter what, but I never heard about it being unbalanced there, covert ops is mostly limited to detroying stuff)

rmarquis:
Oh sorry, I misread your message and conflated both part.

Heating status is stored on players themselves, rather than on the weapon. It's weird, but that helps in understand that specific issue.

rmarquis:
Since this has been merged already, please add a ticket for the FIXME so we don't forget about it.
Edit: Nvm, I'll just reopen #508.

PR #1687, opened 2021-08-20
WIP: introduce dedicated flag to filter 32/64-bit servers

Helldembez:
After a short back and forth with @Aranud we came to the conclusion g_oss is not sufficient for this purpose. It does has a close resemblance to what we need and is already partially implemented in legacy and etjump. Because of this we could check for  g_oss as well, but is not yet pushed in the short serverinfo struct.

Please some feedback as I am not so familiar with the codebase yet and this is hard to test without a master server.

### DIFF
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -1305,7 +1305,7 @@ static int CG_GetSecondaryWeapon(int weapon, team_t team, int playerclass)
 		}
 
 		// is player had the minimum level required to use this weapon
-		if (cgs.clientinfo[cg.clientNum].skill[classInfo->classSecondaryWeapons[i].skill] < classInfo->classSecondaryWeapons[i].minSkillLevel)
+		if (!BG_IsSkillAvailable(cgs.clientinfo[cg.clientNum].skill, classInfo->classSecondaryWeapons[i].skill, classInfo->classSecondaryWeapons[i].minSkillLevel))
 		{
 			continue;
 		}
@@ -1580,7 +1580,7 @@ static void CG_Class_f(void)
 	}
 
 	// Print out the selected class and weapon info
-	if (cgs.clientinfo[cg.clientNum].skill[SK_HEAVY_WEAPONS] >= 4 && playerclass == PC_SOLDIER && !Q_stricmp(GetWeaponTableData(weapon1)->desc, GetWeaponTableData(weapon2)->desc))
+	if (BG_IsSkillAvailable(cgs.clientinfo[cg.clientNum].skill, SK_HEAVY_WEAPONS, 4) && playerclass == PC_SOLDIER && !Q_stricmp(GetWeaponTableData(weapon1)->desc, GetWeaponTableData(weapon2)->desc))
 	{
 		CG_PriorityCenterPrint(va(CG_TranslateString("You will spawn as an %s %s with a %s."), teamstring, BG_ClassnameForNumber(playerclass), GetWeaponTableData(weapon1)->desc), 400, cg_fontScaleCP.value, -1);
 	}
--- a/src/cgame/cg_draw.c
+++ b/src/cgame/cg_draw.c
@@ -3776,6 +3776,20 @@ static void CG_DrawEnvironmentalAwareness()
 		snap = cg.snap;
 	}
 
+	if (snap->ps.pm_flags & PMF_LIMBO
+	#ifdef FEATURE_MULTIVIEW
+	    || cg.mvTotalClients > 0
+	#endif
+	    )
+	{
+		return;
+	}
+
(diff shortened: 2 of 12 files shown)


rmarquis:
Thanks!

However, I think there is a fundamental issue. Brace yourself because it's not easy to grasp at first read: It is not the architecture of the server that matters, but the fact that the **mod** provide a 64 bit version or not.

For example, you can host a Win32 server, with both Win32 mod binaries (so win32 client can connect), and Linux 64 mod binaries (so Linux 64 bit clients can connect).

The same way, a Linux 32 bit client can connect to a Linux 64 bit server *if* the mod provides a Linux 32 bit version of it (as they all do), along with the 64 bit mod. You change above seem to actually prevent 32 bit client to connect to Linux 64 bit servers, regardless of the presence of 32 bits mod binaries.

I know `g_oss` kinda sucks in a lot of way, but I think it is still the way to go. However, it should ideally be set dynamically and automatically on the client.... somehow.

Helldembez:
> Thanks!
> 
> However, I think there is a fundamental issue. Brace yourself because it's not easy to grasp at first read: It is not the architecture of the server that matters, but the fact that the **mod** provide a 64 bit version or not.
> 
> For example, you can host a Win32 server, with both Win32 mod binaries (so win32 client can connect), and Linux 64 mod binaries (so Linux 64 bit clients can connect).
> 
> The same way, a Linux 32 bit client can connect to a Linux 64 bit server _if_ the mod provides a Linux 32 bit version of it (as they all do), along with the 64 bit mod. You change above seem to actually prevent 32 bit client to connect to Linux 64 bit servers, regardless of the presence of 32 bits mod binaries.
> 
> I know `g_oss` kinda sucks in a lot of way, but I think it is still the way to go. However, it should ideally be set dynamically and automatically on the client.... somehow.

Thank you for the explanation. I was specifically looking for this distinction of compatibility, but I was told otherwise. No problem, can you provide any more info on how the value for `g_oss` gets determined?

If we were to use `g_oss` I think it would be best to move the flag to the summarized serverinfo struct, else we will be forced to make double the queries for each server. If that is acceptable however I think it will open the door to fix some other issues as well.

rmarquis:
Let me know if the explanation isn't clear. It's a quite confusing issue, but it helps to consider Linux32 and Linux64 as completely different arch (regardless of cross compatibility on the client), and no different than, let's say, a RaspberryPi server allowing Win32 clients to join.

 > If we were to use g_oss I think it would be best to move the flag to the summarized serverinfo struct, else we will be forced to make double the queries for each server. If that is acceptable however I think it will open the door to fix some other issues as well.

You'd be right here, and I'd theoretically prefer it, but the issue as I see it is that this cvar wouldn't be available to clients unless all the older servers are updated first. And we still see ETL 2.75 servers running, and maybe even older one :(
For this reason I think it's better to treat `g_oss` as a regular mod cvar since it's been implemented years ago, at least for now.

> can you provide any more info on how the value for g_oss gets determined?

`g_oss` is simply set as a bitflag cvar to a default value, depending on the architecture we provide. 

```
(from g_main.c)
// os support - this SERVERINFO cvar specifies supported client operating systems on server
vmCvar_t g_oss; //   0 - vanilla/unknown/ET:L auto setup
                //   1 - Windows
                //   2 - Linux
                //   4 - Linux 64
                //   8 - Mac OS X
                //  16 - Android
```

For 2.77.1: value is 15, and we'll have to update it to 31 as we provide Android binaries in snapshot now.

For the client: 
* if the `g_oss` cvar exist, ideally automatically compare to client architecture (somehow) and filter out servers that are not compatible.
* If the `g_oss` cvar doesn't exist (most third party mods):
  * if the client is win32 or Linux 32, don't filter servers I guess (most if not all servers should provide both, regardless of the mod used)
  * if the client is something else, filter these servers. MacOS is an edge case of a clusterfuck situation here, and I don't know how to handle it.

Note: We'll have to tell mods developers that are still maintained (ETJump, ETRun, Nitmod, ... ) to implement `g_oss`.

isRyven:
Didn't know about `g_oss`, looks like the only sane solution here.

ensiform:
g_oss should be maintained by the server engine itself detecting which client files are available in pak imo. No mod cvar or cvar at all possibly?

isRyven:
This implies only `etlded` would set this. Which rules out `etded` hosted servers.

ensiform:
Should be weaning off of etded servers at this point with etlded and eteded outside of etpro.

isRyven:
Well i believe it's not an option for etjump or etrun yet. Do etlegacy and ete even rely on the same arch prefixes in dlls?

ensiform:
For windows and linux should be good on that front.

Maybe the qagame can do the file check then? Or fallback to server engine as well.

Aciz:
ETJump servers are fine afaik with 32-bit etlded, on 64-bit there's that mempool issue (which was fixed in https://github.com/etjump/etjump/commit/3efd0421d1c1463bd76500af92521d5ba6907f3a) so 64-bit has to wait until next ETJump release.

Issue #1688, opened 2021-08-22
Add "surrender round" vote for attacking team in stopwatch mode


BystryPL:
On some maps after certain time it's obvious that the attacking team won't do the objective in time (for example on supply, when allies can't take the flag on 1st stage of the map). This can prolong the whole match by couple of minutes that are spent on useless shooting. I suggest, if possible, adding voting option for the attacking team to surrender the round (not the whole match) if they are sure that they won't make the objective in time.

Aciz:
There is already surrender vote (and a ref command to surrender too) but the vote is disabled on configs by default. The cvar to allow it is `vote_allow_surrender`. It surrenders the current round (not the whole match).

rmarquis:
We might need to ensure the clock is correctly set in case of surrender (artificially set to the defined time limit).

Aciz:
It correctly sets the clock to fullhold if attackers surrender, and if defenders surrender, it will set the clock to whatever the time was when the vote passed.

rmarquis:
Perfect then. Closing the ticket.

rmarquis:
>  it will set the clock to whatever the time was when the vote passed.

Oops, I misread again. The clock should be set as if the team never achieved the obj until the timelimit finishes. Otherwise, it would be obviously buggy in second round.

Aciz:
I think you misread again :D this is the case if DEFENDERS surrender. If ATTACKERS surrender, the clock will be set to full round time.

rmarquis:
I just realized that... gonna stop reading Github for the day. My bad.

Here's a question though: is there any advantage of the defending team to surrender? Basically they'd shot themselves in the foot, as they would shorten their clock in round 2 (if surrendering in round 1). Shouldn't we prevent surrender by the defending team entirely in SW mode?

Aciz:
Yeah I can't think of any reason really for defenders to surrender, unless they want to make it a challenge for themselves :P So yeah makes no sense really, but I'd say it's also not very critical, you still need to consciously make the decision to surrender the round, it's not like it can happen by accident (unless a defending team member has ref access).

rmarquis:
Attempt to justify the ticket reopening and not appear like a complete fool: SUCCESSFUL :D

Anyway, I think it doesn't hurt to prevent the defending team to surrender. There is no reason it should happen, so let's not allow them to do it.

mittermichal:
Imagine you defend sw_goldrush_te against much better team:
[...]

PR #1691, opened 2021-08-24
mod: allow surrender vote by default on competition configs

Aciz:
Allow teams to surrender by default when `legacy3/5/6` configs are loaded.

refs #1688


rmarquis:
Thx!

PR #1692, opened 2021-08-25
game: fix constructing not working when trace hits ET_MOVER element

ryzyk-krzysiek:
Constructions like tank/truck/CP have elements which some of them are of type `ET_MOVER` and due to changes to fix vanilla issues it is possible for a trace to hit such element making it impossible to be build. I think this is the last issue and constructing should always work just like before.


rmarquis:
Thank for all fixes!

PR #1693, opened 2021-08-25
cgame: add `cg_simpleItemsScale` to control size of simple items

Aciz:
Cvar to control size of items when `cg_simpleItems` is enabled. Scaling factor is limited to __0.1__ - __2.0__.

### DIFF
--- a/etmain/ui/options_customise_game.menu
+++ b/etmain/ui/options_customise_game.menu
@@ -52,11 +52,13 @@ menuDef {
 
 // Items //
 #define ITEMS_Y 194
-	SUBWINDOW( 6, ITEMS_Y, (SUBWINDOW_WIDTH_L), 28, _("ITEMS") )
+	SUBWINDOW( 6, ITEMS_Y, (SUBWINDOW_WIDTH_L), 42, _("ITEMS") )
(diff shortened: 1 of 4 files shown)


rmarquis:
Quick question: is there any real need to get items size twice as big? If anything, I always felt the current size was okayish, but on the limit of being slightly too big already.
Sale question for 0.1... that sounds very, very small to me.

Edit: Also, ideally we'd find some space to add new settings in UI too.

Aciz:
Yeah the limits are a bit silly, but I wanted to just really cover everyone's taste with it. I can change the limits to something bigger/smaller, like let's say __0.25__ and __1.5__

image
image


Also there's plenty of space in settings -> game, where simple items setting is located, can definitely add the menu entry too!

rmarquis:
Ah, great. Adding a UI settings is welcome then.
Lower size range are good to have. Keep in mind that bigger simpleItems icons might be harder to grab, as you need to aim for the lower center to get them. Just being curious, what is your own setting here?

Aciz:
> Just being curious, what is your own setting here?

If you mean what setting I would personally use, probably something like 0.6. The pictures above are with the min/max ranges of 0.25 and 1.5

image

Aciz:
Changed the allowed range to 0.25-1.5, and added menu entry.

rmarquis:
Thanks!

PR #1695, opened 2021-08-27
game: don't count overkill damage towards players stats

ryzyk-krzysiek:
Full damage was counted towards the stats which was inflating numbers on grenades/PF/arty/airstrike.

Player A with PF (400dmg direct hit) hits a Player B with 1hp. Player A gets awarded with 400 damage done and Player B gets awarded with 400 damage received, both stats should be increased by 1.

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -1659,6 +1659,7 @@ void G_Damage(gentity_t *targ, gentity_t *inflictor, gentity_t *attacker, vec3_t
 	// do the damage
 	if (take)
 	{
+		damage        = targ->health;
 		targ->health -= take;
 
 		// can't gib with bullet weapons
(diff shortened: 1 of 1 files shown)


mittermichal:
nice

src/game/g_combat.c
```diff
@@ -1732,7 +1733,7 @@ void G_Damage(gentity_t *targ, gentity_t *inflictor, gentity_t *attacker, vec3_t
 				{
 					// Kill the entity.  Note that this funtion can set ->die to another
 					// function pointer, so that next time die is applied to the dead body.
-					targ->die(targ, inflictor, attacker, take, mod);
+					targ->die(targ, inflictor, attacker, damage, mod);
```
mittermichal, line 1736: does this change how much negative health will the wounded player have?

ryzyk-krzysiek: No, this variable is only ever used to add damage stats to players.

Issue #503, opened 2020-03-05
Fix dead player hitbox direction and bb height


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2014-02-01
Redmine Issue: 503, https://dev.etlegacy.com/issues/503

---

Current hitbox:
http://bani.anime.net/etpro/corpsebb-old.jpg

Better:
http://bani.anime.net/etpro/corpsebb-rtcw.jpg

{FIXME}
The bb height is also wrong. 
Reproduce: throw a nade on dead player.

---

Related commit(s):
a711af4b4e488af467212dd660c2e96e7c5ce1ca - game: hitbox adjustments for proning & dead player..

ETLegacyTeam:
From: @IR4T4 (IR4T4)
Date: 2014-12-21T22:32:22Z

---

Hitbox drection is adjusted (in 2.71). Nades still fly over hitboxes.

ryzyk-krzysiek:
I encountered 2 issues with lower (0 height) BBox that I was unable to reproduce on etpro or height 12, there was another place where I got gibbed because of it but Im unable to reproduce it there so it might have been something else I guess.

https://streamable.com/qeiwn8
https://streamable.com/xoz4ep

ryzyk-krzysiek:
Another 2 examples:

Mentioned above issue: https://streamable.com/9z8wyx
Getting stuck after revive: https://streamable.com/m9wmnq
[2021-10-31-204020-sw_goldrush_te stuck bug.zip](https://github.com/etlegacy/etlegacy/files/7452876/2021-10-31-204020-sw_goldrush_te.stuck.bug.zip)

Helldembez:
https://streamable.com/pwhvh9

Aciz:
This is as good as it gets now without breaking collision, closing.

PR #1696, opened 2021-08-27
game: fix dead bodies blocking missiles because of high BBox, refs #503

ryzyk-krzysiek:
Dead and wounded player bodies have way too big BBox causing all kinds of missiles to be blocked by them. Previously the lower bbox was causing issues but now it seems to be fine.

Old: 
bbox_old

New:
bbox_new

refs #503


mittermichal:
awesome this was pretty annoying when playing with rifle :D
does it affect the reviving?

ryzyk-krzysiek:
Apparently it was affecting it before, but now on local listen/dedicated server I had no issues with it, so hopefully it will not.

PR #1700, opened 2021-08-31
mod: fix scoped FG42 giving light weapons xp

Aciz:
Scoped FG42 was incorrectly giving light weapons XP, and was marked as an Allied weapon, probably c&p virus.


rmarquis:
The change switches it from `TEAM_ALLIES` to `TEAM_FREE`. Is this actually the desired behaviour? Shouldn't it be `TEAM_AXIS` instead?

Aranud:
Both axis and allies have access to FG42, so yes.

rmarquis:
Oh right, I forgot the Allies team doesn't have a specific weapon here.

PR #1704, opened 2021-09-07
Implement 'vignore' and 'vunignore' cmds that'll let you ignore voice…

JumptoHistory:
… chats from certain players.
I think it's nice to be able to ignore annoying voice chat spammers.
Let me know if my code is too messy or I need to fix something.

### DIFF
--- a/src/client/cl_cgame.c
+++ b/src/client/cl_cgame.c
@@ -306,6 +306,12 @@ void CL_ConfigstringModified(void)
 		// parse serverId and other cvars
 		CL_SystemInfoChanged();
 	}
+
+	if (index >= CS_PLAYERS && index < MAX_CLIENTS + CS_PLAYERS)
+	{
+		CL_ClientInfoChanged(index - CS_PLAYERS);
+		return;
+	}
 }
 
 /**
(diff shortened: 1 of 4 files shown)


isRyven:
My two cents. I would  say that you should not implement mod specific features in the engine (there are qagame and cgame modules for this, for example `/ignore` command is implemented there), simply because it's always going to be heuristic, and based on the assumption that all other mods are following the same voicechat command semantics, hence features like this usually belong to mod land. Could be potentially an interesting addition for legacy.

JumptoHistory:
Okay, I thought it's nice having this feature working on most mods but I guess it wouldn't be the best way. Maybe I should implement it in legacy mod?

rmarquis:
I kinda feel that feature would belong more to WolfAdmin (the administration suite we ship) than the mod itself.

rmarquis:
Ah, you mean per-player selection, rather than something global. Hmm..

I'm really not sure what would be the best way to implement this. It feels like a band aid for a deeper issue. On one-hand, if there are spammers then the admins don't do their job as they are supposed to do. On the other hand, the mod might not offer the required noise protection, but since this is server specific, or even time specific, it feels weird to implement a feature just for some troublemakers.

I wonder if something in-between might be better to have, f.e. allowing voice chat of Fireteam members only, so you "whitelist" people you play close to for a given map and ignore the others. This would prevent spammer to simply rename to bypass the filter.

isRyven:
I mean, there is already `/ignore` command like that, it's basically the same thing, except the `ignore` blocks everything coming from user. We should definitely improve the `ignore` itself, to store the state in database. Currently it's client number based, which is not really something reliable, especially when being carried to the next round. Anytime ignored player can be disconnected and on its place another might come, who will remain being blocked. Ideally, this should be guid based, and the mute icon should be shown in a scoreboard for the player is being blocked (something i wanted to implement in etjump some years ago), can be also used for muted players. Then based on that backend you could easily implement the voice ignore command.

On clientside there is also already the `cg_teamChatsOnly` which could have a sibling cvar like `cg_teamVoiceChatsOnly` or such. Dunno whether the naming is good. But would be nice to have something like that as well.

rmarquis:
Right, we also talked about reworking that `/ignore` command some years ago afair. Moving it to be GUID based indeed makes a lot of sense.

`cg_teamVoiceChatsOnly` sounds very good to have, and should be easy and quick to implement, as a first step.

rmarquis:
Closing in favour of the 2 opened tickets. Feel free to discuss other possible implementation.

PR #1707, opened 2021-09-09
cgame: add fireteam name and location customization cvars

Aciz:
Ported over ETPros `b_locationMaxChars` and `b_locationJustify`, and extended the functionality to player name.

* `cg_locationMaxChars` = limits number of chars in location string to `N` (default 0 = no limit)
  * this applies to location strings everywhere, not just fireteam overlay (eg. team chats)
* `cg_fireteamLocationAlign` = aligns location string in fireteam overlay (default 0 = no alignment)
  * < 0 = left align
  * \> 0 = right align
  * when both alignment and max chars are set, width reserved for location display is static
* `cg_fireteamNameMaxChars` = limits number of chars in player names to `N` (default 0 = no limit)
* `cg_fireteamNameAlign` = aligns player names in fireteam overlay (default 0 = no alignment)
  * < 0 = left align
  * \> 0 = right align
  * when both alignment and max chars are set, width reserved for player names is static

### DIFF
--- a/src/cgame/cg_draw.c
+++ b/src/cgame/cg_draw.c
@@ -101,6 +101,50 @@ int CG_Text_Width_Ext(const char *text, float scale, int limit, fontHelper_t *fo
 }
 
 /**
+ * @brief CG_Text_Width_Ext_Float
+ * @param[in] text
+ * @param[in] scale
+ * @param[in] limit
+ * @param[in] font
+ * @return
+ */
+float CG_Text_Width_Ext_Float(const char *text, float scale, int limit, fontHelper_t *font)
+{
+	float out = 0;
+
+	if (text)
+	{
+		const char  *s = text;
+		glyphInfo_t *glyph;
+		int         count = 0;
+		int         len   = Q_UTF8_Strlen(text);
+
+		if (limit > 0 && len > limit)
+		{
+			len = limit;
+		}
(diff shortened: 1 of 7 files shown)


src/cgame/cg_draw.c
```diff
@@ -100,6 +100,50 @@ int CG_Text_Width_Ext(const char *text, float scale, int limit, fontHelper_t *fo
...
 
+/**
+ * @brief CG_Text_Width_Ext_Float
+ * @param[in] text
+ * @param[in] scale
+ * @param[in] limit
+ * @param[in] font
+ * @return
+ */
+float CG_Text_Width_Ext_Float(const char *text, float scale, int limit, fontHelper_t *font)
```
isRyven, line 80: Better to avoid direct duplicates, as it adds unnecessary strain to support.
One simple solution to this problem would be calling `CG_Text_Width_Ext_Float` from `CG_Text_Width_Ext` and casting the result to int.

```c
int CG_Text_Width_Ext(const char *text, float scale, int limit, fontHelper_t *font)
{
    return CG_Text_Width_Ext_Float(text, scale, limit, font);
}
```

I think compiler in the end will do the right thing.

Aciz: Done.

Issue #1487, opened 2020-05-30
Add dynamite chaining


rmarquis:
When many dynamites are set in a single location, dyna chaining blows up all dynamite that are in the blast radius of the first exploding dynamite.

There are a couple of reason to implement this:
* this prevents teammates to be blown up by the second or third dynamite when the first has done its job (f.e., exploding the siwa oasis wall).
* this prevents the "dyna exploit" where players would put two dynamites to blow up a multi-stage objective in a single move, bypassing the intent of the mapper (note: this is still possible if extra dyna are set outside of the first blast radius)
* this greatly simplifies the implementation of the dynamite indicator #1242.

realkemon:
On Discord there was a discussion about pros and cons of this.

Simply detonating all active dynamites in the trigger area changes gameplay behaviour for several entities. Especially, func_constructibles with multiple implemented stages (desstages and constages). These are designed by the game to require multiple dynamites. Removing active dynamites upon changing the first stage means that the second stage remains active. This is in that regard a gameplay change, that now the dynamiting team needs to defend the area for a lot longer time. Which could mean surviving an additional entire enemy spawnwave while they themselves are further away from their spawn. Previously you only needed to defend ticking dynamites. Now you need to defend the ticking dynamites and the engineer to then plant another dynamite which you need to defend.

There are two possible approaches for this.
One would be to make an exception for entity setups purposely designed to be multi-staged like the Fueldump bridge or depot defenses. These entities are `func_constructibles` with multiple constages and desstages implemented.

Another approach would be to only implement dynamite chain for func_explosives. These entities are by design only one time things (e.g. Old City Wall in Oasis, Main/Side Entrance in Radar). Func_constructibles on the other hand can be rebuilt. Another active dynamite ticking means that the defending team cannot simply rebuild the defense that easily. Either they wait for the dynamite to blow up which means they have to defend the area manually, just like the attacker have to when their dynamite is planted. Or they move out of cover to defuse that additional dynamite which means putting the engineer in harms way.

Most of the time defenses like that are easily rebuilt as it is, because the materials are on the defender side of the barrier so removing all active dynamites plays into the hands of the defense.

TL;DR: 
#1 Make an exception to the dynamite chain for func_constructibles with multiple stages implemented (desstages, constages)
#2 Only apply dynamite chain for func_explosive which by design are a single use entity

Helldembez:
[...]

PR #1716, opened 2021-09-15
game: fix dynamite chaining deleting non armed dynamites, refs #1487

ryzyk-krzysiek:
Armed dynamites not on objectives were deleting not armed dynamites that were on objective causing games to be impacted by it. Armed dynamites on objectives were not deleting not armed dynamites ever. 

Changed the behaviour so not armed dynamites are not deleted.

refs #1487

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -2108,7 +2108,7 @@ qboolean G_RadiusDamage(vec3_t origin, gentity_t *inflictor, gentity_t *attacker
 		{
 			G_DPrintf("dyno chaining: inflictor: %p, ent: %p\n", inflictor->onobjective, ent->onobjective);
 
-			if (inflictor->onobjective == ent->onobjective)
+			if (inflictor->onobjective == ent->onobjective && ent->s.effect1Time)
 			{
 				if (g_dynamiteChaining.integer & DYNAMITECHAINING_FREE)
 				{
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -2251,7 +2251,7 @@ qboolean etpro_RadiusDamage(vec3_t origin, gentity_t *inflictor, gentity_t *atta
 		{
 			G_DPrintf("dyno chaining: inflictor: %p, ent: %p\n", inflictor->onobjective, ent->onobjective);
 
-			if (inflictor->onobjective == ent->onobjective)
+			if (inflictor->onobjective == ent->onobjective && ent->s.effect1Time)
 			{
 				if (g_dynamiteChaining.integer & DYNAMITECHAINING_FREE)
 				{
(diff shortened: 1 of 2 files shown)


rmarquis:
Just a note: it should be possible to adjust the code so only non armed dyna on obj are deleted by the exploding dyna on obj (ie, treat "on obj" and "not on obj" as two distinct groups not impacting the others). Not sure how convoluted the code might be, but this might result in a better implementation than never removing non armed dyna.

ryzyk-krzysiek:
Non armed dynamite has `onobjective = NULL` untill the dynamite is armed (and it's on objective), so I don't see how it can be done while keeping the current behaviour of not on objective dynamites not affecting dynamites on objective and vice versa without adding checks for `onobjective` in runentity(?). I wanted to keep it simple and this also doesn't really affect any multistage exploiting or other dynamite chaining functionalities.

rmarquis:
True, the implementation might be too convoluted.

Related question: is there any interest in having non-armed dyna removed at all? If not, and if the gameplay (with multi stage obj and such) isn't affected, would it make sense to enable dyna-chaining by default?

ryzyk-krzysiek:
>  is there any interest in having non-armed dyna removed at all?

Imo there isn't really.

> would it make sense to enable dyna-chaining by default?

dyna-chaining is enabled always, you can't disable it, it's either additional dynamites get deleted (just disappear) or exploded (dealing damage too).

rmarquis:
I meant enabled by the `g_dynamiteChaining` cvar here. It's disabled by default.

Aciz:
> I meant enabled by the `g_dynamiteChaining` cvar here. It's disabled by default.

Just to clarify: the cvar being 0 does not mean the feature is off, it means that dynamite chaining is functioning by exploding the dynamites. If the cvar is set to 1, it functions by deleting them instead of exploding.

https://github.com/etlegacy/etlegacy/blob/a1e3120b350ccfeed6e45b613718d51e5b311b47/src/game/g_local.h#L2125-L2127

Aciz:
Also, multi-stage constructibles are not handled properly, you can destroy an entire multi-stage constructible once the first dynamite exploded, if you plant two.

https://streamable.com/r1p53z

rmarquis:
Hu, gotcha. I wrongly assumed turning the cvar off would give vanilla behaviour.

And yup, multistage is completely broken somehow.

Aciz:
There is absolutely no check for it anywhere, and it's set to explode the dynamites 250ms after the first one, so it's making the original "exploit" of planting multiple dynamites on objective even worse.

Aciz:
I started poking around the code and was wondering what's the intended behavior of dynamite chaining on multi-stage constructibles, should it

* not explode dynamites until we're at the last stage? So essentially disable dynamite chaining until the next explosion would completely destroy it

or 

* remove the dynamites from the objective without exploding them? This would be a gameplay change from vanilla with no way to disable it, keeping in mind what @realkemon wrote in the original ticket https://github.com/etlegacy/etlegacy/issues/1487#issuecomment-656621909

rmarquis:
I'm merging this "deleting non armed dyna" patch. For multi-obj, let's continue the discussion on the original ticket.

PR #1718, opened 2021-09-15
game: fix corpses clipping to solids on suicide animation

Aciz:
Corpse placement uses snapped values for origin, but players are floating 0.125 units off the ground. This caused trace to instantly start in solid and return an invalid position, pushing corpses inside solids when your back was against them.


rmarquis:
Sounds like a great fix! Good job!

PR #1724, opened 2021-09-17
game: fix dynamite chaining on multi-stage objs, refs #1487

Aciz:
Fixes behavior of dynamite chaining on multi-stage objectives to match behavior without dynamite chaining. Dynamites are now chained only if a multi-stage objective is on it's last stage before being completely destroyed. Note that we don't need a check for `chainDynamites` boolean at `G_ChainFree` since it's only called from the radius damage functions, and is already handled at that point.

@rmarquis I never got an answer back as to which behavior is wanted, this or deleting dynamites and requiring another plant. I went with the vanilla behavior, as I think deleting the dynamites and requiring another plant is too much of a gameplay change, since maps over they years have dealt with this behavior and balanced accordingly. Besides, I don't really think a team should be "penalized" if they manage to plant two dynamites before the first one explodes.

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -2053,7 +2053,8 @@ qboolean G_RadiusDamage(vec3_t origin, gentity_t *inflictor, gentity_t *attacker
 	vec3_t    dest;
 	trace_t   tr;
 	vec3_t    midpoint;
-	int       flags = DAMAGE_RADIUS;
+	int       flags          = DAMAGE_RADIUS;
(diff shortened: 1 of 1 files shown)


rmarquis:
Yup, I've been extremely busy today I hadn't time to comment further (nor anyone else apparently).
I kinda see the multi dyna as an exploit, but after giving some thought I agree with you that the proposed solution might be the least invasive. In any case it's much better than the current buggy behaviour which make that exploit even worst.

Issue #1316, opened 2020-03-05
Add option to drop objective (documents, gold, parts etc.)


ETLegacyTeam:
Author: @BystryPL (Bystry)
Date: 2019-06-26
Redmine Issue: 1316, https://dev.etlegacy.com/issues/1316

---

Might be an interesting feature to let players who took the objective to drop it and allow somebody else take it. It's present in other games (CS:GO, for example). I guess it's something to consider. It's not a big game-changer, but certainly would add some interesting aspect to the game. 

---

Related issue(s): #302

---

Related commit(s):
79f9a0593a0d3eebb8e091e9cc036a908be22090 - game: added dropping objective ability, fixes 131..
 777db81da2ce06a938ac759aba00be3333123089 - game: fixed drop objective possible spam, refs 13..
 52b1842b3bb0bb22abb504d50e87ab16a9c1e163 - game: fixed dropped objective being stuck in solid..
 c733a290fed5de517010e11143fc5c6df896456d - game: fixed drop objective timer, refs 1316

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-06-29T16:46:44Z

---

Applied in changeset commit:79f9a0593a0d3eebb8e091e9cc036a908be22090.

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-07-05T07:32:10Z

---


{TODO}
* -Try making the drop message somehow less spammy. We could print a message right after the taking objective, rather than everytime F is pressed.- (see below)
* -consider using a proper bind to drop obj (H key?)- (done, H key)
* -It is also possible to spam with drop messages. See if restricting the number of time a players can pickup an obj (after voluntarily dropping it) can be limited during a defined time.- (done, you might not drop obj after 10 sec of picking it up)
* -objective can be dropped into walls. Prevent this by making it bounce back.- (done, drop it on the ground if view vector is obstructed, like ammo/medipack).

PR #1739, opened 2021-09-27
game: added cvar for delay time when dropping and picking obj back up

Helldembez:


src/game/g_main.c
```diff
@@ -661,6 +663,7 @@ cvarTable_t gameCvarTable[] =
 	{ &g_xpSaver,                         "g_xpSaver",                         "0",                          CVAR_ARCHIVE,                                    0, qfalse, qfalse },
 	{ &g_playerHitBoxHeight,              "g_playerHitBoxHeight",              "36",                         CVAR_ARCHIVE | CVAR_SERVERINFO,                  0, qfalse, qfalse },
 	{ &g_suddenDeath,                     "g_suddenDeath",                     "0",                          CVAR_ARCHIVE,                                    0, qtrue,  qfalse },
+	{ &g_objPickupDelay,                  "g_suddenDeath",                     "10000",                      CVAR_ARCHIVE,                                    0, qtrue,  qfalse },
```
mittermichal: wrong cvar name? :)

Helldembez: thanks! :D

rmarquis:
If I may ask, what is the underlaying thinking behind that cvar? Ie, why is the default value not appropriate, and why is a cvar necessary here instead of a potentially better value? What would be that value for competition config, and why couldn't it apply to public gameplay? etc.

I'm thinking again of minimizing difference between servers, and especially avoiding server cvars for the sake of it.

Helldembez:
Thank you for your question. I agree you don't want to add cvars uncontrollably, but I do feel this one is warranted. When I discovered it was hardcoded I was surprised actually. 

The need for this cvar comes mainly from the competitive scene. 10 seconds is too much to make it a useful feature right now. We could indeed think about a value and change it in code. But like with many things in competition metas change. First reason is that we simply don't know yet what value will be good. We need to experience that by playing multiple matches with different people. So for testing this value alone it would be useful to have it configurable instead of having to commit each testing phase. The community is not decided on the right value yet. There are arguments for it to be low, but also to be longer but not as long as 10 seconds. The same reason comes back if you think long term. In the past competition configs changed through the years. This is a value that can be up for debate when metas change. 

Competitive and public require vastly different settings. But even from a public point of view I see use for this to be configurable. Not all public servers are the same. One might be focused on fragging, another one actually wants objective play to be considered.

Besides these points I am curious why the perceived pushback on adding cvars in general. I do get that you want to keep code clean as possible, cvars might help with that and sometimes they don't. But minimizing difference between servers sounds weird to me. There is already a big divider between comp & public. Then there is another divider between publics itself as not everyone might like the play style of a server. If that were the case there is no need for more than 1 server besides amount of players. This is not the versatility I remember from the good old public days.

Aciz:
Objective dropping is a new thing for competitive play, and the current delay is a bit too high. Since it has really no significance on public play, making it a cvar makes sense so finding a good value that's not too powerful, but also not too long, as experimenting with different values is possible. 

Edit: I'm slow :s

rmarquis:
> The need for this cvar comes mainly from the competitive scene. 10 seconds is too much to make it a useful feature right now. We could indeed think about a value and change it in code. But like with many things in competition metas change. First reason is that we simply don't know yet what value will be good. We need to experience that by playing multiple matches with different people. So for testing this value alone it would be useful to have it configurable instead of having to commit each testing phase. The community is not decided on the right value yet. There are arguments for it to be low, but also to be longer but not as long as 10 seconds. The same reason comes back if you think long term. In the past competition configs changed through the years. This is a value that can be up for debate when metas change.

All right. So many questions are arising following your answer here. But first, here's the background of the reason a delay has been implemented in the first place (see #1316):

Picking/dropping am objective is broadcasted to all players. However, some idiot might take advantage of it to spam everybody for fun, by dropping/pocking up the objective repeatedly. Afaik, the *only* reason we have a delay here is to prevent spam, and it's been added after we saw the potential issue (777db81da2ce06a938ac759aba00be3333123089). It's obviously implemented for public servers, and the used value is a compromise.

From there, I'd completely agree that *not* having a delay in comp setting would make sense, but I have trouble to understand how *some* delay would be beneficial (since it's originally an anti-spam feature). I also feel an on/off cvar would make more sense here, as opposed to a *timed* cvar. In fact, I'd rather have no delay implemented and find another way to prevent spam by idiots.
Could you expand on this, from a comp perspective?

> Competitive and public require vastly different settings. But even from a public point of view I see use for this to be configurable. Not all public servers are the same. One might be focused on fragging, another one actually wants objective play to be considered.
>
> Besides these points I am curious why the perceived pushback on adding cvars in general. I do get that you want to keep code clean as possible, cvars might help with that and sometimes they don't. But minimizing difference between servers sounds weird to me. There is already a big divider between comp & public. Then there is another divider between publics itself as not everyone might like the play style of a server. If that were the case there is no need for more than 1 server besides amount of players. This is not the versatility I remember from the good old public days.

This is a bit OT and more of a global discussion, and something that has been discussed over the years, but still vaguely defined. I'll try to do my best to shed some light. Not everyone might agree, but you might understand my personal point of view better and why I'm quite wary when adding new settings (and I'd say "excessively wary" from my own accord but there is a reason to that). 

1/ Focus on core gameplay:

W:ET has many mods, and most of these are "fun" mods with gazillion of options or feature. Legacy however "aims to add many useful features and improvements while staying close to the original gameplay", and the original gameplay to me is the fantastic objective gameplay. It works on comp, but also (in a more erratic manner) on public servers, and is what ET is.

When you say "one server might focus on fragging", then I don't think that we should aim to help servers be fragging servers. ET is not Quake, and I'm personally against any feature that encourages fragging to the detriment of objective play. Kills numbers don't matter, but getting the obj does. F.e., I've been against shipping some fragging counter or "Ultra Kills" sound pack in our version of WolfAdmin simply because they are detrimental to the core gameplay. We can't ultimately prevent server admins to customize their server the way they want, but I don't see why we should get out of our way to help them.

The same way, Legacy mod doesn't ship "fun" features by design, and admins that want such features are free to use any of the existing fun mod.

Legacy is also quite successful on public servers, so I take that providing a genuine "Wolf:ET" experience is something many want, as opposed to all the "bastardized" mods out there. Technically, there's only ETPro and Legacy in that category, with the former being stuck in the past. While not aiming to be a perfect clone, I see Legacy as not only being a proper ETPro alternative (or very close to it), but also and foremost as being the only provider of the original W:ET experience.

There's more to talk here about the ETPro vs Legacy aspects, but it's always delicate, and even more OT, but you should get the right idea.

2/ Versatility

You touched a bit on that above, and indeed from a code perspective we should avoid cvar for the sake of it. Many of the cvars in some mods (etpro or other) are also simply implemented as default, because they act more like bugfix and result on a better gameplay, with virtually no advantage to have a cvar.

In addition, we moved some features not deemed as "core gameplay" outside of the main code, by allowing these to be optionally available through Lua scripts. F.e., the syringe that heals (instead of medpack) was for some time available as a feature, but we eventually removed it without impeding public servers to use it.

But first and foremost, I see adding a cvar an easy, quick answer to a perceived problem that doesn't ask the right question, and on the long term results on a poorly thought mess. "I want to do X, so let's add a cvar" while I'm more interested in knowing "Why is X needed? Could be X be advantageous to the global gameplay? Is there a way to implement X without drawback for most scenario?". Focusing on the underlying reason for a "needed" feature quite often reveal there is a better way to get the desired behavior.

As a general rule, I see adding feature kinda in the following steps:
* Is the feature an advantage for the core obj gameplay?
* Does it make sense for comp? If not, is it what we'd classify as an undesirable "fun" feature?
* If yes, would public servers make use of that feature, and if not, why?
* Lastly, do we really need a cvar for it?

Of course, there is always a case by case basis for any addition, and it most of the time require iterating on features. We almost never get it right on the first try, but imho the extra thinking and design work result on a better global package.

And finally, to me, the global objective of the project is to:
* steer public servers towards objective gameplay / core ET experience
* provide a well defined "standard" for comp, with a similar experience between servers
* to some extend, allow optional extensibility by community made scripts

BystryPL:
_"From there, I'd completely agree that not having a delay in comp setting would make sense, but I have trouble to understand how some delay would be beneficial (since it's originally an anti-spam feature). I also feel an on/off cvar would make more sense here, as opposed to a timed cvar. In fact, I'd rather have no delay implemented and find another way to prevent spam by idiots."_

It's really hard to answer this question, because - as Helldembez pointed out - we first have to try different values (no delay/some delay - 5s?) to see what will make the most sense for competitive gameplay. Regarding _no delay_ at all, potentially - from competitive gameplay perspective - this feature could be too OP (currently on document/gold maps usually on the last - delivery - stage you play 3v2 or 6v5 respectively), because it would make possible to 1/ take the docs /2 drop them, push enemies, selfkill 3/ respawn, take the docs for 1s to prevent the docs from disappearing, push instantly, thus letting the attacking team play 3v3 and 6v6 respectively on the last stage. 

What I mean is, it COULD break the meta, but we don't know it yet, we have to first test it in a number of matches to decide on that.

rmarquis:
Ah I see. Indeed, I didn't originally foresee the potential issue with selfkill.

Anyway, before merging:

* Is there any default value that would better match gameplay here? If 10 sec is way too much to be any useful, would 5 be okayish? I'd rather have a (somewhat) better default value for gameplay, rather than setting it for potential spam issue. We can always see if that shorter delay is good enough to fight spam.
* Please open a new ticket so we remember to put good, well tested default value in comp config later on. And maybe, ideally remove the cvar if it is not deemed worth keeping.

Helldembez:
If the default is meant to combat spam I find it hard to suggest a default, since I haven't experienced the situation. So I leave that up to others.

Ryzyk also noticed the cvar name doesn't make sense and should actually be something like `objdropdelay` or `objdropcooldown`. Any thoughts on that? I'm fine with `dropObjDelay` or `objDropDelay`, not sure on the grammatical order.

I've added a new ticket as requested, see #1744

rmarquis:
I mean, it's apparently both for spam and gameplay issue, so I'd rather see it as a gameplay cvar first and foremost.
If 10 sec isn't good, what would be the next value you'd test on server?

We use `dropObjectiveTime` as internal variable, so I'd say `g_dropObjectiveDelay` maybe? `g_dropObjDelay` is also good imho.

Issue #1404, opened 2020-03-07
Reduce hitboxes size variability by using capsules instead


rmarquis:
Because hitboxes are cubes that are always oriented in the same direction related to the map, players can be hit with a relative variability of ~41% (100% from the side, 141% from a diagonal). This variability doesn't depend on the players position relative to each others, but indeed depends on the N/S and E/W orientation of the players.

Using an octagon instead would only vary in a 8% range, and this variability would drop to less than 2% with a 16-side shaped box.

Investigate if using such capsule shaped box would be adequate, for both body and head.

Notes:
* Some similar code is already used by the collision box system.
* Beware about reducing the size, as it might be needed to keep the average size so it would be equivalent to 120% but from all directions.
* See: [Quake Champion](https://www.youtube.com/watch?v=1JWaaTJ05x0)
* See if we can do something better for the swimming hitbox (currently using standing hitbox underwater...)

See also: #198

rmarquis:
Q: is there still a need for the fined grained `g_playerHitBoxHeight` cvar?

I'd like to replace it by a simpler `g_playerHitBox` 0/1 (vanilla-level/shoulder-level), so we could expand it to value 2 to allow easy testing of capsule boxes in the not too distant future, and hopefully only temporarily until we settle to more consistent hitboxes.

Note to myself: remember to adjust configs files too.

Calinou:
> See if we can do something better for the swimming hitbox (currently using standing hitbox underwater...)

Could the prone hitbox be used while swimming? The player's visual appearance seems to roughly match the one from the prone animation.

PR #1745, opened 2021-09-28
mod: remove `g_playerHitBoxHeight` and use vanilla hitbox height, refs #1408

Aciz:
This was added as a test some 9 months ago, but basically ever since then, `g_playerHitBoxHeight` has just been set to __48__ on servers, so let's just remove the cvar and use the vanilla hitbox height, as it's what's currently being used.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2584,8 +2584,6 @@ typedef struct cgs_s
 	mlType_t currentMenuLevel;
 #endif
 
-	int playerHitBoxHeight;
-
 	qboolean sv_cheats;         // server allows cheats
 	int sv_fps;                 // FPS server wants to send
 	sampledStat_t sampledStat;  // fps client sample data
--- a/src/cgame/cg_predict.c
+++ b/src/cgame/cg_predict.c
@@ -150,7 +150,7 @@ float CG_ClientHitboxMaxZ(entityState_t *hitEnt, float def)
 	}
 	else
 	{
-		return cgs.playerHitBoxHeight;
+		return DEFAULT_BODYHEIGHT;
 	}
 }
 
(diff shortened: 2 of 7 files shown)


rmarquis:
Not going to lie, I absolutely hate it.

Aciz:
Well of course this doesn't need to be done, it can be left as it is. But the reality is that for example all TM servers, including their main servers and scrim servers, have been running with vanilla hitbox basically ever since this became configurable, so I don't really see the point of having it.

Aranud:
> 
> 
> Not going to lie, I absolutely hate it.

@rmarquis We are talking about the cvar, right ?

rmarquis:
Specifically the return to vanilla height as default. I'm afraid it will make it much harder to move to something more sensible like capsule, if that make sense.

Aciz:
> Specifically the return to vanilla height as default. I'm afraid it will make it much harder to move to something more sensible like capsule, if that make sense.

I really, really doubt having more accurate hitboxes will bring any benefits to the gameplay whatsoever, and I think if you asked the players, basically no one actually wants that.

Keep in mind that this game has:
1. 20 year old netcode
2. quite a lot of spread on weapons and long range gunfights
3. VERY high TTK and damage drop-off at range

A medic with 140HP takes 8 bodyshots to kill with an SMG, barring no damage falloff. With __50%__ accuracy (which is crazy high for the average player), that is __16(!)__ shots. Going for more accurate hitboxes will only make these numbers worse, resulting in lower accuracy and higher TTK.

rmarquis:
I wouldn't call it "more accurate", but globally "more consistent". The idea is *not* to get reduced hitboxes that fit the model as close as possible (like that one jaymod option does), but first and foremost reduce the hit variability that is currently dependent on... the map orientation. Which is completely asinine when you think about it.

See #1404 to get an idea of what this is about, with a very basic implementation idea (which could be improved as the collision system already has something similar apparently). Yes, it would be different than etpro, but certainly not worst like some seem to believe.

PR #1747, opened 2021-09-29
cgame: conditionally disable wepalt switching from weaponbanks

Helldembez:
There have been some complaints when playing rifle that the grenade got loaded/unloaded when pressing weaponbank 3. This has been added as intended behaviour as far as I understand, but disrupts the gameplay when trying to switch weapons quick causing you to shoot without a rnade while you planned on it or the opposite, sometimes even when having no charge so ending up being cannon fodder.

To solve this I have added a check to prevent this from happening with a cg cvar to enable/disable it. Default behaviour stays as it is. I've talked with @Aciz about it and they mention that some players like the effect while in the competitive scene many people dislike it. So hence the cvar. 

I'm not sure if this is the right place/way to do it, but I'm sure its a good way to spark the discussion. Since the same behaviour applies to covert ops silencer, this solution catches that as well. What I was surprised by was that the altwep effect of a sniper is not included in the behaviour of pressing weaponbank 3.


rmarquis:
No idea if that's the best location to do, but I'm usually not against useful client-side cvar.

A few remarks however:
* `cg_altWeapalts` is a bit weird of a name, if I may say. Maybe something like `cg_weapaltSwitches` might be more descriptive? It'd be similar to the current `cg_weapaltReloads` too.
* current code style put the first `{` on a new line too (it's a unlike usual modern standard that your IDE might apply)
* it would be great to add the setting in UI too, if you find some space to do so.

Helldembez:
> cg_altWeapalts is a bit weird of a name, if I may say. Maybe something like cg_weapaltSwitches might be more descriptive? It'd be similar to the current cg_weapaltReloads too.

Yeah thinking of descriptive names is always the hard part. So I decided to just go with something and discuss it later. `cg_weapaltSwitches` sounds good to me.

> current code style put the first { on a new line too (it's a unlike usual modern standard that your IDE might apply)

Ah thanks. Forgot to run uncrustify. Still needed even when so little lines added :D 

> it would be great to add the setting in UI too, if you find some space to do so.

I can try. Haven't done any ui work yet, but I will give it a shot.

Helldembez:
Changed the cvar name and added the ui entry. Tested.

Aranud:
> What I was surprised by was that the altwep effect of a sniper is not included in the behaviour of pressing weaponbank 3.

You don't get the scope weapon by pushing 3 because the weapon isn't present in the weapon bank https://github.com/etlegacy/etlegacy/blob/7a313426df64bef8cc48c1bfdb9521c3cd0a666b/src/cgame/cg_weapons.c#L47
so the following condition https://github.com/etlegacy/etlegacy/blob/7a313426df64bef8cc48c1bfdb9521c3cd0a666b/src/cgame/cg_weapons.c#L5152-L5155 isn't triggered on weapon selection.

Helldembez:
@rmarquis Can this PR be merged or is something still stopping us?

rmarquis:
Yep. Squashed/Merged in 4b79416513d56fa9b3e98954f8af5170eab7c094 and 13808480c1d8f426693dc5feef13f83708987ba1.

Issue #1701, opened 2021-09-02
Make creating/joining a fireteam automatic in stopwatch mode


BystryPL:
A suggestion for competitive gameplay: make creating/joining a fireteam automatic in stopwatch mode. Fireteam is a must in competitive games. It's always a bit of pain in the ass to gather all players in 1 fireteam, also sometimes people forget about it and have to do it just before the game starts or even during the game. Forcing it/making it automatic would be an interesting feature and would prevent those situations. I'm not quite sure how it should work though (forcing a fireteam without leader in each team which players would join automatically upon joining a team?), so I leave it for further discussion.

isRyven:
I would perhaps rather make a cvar for this, like `g_autoFireteam`, which would automatically add a player into a free fireteam on game join. Then you could just enable the cvar in comp config. Making game type specific behavior, is just additional hassle in this case.

Aciz:
> I would perhaps rather make a cvar for this, like `g_autoFireteam`, which would automatically add a player into a free fireteam on game join. Then you could just enable the cvar in comp config. Making game type specific behavior, is just additional hassle in this case.

`g_autoFireteams` is already a cvar, when enabled, it will prompt about joining/creating a fireteam when you first join a team. This could be extended with value __2__ to make the joining automatic too.

isRyven:
Ah yeah, thought so. Sounds good.

PR #1749, opened 2021-09-29
expanded autoFireteams to 2 automatically joining a FT without asking

Helldembez:
Haven't been able to test it when joining an existing fireteam, but I think it should work correctly. I couldn't get connecting with a second client to my debug environment working. `awaiting challenge`


mittermichal:
refs: #1701

Helldembez:
Tested with two clients and there was still popup text for the joining player. This is now removed and should work as intended

Aranud:
Thank you ! Work fine on my side.

Issue #1750, opened 2021-09-30
Cancel votes & ref voting behaviour


Helldembez:
I have received feedback that players want the ability to cancel votes. If I recall correctly, in ETPro you could cancel by pressing ESC or F2 when being the votecaller. Being a ref also had the effect of automatically passing or cancelling the vote.

This implementation of ref voting might not be ideal as it leads to accidental vote passes / fails. Mostly in a comp setting where the ref is also playing as a normal player.

For now I suggest just being able to cancel votes as votecaller and start the discussion regarding the rest.

Aranud:
refs #2085

PR #1751, opened 2021-09-30
game: added vote cancel feature #1750

Helldembez:
See #1750


Helldembez:
Tested with 2nd client and works as intended.

Aranud:
Work fine !

PR #1759, opened 2021-10-02
cgame: swap `cg_simpleItems` values

Aciz:
Makes more sense to "add" features with a higher cvar value, and makes it less awkward to force simple item objectives off for competitive configs.


rmarquis:
Indeed. Thx!

Issue #1740, opened 2021-09-27
Stats reset on 2nd round of stopwatch after map restart or map reset


BystryPL:
This little thing has been annoying me for a long time, it also exists in ETPro. The issue is that if there is a map restart or map reset callvoted before even the 1st round of stopwatch starts, the stas will get resetted on round 2 instead of adding up to round 1 stats.

Aciz:
This is practically impossible to fix as there are limited amount of data that can be stored so that it's persistent across map restarts, and adding more is not possible if we want to keep compatibility with 2.60b. There are few fields free in client persistent, but I'm not sure if stats can even be stored there as I think they consist of multiple fields in `ps->stats`

Edit: read the issue a bit wrong, this seems like a bug that could be solved.

BystryPL:
Some more information/clarification:

[00:43] Bystry: so imagine this, you start the game, round 1. Axis player spawns back due to an error. Players callvote map reset. We are back to round 1, the round starts. The stats after round 1 are correct. We're on to round 2. Round 2 is finished and the stats now include ONLY round 2. The stats from round 1 are not included, as if round 2 was round 1.

[00:46] Helldembez: its not just stats
[00:46] Helldembez: its demo's/stats etc things included in cg autoaction as well
[00:46] Helldembez: that might give you a hint of where it goes wrong

Helldembez:
So I've managed to fix the stats/demo bug. But while working on this it was pretty confusing what the difference is between `map restart` and `match reset` and why.

Differences I noticed:
`map restart` doesnt reset round, but does switch you back to the team you started in in round 1.
`match reset` does reset round, but stays in current team on cmd call.

If `match reset` has the behaviour to reset rounds, then why not switch teams? And if `map restart` already does that too, what the hell is the difference between them if we will edit the behaviour?

PR #1763, opened 2021-10-03
game: reset stats and stop recording on map restart / match reset #1740

Helldembez:
I was looking for a better place to put it, but I couldn't find one. I was considering to put it somewhere around [here](https://github.com/etlegacy/etlegacy/blob/master/src/game/g_session.c#L565) by passing the restart param that's received by the calling function, however that restart does not mean exclusively map restarts. Starting the first round of a game is also a restart apparently.

Also removed some redundant code. I encounter more of this kind of stuff including commented code, but not sure if it was left intentionally.


Aranud:
> Also removed some redundant code. I encounter more of this kind of stuff including commented code, but not sure if it was left intentionally.

I can't this the related commit. Is it missing ?

rmarquis:
This doesn't seem to reset stats in normal game mode (obj and others).

Afair, that's the reason the `G_deleteStats` is set explicitly here. I also remember I got some headache last time I look at that very part of the code.

Helldembez:
> I can't this the related commit. Is it missing ?

`src/game/g_session.c` in files changed.

> This doesn't seem to reset stats in normal game mode (obj and others).
>
> Afair, that's the reason the G_deleteStats is set explicitly here. I also remember I got some headache last time I look at that very part of the code.

Have you tested this or this from memory? Or what do you mean exactly? I tried the defaultpublic config and it still worked fine for me and I see not reason it would not work. The only case it would not work is if in a existing other situation resetStats wouldn't be set to true properly.

rmarquis:
I tested your changes locally. Keep in mind SW and other modes behave quite differently regarding stats, same for campaign mode. You'll likely have to use some if/else condition to ensure the behaviour is correct in all game modes.

Do note game modes are a different thing than comp/public config.

Helldembez:
> I tested your changes locally. Keep in mind SW and other modes behave quite differently regarding stats, same for campaign mode. You'll likely have to use some if/else condition to ensure the behaviour is correct in all game modes.
> 
> Do note game modes are a different thing than comp/public config.

I'm still not sure on what you mean. Maybe it's because we are looking at different things, but for me /scores reset just fine in any gametype. Which I thought you meant wasn't working because of the `G_deleteStats` line i removed. The only thing I notice in LMS is that map_restart does not reset the rounds / score (not scores).

If this is what you mean I can also fix that although that was not the intent of issue #1740 . But while I'm at it could you describe to me the situations that don't reset properly yet? Unless I'm still misinterpreting you.

rmarquis:
Hmm, I might have botched the local test. I'll try again and report back.

I also didn't check `/scores`, but I checked the stats window. It's possible the stats were reset server side, but the changes weren't reflected client side. 

Also: I don't see what you changed on `g_session.c`. The only change I see is the deleted line about G_deleteStats, nothing else.

Helldembez:
> I also didn't check /scores, but I checked the stats window. It's possible the stats were reset server side, but the changes weren't reflected client side.

Scores get reset when next warmup/match start initiates.

rmarquis:
I confirm it doesn't work as I would expect.

I checked the build, refreshed CLion cache, ensured the correct dll are loaded. Nothing.
At this point, if something goes wrong in the build process I wonder what I am missing, or I clearly didn't understand what you were trying to achieve here (ie, `map_restart` not resetting stats unlike what the commit message describe?)

Helldembez:
Are you calling map_restart directly in some way or through voting? I tested this through voting.

rmarquis:
Calling `map_restart` directly. That might be my issue.

Helldembez:
Could you find a moment to test it by voting?

rmarquis:
Yep, I'll try to find some time tomorrow.

Edit: I'm not at home these day and don't have access to my machine, so please bear with my slowness

rmarquis:
Yep, seems better by voting indeed!

Obviously, now I do wonder if putting the code at the first location you were thinking above wouldn't be better. It would require a bunch of if/else to handle some case properly, but it would also fix the general case of using `map_restart` (rarer case, but still useful in case of using rcon f.e.).

Helldembez:
> Obviously, now I do wonder if putting the code at the first location you were thinking above wouldn't be better. It would require a bunch of if/else to handle some case properly, but it would also fix the general case of using map_restart (rarer case, but still useful in case of using rcon f.e.).

As you also worked in this part of the code a bit I think you understand when I say I'm not confident that when the functionality is moved to that place it will be bug free. Those cases to handle do require some more effort to figure out. If you are convinced that is the better place I can give it a go.

One would almost start to think about unit tests 😄

Helldembez:
@rmarquis What about this one? :D

rmarquis:
I'd rather merge this once the hotfix is released, since it touches server side. Something about hotfix and minimizing risks.
Someone more courageous than I am might merge it.

rmarquis:
Squashed/merged in e3c4618253f75b00ac64823ac4232c031f6bc2ce.

PR #1764, opened 2021-10-04
Add PrefersNonDefaultGPU flag to Desktop Entry

Eonfge:
This means that the application will automatically start with the dedicated GPU of laptops, instead of the iGPU.


rmarquis:
Thx!

Issue #1761, opened 2021-10-03
Hide demo recording message


zenixje:
Competitive players always record demos, it would be nice if we can hide the message (or make it shorter/smaller)

https://postimg.cc/nMwhqTFY

Aciz:
You can already achieve this with `cg_recording_statusline` cvar, which controls the Y position of the message, by just moving it off screen. But yeah a dedicated cvar would probably be more convenient.

rmarquis:
Yeah, I kinda hated that long line even for casual play. I am not sure displaying the current time and size on screen is useful at all to be honest. A simpler `⦿ REC` as default would be enough imho.

And by extension, we might simply print `<triangle> PLAY` when playing demo, if that's enough (edit: nvm, the date is still useful here).
Note: unicode chars are simply printed as `.` in the 2.60b client.

rmarquis:
I quickly had a look the other day: I first wanted to move the extra details in console instead, but that would mean cleaning up cgame code to move them to the client. It's more work than it seems worth the hassle, and I am not sure how reliant other mods are on that specific code so it might be better to keep it as is it.

However, what we can do is finding a better cgame cvar to replace `cg_recording_statusline` (using a y-pixel value which seems unnecessary), with the following options:

* 0: hidden
* 1: default, more discreet message only
* 2: longer message, equivalent of what we have currently

Something like `cg_demo_status`.

mittermichal:
I would rather put it as hud element to have these settings nicely at one place
your mentioned values 1&2 could be represented in hud element style attribute

rmarquis:
That sounds indeed better!

zenixje:
The current recording message, red recording is fine.

PR #1767, opened 2021-10-08
cgame: add chat customizability options

Aciz:
Adds multiple customizability options for the in-game chat.

* `cg_chatX/Y` - chat X/Y position
  * note that due to the way chat was previously drawn, addition of these cvars makes it so that by default, chat is slightly more to the right on widescreen aspect ratios
* `cg_chatScale` - chat size scale (__0.5 - 2.0__)
* `cg_chatAlpha` - chat opacity
* `cg_chatBackgroundAlpha` - chat background opacity
* `cg_chatShadow` - toggle shadow on chat
* `cg_chatFlags` - toggle team flags on chat
* `cg_chatLineWidth` - max line width on chat before line break is applied (__10 - 70__)
  * this in includes the chatters name

Also added these options to the in-game menu, under other chat options, and fixed an issue with the background shader drawing incorrectly (there was a solid line on the right side) by drawing it with `clampMap`.

### DIFF
--- a/etmain/scripts/legacy.shader
+++ b/etmain/scripts/legacy.shader
@@ -1425,6 +1425,18 @@ gfx/hud/ranks/xrank11
 	}
 }
 
+// fixed chat background bar using clampMap instead of map for drawing
+gfx/2d/colorbar
+{
+	nomipmaps
+	nopicmip
+	{
+		clampMap gfx/2d/colorbar.tga
+		blendFunc GL_SRC_ALPHA GL_ONE_MINUS_SRC_ALPHA
+		rgbGen vertex
+	}
+}
+
 // disguised shader
 gfx/2d/friendlycross
 {
--- a/etmain/ui/options_customise_game.menu
+++ b/etmain/ui/options_customise_game.menu
@@ -66,22 +66,35 @@ menuDef {
 
 // Chat //
 #define CHAT_Y 32
-	SUBWINDOW( 6+(SUBWINDOW_WIDTH_L)+6, CHAT_Y, (SUBWINDOW_WIDTH_R), 76, _("CHAT") )
+	SUBWINDOW( 6+(SUBWINDOW_WIDTH_L)+6, CHAT_Y, (SUBWINDOW_WIDTH_R), 172, _("CHAT") )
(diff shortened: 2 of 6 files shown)


rmarquis:
I'm thinking about michal comment in #1761 and I wonder: since chat is another HUD element, wouldn't it make sense to implement it as another customizable HUD "widget"?

Helldembez:
> I'm thinking about michal comment in #1761 and I wonder: since chat is another HUD element, wouldn't it make sense to implement it as another customizable HUD "widget"?

Probably yes. Currently HUD elements only support x/y/w/h/style/visibility. So not all these cvars can be replaced by that. I suggest merging this functionality and include this pr/discussion in the HUD issue as this code needs to be modified at that point anyway.

Aciz:
> I'm thinking about michal comment in #1761 and I wonder: since chat is another HUD element, wouldn't it make sense to implement it as another customizable HUD "widget"?

Uh, maybe, but I also do want to include these as cvars, as with our current HUD system, making them only available via custom HUD limits their usage to well, only those who are using custom HUD.

Edit: and yeah as @Helldembez pointed out, not everything added here is customizable via the custom HUD system anyway.

I talked about this very thing today with @Aranud while I was making this, and in some distant future, it would be ideal to ditch the standard, cvar-based HUD system completely, and use only the custom HUD system with some nice user friendly customizability interface, to basically allow customization of everything in the HUD, without having to add a ton of cvars.

rmarquis:
Yes, on the long term that's we should aim for.
As a side note, I think etpub has some sort of in-game HUD editor, but I'm not sure if it is 100% functional. Might provide a good starting codebase.

Aciz:
> Yes, on the long term that's we should aim for. As a side note, I think etpub has some sort of in-game HUD editor, but I'm not sure if it is 100% functional. Might provide a good starting codebase.

Yeah ETPub has one, and it's decent, though iirc it can be only used to move things around and nothing else. But indeed, possibly a good starting base.

So should I also still add custom HUD options for this, namely X/Y position? I've never done anything with the custom HUD code, so I'm not sure if I can do the scale since it takes both X and Y scale, and what even the style option does (textstyle normal/shadowed?).

Helldembez:
> So should I also still add custom HUD options for this, namely X/Y position? I've never done anything with the custom HUD code, so I'm not sure if I can do the scale since it takes both X and Y scale, and what even the style option does (textstyle normal/shadowed?).

Not all HUD elements have width/height/style implemented.

Issue #278, opened 2020-03-05
Servers-side and per-client demo recording


ETLegacyTeam:
Author: @boutetnico (boutetnico)
Date: 2013-04-03
Redmine Issue: 278, https://dev.etlegacy.com/issues/278

---

Implement a per-user server-side demo recording feature.
Provide commands (start/stop) to qagame to allow demo recording control.

Links:

Warsow-race implementation: https://github.com/Racenet/racesow/blob/master/source/server/sv_demos.c
Xonotic implementation: https://github.com/xonotic/darkplaces/blob/div0-stable/sv_demo.c
Q3-arena demo file specification: http://skuller-vidnoe.narod.ru/q3dspecs.htm

---

Related issue(s): #229, #913

---

Related commit(s):
f4e462d8e262d7d8c3de1976a0029928e1199f3c - initial commit of server side demo recording based..
 faa3cc152e9661362cb417f64d75c89362096d53 - server: memory leaks fixed in sv_demo.c
 c75d03b8e7f472ef94888cc4943f77ca4162337a - qcommon: catch invalid playerState field count in ..
 7e1d474f183f565cde9ad99abc3fb4d44824ae3f - qcommon: exact error messages
 93669de22099ff0aa4a490c59f92da6702c5a0fe - server: our world doesn't know the tell cmd
 d36ae5c78a08d969b2bb0e6dd4563b5e1e2ac9e5 - server: fix illegal memory access for bad demo mes..
 d340cad23684fd86c45fa9d338633d6b6aba8159 - server: move the demo cl command capturing to its ..
 3c3e6b20100ba420c6747629bb17c6d805cd6072 - server: cleaned up the server side demo impl
 8ed2762015ee1dfaaccb73ba1126a3f6464df265 - build fix
 c6177b1ba20f05bea5f75e1a23902f5c4cb3ee5a - renderer2: fix char array comparison (not empty)
 972a2b237cd2761b731ef18b5b8c5822cc455556 - server: NULL check for userinfo removed -  MSG_Rea..
 c2555d500bb518b9ef0f038ee94157d35c49a784 - svdemo: writing userinfor in the wrong place
 925cd95b5bee42fa2fd38eec974947740ce9f363 - server: Fix mover state in demo in SV_DemoReadAllE..
 fa59f3ee99c31adc6f9e6b560e6dfe2eeed01347 - docs: README-serverside-demos.md
[...]

Issue #1346, opened 2020-03-05
Add GUID collision check


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2019-09-08
Redmine Issue: 1346, https://dev.etlegacy.com/issues/1346

---

Since PB was responsible for giving unique ID and check for collision (dropping players with non unique ID), it's removal from the open source code led to adding an etkey generator in the client.

However, 2.60b clients that have no GUID can still possibly connect to any mod. In Legacy mod, we added a check in 73f57387ee5a1d170cdde9924bec2eebc75f042d to prevent players with no guid to connect (due to issue with Skill Rating, see #1344, and obviously invalid guid make administration harder). It might be wise to prevent this server side too.

* ~~Add GUID check for validity (empty, "unknown", "NO_GUID", wrong format)~~ (done)
* ~~Add check for duplicate GUID on server (beware of multiple clients use for dev mode though!)~~ (done)
* ~~Drop clients with invalid GUID with a nice message ("Go get ET:L client, or use etkey.org")~~ (done)
* ~~If the check is done server side, the relevant mod code could be removed (Legacy mod is not designed to run on the vanilla server, and Skill Rating depends on database code added to ETLegacy).~~ (done)
* Maybe move some other part of the UserInfo string to server check? (f.e. empty name).

rmarquis:
TODO:
* ~~Also add GUID content check (hex string). Currently only the length is checked.~~ (done)

Aranud:
@rmarquis 

> If the check is done server side, the relevant mod code could be removed (Legacy mod is not designed to run on the vanilla server, and Skill Rating depends on database code added to ETLegacy).

I don't understand what should be done. All the GUID check are already done in game mod (server side so). Am I missing something ?

rmarquis:
For Legacy mod, this is fully done (nice addition btw).

The question is: if this worth adding support for all mods? In which case, the check should be done in the server (sv) part of the code. So third party mods that don't have such a check would have it when running on etlded. If this is ever done, the code in Legacy game code would become redundant and would need to be removed.

I don't know how complex or even possible it is, but I personally _don't_ think this is worth it, since I don't really care for these mods anyway, and some of them already have this feature anyway (like NQ). I'd _personally_ close the ticket and consider it done :)

Aciz:
Continuing the discussion from #1769 

> rmarquis commented
[...]

PR #1769, opened 2021-10-08
server: fixing server demos v2, refs #278

ryzyk-krzysiek:
Most important changes:

- fixed server demo playback on non-dedicated server (can start watching server demos in etl client)
- fixes to gamestate replaying and new metadata
- reworked handling of game commands
- added dummy client connecting at the start of recording so various game stats can be saved to demo
- added `sv_autoDemo 2` - demos will be auto recorded only when there are players on server (not bots) and will stop if they leave

There could be a lot to write to explain why I did some things the way I did, so instead trying to write them (there are comments in code explaining in some cases anyway) I can answer questions if there are any.

refs #278

### DIFF
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -601,7 +601,7 @@ void SV_RestartGameProgs(void);
 qboolean SV_inPVS(const vec3_t p1, const vec3_t p2);
 qboolean SV_GetTag(int clientNum, int tagFileNumber, char *tagname, orientation_t *orientation);
 int SV_LoadTag(const char *mod_name);
-void SV_GameSendServerCommand(int clientNum, const char *text);
+void SV_GameSendServerCommand(int clientNum, const char *text, qboolean demoPlayback);
 
 void SV_GameBinaryMessageReceived(int cno, const char *buf, int buflen, int commandTime);
 
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -90,6 +90,17 @@ typedef enum
 } serverState_t;
 
 /**
+* @struct svdemoPlayerStats_t
+* @brief Stores player stats during recording
+*/
+typedef struct svdemoPlayerStats_s
+{
+	char guid[9];
+	char name[MAX_NAME_LENGTH];
+	char stats[MAX_STRING_CHARS];
+} svdemoPlayerStats_t;
+
+/**
  * @struct server_t
  * @brief
  */
@@ -148,6 +159,7 @@ typedef struct
 	// serverside demo recording - previous frame for delta compression
 	sharedEntity_t demoEntities[MAX_GENTITIES];
 	playerState_t demoPlayerStates[MAX_CLIENTS];
+	svdemoPlayerStats_t demoPlayerStats[100];
 
 } server_t;
 
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -482,7 +482,7 @@ extern cvar_t *sv_guidCheck;    ///< enable check for client guid validity
 // sv_demo.c
 void SV_DemoAutoDemoRecord(void);
 void SV_DemoRestartPlayback(void);
-void SV_DemoReadFrame(void);
+qboolean SV_DemoReadFrame(void);
 void SV_DemoWriteFrame(void);
 qboolean SV_DemoClientCommandCapture(client_t *client, const char *msg);
 void SV_DemoWriteServerCommand(const char *cmd);
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -492,7 +492,6 @@ void SV_DemoWriteClientUserinfo(client_t *client, const char *userinfo);
 qboolean SV_CheckLastCmd(const char *cmd, qboolean onlyStore);
 void SV_DemoStopAll(void);
 void SV_DemoInit(void);
-void SV_DemoShutdown(void);
 
 // sv_demo_ext.c
 //int SV_GentityGetHealthField(sharedEntity_t *gent);   // Test purpose
--- a/src/server/sv_ccmds.c
+++ b/src/server/sv_ccmds.c
@@ -341,7 +341,7 @@ static void SV_MapRestart_f(void)
 			continue;
 		}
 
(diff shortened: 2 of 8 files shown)


ryzyk-krzysiek:
- added fast-forward `demo_ff` command to server demo playback - the fast-forward will stop on gamestate change or demo end
- `tinfo` and `entnfo` game commands will not be saved to demos anymore, during playback qagame will handle it and the data they contain can be extracted from saved sharedentities etc.

Last big feature missing imo is generating stats file at the end of the recording. Gathering such data shouldn't be too hard as current implementation of game commands storing during playback could be used during recording and dump it at the end of the demo. What I'm not sure about is how it should be presented and how to manage identifying players as nicknames and even clientNums could change during the game. I imagine the file should contain `gstats`, `/scores (sc)` and maybe obituaries (`EV_OBITUARY`).

rafal1137:
@ryzyk-krzysiek There is an issue with your PR here. When you start a server with server demo playing you cannot connect to it.
Saying `Server is Full`.

ryzyk-krzysiek:
> 
> 
> @ryzyk-krzysiek There is an issue with your PR here. When you start a server with server demo playing you cannot connect to it. Saying `Server is Full`.

It most likely isn't the issue with PR but the demo you try to watch was recorded on a server with `sv_maxclients 64`. When playback starts the server is set to reserve 64 client slots for demo clients leaving you no space to join as eltded supports only 64 players on the server.

rafal1137:
@ryzyk-krzysiek So has it been changed by you ?
Because on previous version you could start a demo on server and join it while rest of the free slots were reserved for demo clients.

ryzyk-krzysiek:
> 
> 
> @ryzyk-krzysiek So has it been changed by you ? Because on previous version you could start a demo on server and join it while rest of the free slots were reserved for demo clients.

If you recorded demo on a server with `sv_maxclients 64` there would be 64 slots reserved for demo clients and none for you because you reached max, as of now you can't watch demos recorded on a server that had `sv_maxclients 64` set.

If you record your own demos make sure you first set `sv_maxclients` to less than 64, and if you watch old demos then don't watch them, they will not work correctly regardless.

ryzyk-krzysiek:
And no I didn't change it, it has always worked like that and it's one of the many flaws this server demos have.

ryzyk-krzysiek:
Added saving player stats at the end of the recording, it most likely isn't the best way to do it or even good way, but that's the best I could come up with at the moment. Stats include: 8 first characters of guid, nickname, gstats game command. File example:
```
214FF34C\ryzyk3\gstats 0 0 0 0 0.00 0.00 0
484D4A53\ryzyk2\gstats 1 1 2048 0 0 0 1 0 0 262 0 0 0 0 0 0 20.5 0 0.00 0.00 0
8EF70C6E\ryzyk1\gstats 2 1 2088 2 5 0 0 0 4 6 1 0 3 2 1 1 0 0 464 58 0 0 1 0 0 0 100.0 0 0.00 0.00 0
```

gstats include all relevant stats players have (the first two arguments `gstats 0, gstats 1 etc.` should be ignored) - all weapon stats like acc, hits, headshots and overall stats like dmg, tdmg etc. 

The stats are based on guid not on `clientNum` so after players disconnect/reconnect the stats shouldn't be overwritten by other players.
Parsing gstats in cgame: https://github.com/etlegacy/etlegacy/blob/63d4bc416ad813096c2e866510c916666a876e8d/src/cgame/cg_servercmds.c#L2140-L2378

Generating gstats in qagame: https://github.com/etlegacy/etlegacy/blob/108d75925b2c38e1cfdb81749f3231e61aa9ed92/src/game/g_match.c#L395-L500

rafal1137:
> 
> 
> > @ryzyk-krzysiek So has it been changed by you ? Because on previous version you could start a demo on server and join it while rest of the free slots were reserved for demo clients.
> 
> If you recorded demo on a server with `sv_maxclients 64` there would be 64 slots reserved for demo clients and none for you because you reached max, as of now you can't watch demos recorded on a server that had `sv_maxclients 64` set.
> 
> If you record your own demos make sure you first set `sv_maxclients` to less than 64, and if you watch old demos then don't watch them, they will not work correctly regardless.

Yeah setting `sv_maxclients` to lower value made it work. Yes I know I did recorded a new fresh server demo right after I did compiled etl with your changes since I though about compability issues that can happen with old demos.

mittermichal:
is it possible for player to not have GUID? If yes, what happens then?

ryzyk-krzysiek:
In legacy mod I don't think it's possible to not have guid, incase of missing guid stats will not be saved for that player, but probably would be good to add a check for it too in the future as there is no need do stuff when there is no guid.

Other mods are not really supported anyway (opposite to what was claimed initially by the author), because they can and do have different configstring nums and different `gentity_t` which means we can't set things we have to set otherwise the server demo playback doesn't work.

rmarquis:
GUID check has been implemented in #1346. All the database stuff kinda relies on it (SR, WolfAdmin and maybe other stuff), so it's basically made mandatory.

I am aware of one edge case we had in the past, which involves a player connecting to a Legacy server directly from another mod, which kinda leaks the old userinfo into the new one as a result (and thus in some case, no GUID). I am not sure if we actually fixed that case. Ryven might remember some details here.

Aciz:
I think even on legacy mod you can have no GUID, if the server is running with `sv_guidCheck 0`?

rmarquis:
You are right, it's not actually "mandatory" but enabled by default. The ETL client ensure the GUID is properly set, but the old 2.60b client does not, and that's the one that has the edge cases afaik.

Aciz:
And TM main for example runs with `sv_guidCheck 0`, as well as TM scrim servers, to keep compatibility with 2.60b... I am literally at loss of words how stupid this cvar is, all it does basically is prevent 2.60b clients by default from playing on ETL servers. This check should be in the mod, not in the server, because it's the mod that cares about if you have GUID, not the server.

rmarquis:
There's some discussion about this server/mod stuff in #1346. This might need to be reversed if it is really problematic.

rmarquis:
Merged, thx!

PR #1770, opened 2021-10-09
cgame: allow selectively disabling weapon animations

Aciz:
Allow players to selectively disable certain gun animations with more granular control than `cg_gun_frame` offers via `cg_weapAnims` cvar. This is a bitmask cvar, with following values:

* 1 = movement/idle animations
* 2 = firing animations
* 4 = reload animations
* 8 = switch animations

By default everything is enabled of course ( `cg_weapAnims 15`)

### DIFF
--- a/etmain/ui/options_customise_game.menu
+++ b/etmain/ui/options_customise_game.menu
@@ -46,19 +46,20 @@ menuDef {
 
 // Weapons //
 #define WEAPONS_Y 146
-	SUBWINDOW( 6, WEAPONS_Y, (SUBWINDOW_WIDTH_L), 42, _("WEAPONS") )
+	SUBWINDOW( 6, WEAPONS_Y, (SUBWINDOW_WIDTH_L), 54, _("WEAPONS") )
 	YESNO( 8, WEAPONS_Y+16, (SUBWINDOW_WIDTH_L)-4, 10, _("Automatic Reload:"), .2, 8, "cg_autoReload", _("Automatically reload when ammunition clip runs out") )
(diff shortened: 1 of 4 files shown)


rmarquis:
I suppose the purpose is to be less distracted?

I'm wondering if we really need to have these "hacky" features in UI settings, as they clearly don't apply to the vast majority of players. Or at least, we might want to have in some separate section, and regroup them with the similar cvars like `cg_gun`, `cg_gunX`, `cg_gunY` etc. that currently aren't available in UI. Could be on the same setting page, just not in the middle of the more "common" settings, if that makes sense.

Edit: it's technically "view", not "gameplay" options. Same for chat view options I guess, though both categories tend to overlap sometimes. We might even think about merging them.

Aciz:
Yes, some people like to play with `cg_gun_frame 1` (or another commonly used frame 6) for example, which freezes all weapon animations to frame 1. This is essentially just a more flexible and advanced variation of that, while also fixing two issues `cg_gun_frame` has, which are that some weapons don't really work well on animation frame 1/6 (as evident by the switch case I wrote for default weapon frame) and the fact that if you have it enabled, your game will crash upon map load unless you remember to turn it off before that.

I specifically wanted to add this to UI as a bitmask value like this can be hard to remember, but of course I can move it somewhere else, I just put it there along with other gun related options.

mittermichal:
usage of custom cg_gun_frame and cg_gun*: https://www.twitch.tv/videos/1173155312 :)

rmarquis:
Thanks to both of you for input!

Yes, the bitmask in UI makes sense indeed. I'm more worried about the UI becoming even more a mess than it already is, but that's not related to the feature itself.

The ticket can be merged now, we can add other gun related cvars and clean up the UI later on (ideally soon, before the next release).

Aciz:
Yep, we should give the UI some actual thought instead of just seemingly randomly slapping things here and there.

PR #1773, opened 2021-10-10
app: use same versioning string for Android snapshot

rmarquis:
This adjusts the version string for Android snapshots, as to make it identical to other platforms.
Adds the 'g' and reduce the hash to 7 chars.

Currently:
* android: etlegacy-v2.78.0-11-f52f83c2
* others: etlegacy-v2.78.0-11-gf52f83c


rafal1137 (approved):
Good catch.

PR #1784, opened 2021-10-17
 cgame: fix SIGSEGV when parsing hud in 64bit arch 

mittermichal:
I believe structure padding was the culprit. 
Used same 'trick' that's used in `entityStateFields[]`. 
Order of `hudComponentFields` doesn't matter now.

Not sure whether to use `size_t offset` or `int offset`. `size_t` is supposed to be correct and it should not bring warnings. `int` uses less memory and is big enough now.

This would also allow creation of field aliases e.g. adding `HUDF(compass)` to `hudComponentFields[]`, to fix the typos in HUD component names, while keeping backward compatibility.

Thanks to @mightycow for pointing out wrong pointer arithmetic :cow:

### DIFF
--- a/src/cgame/cg_draw_hud.c
+++ b/src/cgame/cg_draw_hud.c
@@ -79,39 +79,49 @@ hudStucture_t hud0;
 lagometer_t lagometer;
 
 /**
-* @var hudComponentName
-* @brief Orderer following hudStucture_t fields declaration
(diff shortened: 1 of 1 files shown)


rmarquis:
Good catch. That might also explain why it happened on macOS and not windows (which is still 32 bit).

PR #1785, opened 2021-10-17
mod: fix HUD component name typos

mittermichal:
- keep backward compatibility by using aliases
- exclude aliases from CG_PrintHud


ensiform:
Using offsetof macro would be cleaner than the casting to 0 ugliness.

mittermichal:
Agree, I found that `offsetof` exists while working on this, but remembered msg.c:NETF macro, so I just went with approach.

mittermichal:
Used the `offsetof` macro, but not sure if I included `<stddef.h>` at correct place

PR #1787, opened 2021-10-22
Adjust legacy competitive configs

BystryPL:
Adjust legacy1, legacy3, legacy5 and legacy6 configs to match the values currently used in actual competitive Legacy games. Provide a universal .cfg + .config files package for competition server administrators. 

What has been done:

- force `sv_fps` and `snaps` value to "40" (value has been tested for 2-3 months now and it seems is pretty much safe to play on);
- set `g_countryFlags` to "1" to allow the shoutcasters see the players' flags in shoutcaster mode;
- force `cg_visualEffects` to "0" to prevent plane giving away airstrike's place of explosion;
- force `cg_drawEnvAwareness` to "0" as the feature is too over-powered for competitive games;
- set `g_autoFireteams` to "2" to enable automatic fireteams;
- ~~set `g_dynamiteChaining` to "1";~~ - CVAR obsolete
- adjust `rate` and `cl_maxpackets` value;
- add some cvars from `etl_server.cfg` that should be in the `.config` files: `sv_floodProtect`, `g_misc`, `g_skillrating`, `pmove_msec`, `g_maxWarp`, `g_antilag`, `g_realHead`;
- add `vote_allow_antilag` cvar to prevent changing antilag value through voting mid-game;
- add to the configs the newest versions of ETL map overhauls that are currently being used in competitive games instead of standard versions: `etl_ice_v8`, `etl_adlernest_v2`, `etl_supply_v6`, `etl_frostbite_v15`. Those will have to be updated regularly together with the release of new overhauls versions;

TODO:

- provide a `etl_server_comp.cfg` template file for competition servers that will complement the `legacy1-6` configs. This will provide a ready-to-go "package" for competition server administrators (`etl_server_comp.cfg` + `legacy1-6` configs) that would require little to no adjustment on their side and that could be efortlessly deployed on the competition server;
- adjust `defaultpublic` and `defaultcomp` configs so that the cvars values changed by competition configs get reverted to default ones;

### DIFF
--- a/etmain/configs/legacy1.config
+++ b/etmain/configs/legacy1.config
@@ -10,6 +10,7 @@ init
 {
 	setl sv_pure "1"
 	setl sv_cheats "0"
+	setl sv_fps "40"
 
 	setl g_gameType "3"
 	setl g_warmup "15"
@@ -36,6 +37,7 @@ init
 	setl g_speed "320"
 	setl g_gravity "800"
 	setl g_knockback "1000"
+	setl g_countryFlags "1"
 
 	setl team_maxSoldiers "0"
 	setl team_maxMedics "0"
@@ -126,15 +128,16 @@ init
 	command "sv_cvar cl_yawspeed EQ 0"
 	command "sv_cvar cl_timenudge EQ 0"
 
-	command "sv_cvar cg_simpleitems IN 0 1"
-
+	command "sv_cvar cg_simpleItems IN 0 1"
+	command "sv_cvar cg_visualEffects EQ 0"
+	command "sv_cvar cg_drawEnvAwareness EQ 0"
 	command "sv_cvar cg_fov IN 75 130"
 	command "sv_cvar cg_shadows IN 0 1"
 	command "sv_cvar cg_autoaction IN 2 7"
 
-	command "sv_cvar rate EQ 25000"
-	command "sv_cvar cl_maxpackets EQ 125"
-	command "sv_cvar snaps EQ 20"
+	command "sv_cvar rate IN 25000 45000"
+	command "sv_cvar cl_maxpackets IN 100 125"
+	command "sv_cvar snaps EQ 40"
 	command "sv_cvar com_maxfps IN 40 250"
 
 	command "sv_cvar m_pitch OUT -0.015 0.015"
@@ -158,7 +161,6 @@ init
 	command "sv_cvar r_nv_fogdist_mode INCLUDE NV GL_EYE_RADIAL_NV" // vanilla clients only
 	command "sv_cvar r_subdivisions IN 1 20"
 	command "sv_cvar r_lodcurveerror GE 60"
-
 	command "sv_cvar r_wolffog EQ 0"
 	command "sv_cvar r_zfar EQ 0"
 }
--- a/etmain/configs/legacy1.config
+++ b/etmain/configs/legacy1.config
@@ -40,7 +40,9 @@ init
 	setl g_knockback "1000"
 	setl g_countryFlags "1"
 	setl g_misc "0"	
-	setl g_skillrating "0"			
+	setl g_skillrating "0"
+	setl g_playerHitBoxHeight "48"
+	setl g_dropObjDelay "3000"
 
 	setl team_maxSoldiers "0"
 	setl team_maxMedics "0"
--- a/etmain/configs/legacy1.config
+++ b/etmain/configs/legacy1.config
@@ -11,6 +11,7 @@ init
 	setl sv_pure "1"
 	setl sv_cheats "0"
 	setl sv_fps "40"
+	setl sv_floodProtect "0"
 
 	setl g_gameType "3"
 	setl g_warmup "15"
@@ -38,6 +39,8 @@ init
 	setl g_gravity "800"
 	setl g_knockback "1000"
 	setl g_countryFlags "1"
+	setl g_misc "0"	
+	setl g_skillrating "0"			
 
 	setl team_maxSoldiers "0"
 	setl team_maxMedics "0"
@@ -70,6 +73,8 @@ init
 	setl sv_maxping "0"
 
 	setl pmove_fixed "0"
+	setl pmove_msec "8"
+
 	set nextmap "map_restart 0"
 
 	setl g_allowVote "1"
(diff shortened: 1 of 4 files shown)


rmarquis:
Thank you!

There's still an issue, when the comp legacy1-6 configs are loaded, and then reverted back to public config: these cvars won't be set back to "proper" public settings. As such, we need to ensure they are reverted in the defaultpublic config too.

I'm also not sure how to treat the global "defaultcomp" config. Is that one used on servers with more than a few players? If so, the respective values should be set to what works best (snap, sv_fps and such).

rmarquis:
Also: `g_dynamiteChaining` is now obsolete. The default behaviour is set to what was `1`, but the cvar doesn't exist anymore.

Aciz:
> There's still an issue, when the comp legacy1-6 configs are loaded, and then reverted back to public config: these cvars won't be set back to "proper" public settings. As such, we need to ensure they are reverted in the defaultpublic config too.

I think a cvar backup system for server side, like clients have for when cvars are forced by server, would be a good solution to this.

rmarquis:
Yep indeed. The current behaviour is quite annoying.

BystryPL:
> There's still an issue, when the comp legacy1-6 configs are loaded, and then reverted back to public config: these cvars won't be set back to "proper" public settings. As such, we need to ensure they are reverted in the defaultpublic config too.

I can take care of adjusting the `defaultpublic` and `defaultcomp` too if it's needed.

> I'm also not sure how to treat the global "defaultcomp" config. Is that one used on servers with more than a few players? If so, the respective values should be set to what works best (snap, sv_fps and such).

Well I've been actually thinking about that too and I don't see any actual use for `defaultcomp` config. In competitive games, this config is never used, only `legacy1-6` configs are used. I guess this config could be similar to legacy6 config, with some small modifications perhaps - I can maybe suggest something. Also never seen this config in use on public servers, as they usually just use the server.cfg instead of configs.

> Also: `g_dynamiteChaining` is now obsolete. The default behaviour is set to what was `1`, but the cvar doesn't exist anymore.

Thank you, I will adjust this.

rmarquis:
> I can take care of adjusting the defaultpublic and defaultcomp too if it's needed.

That would be very welcome.

Yes, I am not sure if `defaultcomp` is used at all. I've always seen it as some sort of "casual comp" training when there are too many people to actually play 6vs6, kinda a mix between pub and comp play, maybe without weapon restriction etc. Feel free to suggest good values here.

If that conf is deemed completely useless, we might think about removing it entirely.

rmarquis:
A few other remarks:

> set g_countryFlags to "1" to allow the shoutcasters see the players' flags in shoutcaster mode;

Do note that while `g_countryFlags` allow the server to display flags, the behaviour is still controlled client side by `cg_allowGeoIP`. There are some privacy issue inherent to that feature, so I am not sure it would be a good idea to force it (cg_allowGeoIP) either. In a nutshell, a client that disables it will show a default UN flag on the server.

> force cg_drawEnvAwareness to "0" as the feature is too over-powered for competitive games;

Do you mind quickly expanding on this? Env awarness is still in the work, there might be some useful insight here to make it more adequate also for public play.

BystryPL:
> Do note that while `g_countryFlags` allow the server to display flags, the behaviour is still controlled client side by `cg_allowGeoIP`. 

Yes, I'm aware of that. It's just to make sure that the flags are enabled on the server side regardless of user's settings. So the shoutcasters can see as many flags as possible.

> There are some privacy issue inherent to that feature, so I am not sure it would be a good idea to force it (cg_allowGeoIP) either. In a nutshell, a client that disables it will show a default UN flag on the server.

You mean actually forcing the `cg_allowGeoIP` cvar server-side? I haven't thought of that actually, and indeed that might not be a good idea from privacy point of view.

> Do you mind quickly expanding on this? Env awarness is still in the work, there might be some useful insight here to make it more adequate also for public play.

Well I was mainly having in mind the fact that you can see the reviving icons through walls which in my opinion is a bit too intrusive and too over-powered (usually finding a wounded teammate requires some voice communication). 

I have another question though, I'm also currently working on a `server.cfg` template for competition servers that will complement the legacy1-6 configs. That template could be shipped with ET: L just like `etl_server.cfg` and `legacy.cfg` are shipped now. This would provide a ready-to-go "package" for competition server administrators (`etl_server_comp.cfg` + `legacy1-6` configs adjusted to current competitive standards) that would require little to no adjustment on their side and that could be just deployed on the server. My aim is to keep `.cfg` file as clean as possible and ideally move as many cvars as possible to `.config` files.

And the question: should I open a new PR once the .cfg template is ready or rename and stick to this PR?

BystryPL:
And one more thing, would be really good if this https://github.com/etlegacy/etlegacy/pull/1745 got merged, we're currently using the well-tested `48` value for body hitbox height and the cvar is actually in our legacy1-6 configs, but I don't really want to include this cvar here in the stock configs. So that we could delete the cvar from our configs and also not include it here in the stock configs.

rmarquis:
> You mean actually forcing the cg_allowGeoIP cvar server-side? I haven't thought of that actually, and indeed that might not be a good idea from privacy point of view.

Yes. It might be a good idea to force it in some "official" tournament competition (depending on what's organized), but it is a slippery slope.

 > Well I was mainly having in mind the fact that you can see the reviving icons through walls which in my opinion is a bit too intrusive and too over-powered (usually finding a wounded teammate requires some voice communication).

Oh yes, that should definitely something to change in the current behaviour, as it goes well too far (imho) in what these objective indicators are supposed to display here. Will keep a note and try to iterate for the next release.

> And the question: should I open a new PR once the .cfg template is ready or rename and stick to this PR?

I'd say open a new PR. I am not exactly sure what you'll come with here that couldn't be achieve with configs, so that might be interesting. My first reaction would be to say "keep it simple", and for all I care the default config shipped could be a "ready for competition" config, as long as switching back to public config works correctly. 

> And one more thing, would be really good if this #1745 got merged

I'd say keep the cvar for now. But I think it should be renamed to something more global, as to allow easier testing of capsule boxes (or whatever we can come up next) in the hopefully not too distant future.

rmarquis:
Squashed/merged, thanks!

PR #1789, opened 2021-10-23
misc: add etl_server_comp.cfg template file

BystryPL:
As explained here https://github.com/etlegacy/etlegacy/pull/1787, this provides a `etl_server_comp.cfg` template file for competition servers that will complement the legacy1-6 configs. 

This template together with `legacy1-6` configs is aimed to provide a ready-to-go "package" for competition server administrators (`etl_server_comp.cfg` + `legacy1-6` configs) that would require little to no adjustment on their side and that could be just deployed on the server. My aim was to keep the `.cfg` file as clean as possible and ideally move as many cvars as possible to `.config` files. 

This is a template, so it needs filling in the empty cvars from server administrators. I tried to keep the empty cvars to a minimum.

I did this for my own purposes, as recently on ET: Legacy Scrims Discord we've been getting more and more competition servers for gathers set up by different contributors, so I thought by providing a simple server.cfg template file together with the adjusted legacy1-6 configs would help to keep consistency across all servers managed by different people (this is also needed for the 3v3 cup starting today - over 20 teams signed up) and also would make the initial process of setting the server up a lot easier, with little to no help needed from our side.

Now this could also be shipped to ETL if you consider it beneficial in some way.

### DIFF
--- a/misc/etmain/etl_server_comp.cfg
+++ b/misc/etmain/etl_server_comp.cfg
@@ -1,0 +1,84 @@
+// This is a ET: Legacy competition server template config file.
+// This template should be used together with legacy1, legacy3, legacy5 and legacy6 configs.
+// Fill in empty cvars to finish the server configuration.
+// For ET: Legacy engine servers only.
+
+
+// CLIENTS
+
+set sv_maxclients 			""						// number of players including private slots
+set sv_privateclients   	"0"           			// if set > 0, then this number of client slots will be reserved for connections 
+set sv_privatepassword  	""                		// that have "password" set to the value of "sv_privatePassword" 
+
+// PASSWORDS
+
+set g_password				""      				// set to password the server 
+set rconpassword			""       				// remote console access password 
+set refereePassword 		""    					// referee status password
+set ShoutcastPassword		""						// shoutcaster status password
+
+// NETWORK
+
+set sv_timeout 			"40"					// Seconds without any message from connected clients
+set sv_dl_timeout 			"240"					// Seconds without any message from downloading or preparing clients
+set sv_minping 			"0" 					// minimum ping required on connect (0: no minimum)
+set sv_maxping 			"0" 					// maximum ping allowed on connect (0: no maximum)
+set sv_ipMaxClients		"0"						// limits connections per IP to cvar value (0: no maximum)
+//set net_ip ""                                 	// set to override the default IPv4 ET uses
+//set net_port "27960"                          	// set to override the default port ET uses
+//set net_ip6 ""                                	// set to override the default IPv6 ET uses
+//set net_port6 "27960"                         	// set to override the default port ET uses
+
+// DOWNLOAD
+
+set sv_maxRate             "45000"                 // 10000 standard but poor for ET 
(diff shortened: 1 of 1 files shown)


rmarquis:
Thanks! Merged with some modifications in 1db48ef0d9eab124ae6f66d34ecda218382903fa.

List of changes:
* changed some categories order, as to sync with the main `etl_server.cfg` config and make it more logical and help to compare both files
* added `dedicated` and `sv_advert`. These are basic option and should always be explicitly displayed imho (also helps in case of LAN, or if people don't want to advert their server publicly)
* added explicit `sv_floodProtect` (but disabled for comp), as all security options should be explicitly displayed imho
* removed `g_ipMaxClients` which has been deprecated a while ago (replaced by `sv_ipMaxClients`)
* format cleanup (space instead of tab and other ugly madness)
* some comment cleanup as to sync with main file, and fixed a few typo here and there

Also cleaned up the main `etl_server.cfg` file, with:
* comment cleanup
* categories reorder to make it more logical and help to compare both files
* moved all mod settings to `legacy.cfg`. They shouldn't have been in the main `etl_server.cfg` to begin with.
* added a few missing vote options

I double checked it, but please have a quick look at the commit, just to ensure I didn't introduce any unwanted changes/mistakes in the `etl_server_comp.cfg` config file.

BystryPL:
> I double checked it, but please have a quick look at the commit, just to ensure I didn't introduce any unwanted changes/mistakes in the etl_server_comp.cfg config file.

It looks good! I got rid of `dedicated 2` on purpose, as it's always set in the command line anyway. Also, I didn't really like the categories order in the `etl_server.cfg`, and that's why I changed it a bit, but of course I understand you want to keep consistency with that file and that's fine.

> format cleanup (space instead of tab and other ugly madness)

It's strange, I swear to God that it looked perfectly fine at my PC and in GitHub Desktop. It's only when I pushed the commit that I noticed at GitHub that the indentation and whatnot got messed up... Even tried to fix it, but to no avail.

rmarquis:
Yes the more important aspect was to get both files mostly similar, so it's easier to compare/update. The original file was kinda a mess to begin with, so I tried to find something better but that's very subjective.

Current order is:
* HOSTNAME & MOTD
* CLIENTS
* PASSWORDS
* NETWORK
* DOWNLOAD
* LOGGING & PROTECTION
* (other stuff)
* WATCHDOG

Feel free to suggest something that makes more sense.

> It's strange, I swear to God that it looked perfectly fine at my PC and in GitHub Desktop. It's only when I pushed the commit that I noticed at GitHub that the indentation and whatnot got messed up... Even tried to fix it, but to no avail.

This happens because tabs characters are rendered as the equivalent of 2, 4 or sometimes even 8 spaces, depending on platform and user preferences. There is no real standard, and the only way to get proper text alignment in config files is by using spaces only, or using a tool that automatically does the tabs/spaces conversion (IDE, or external tools like uncrustify). We also use tabs for code indentation in ETL project, so that doesn't help to avoid mistake.

If you are using an IDE, it helps to display whitespace by default, so you can spot tabs and spaces easily.

Issue #1760, opened 2021-10-02
Activating func_invisible_user while standing inside it is impossible


Aciz:
If you're standing inside a `func_invisible_user`, it's impossible to activate it with `+activate`. Only way to activate them seems to be standing completely outside them. This severly affects gameplay on certain maps by making it unnecessarily hard to activate buttons and such (`missile_b3` being perhaps the most outrageous example, as you can probably tell by the size of this trigger in the picture).

![image](https://user-images.githubusercontent.com/14221121/135731859-152bf3f5-bf84-4916-b609-63843d94466a.png)

ryzyk-krzysiek:
It looks like the same issue as trace failing when it starts inside BBox (dynamite/uniform stealing issue). Could be that applying the same fix as there would help. I tested it briefly and it worked, but I'm not willing to test if it breaks anything else.

```
	G_TempTraceIgnoreEntities(ent);
	trap_Trace(&tr, offset, NULL, NULL, end, ent->s.number, (CONTENTS_SOLID | CONTENTS_MISSILECLIP | CONTENTS_TRIGGER));
	if (tr.startsolid && tr.entityNum == ENTITYNUM_WORLD)
	{
		vec3_t boxmins = { -10, -10, -10 };
		vec3_t boxmaxs = { 10, 10, 10 };
		trap_Trace(&tr, offset, boxmins, boxmaxs, offset, ent->s.number, (CONTENTS_SOLID | CONTENTS_MISSILECLIP | CONTENTS_TRIGGER));
	}
	G_ResetTempTraceIgnoreEnts();
```

jackeri:
Aciz reported that this is fixed so closing.

PR #1793, opened 2021-10-27
game: fix activate function failing when trace starts inside BBox, refs #1760

ryzyk-krzysiek:
refs #1760


Aranud:
Work on my end

Issue #1794, opened 2021-10-27
FT icon displayed upside-down in 2.60b client


rmarquis:
That's minor only, but the root of the issue might be deeper and other stuff might be affected, so here you go:

The fireteam icon (greed/red triangle) are displayed upside down in 2.60b client (and also ETe). They're correctly displayed in ETL client.

![2021-10-27-111315-oasis](https://user-images.githubusercontent.com/162442/139037697-ffa66191-08bc-47b8-918d-6ac9de64ec5e.jpg)

Other icons seem to be correctly displayed (see the bulb icon). I have no idea how such a feat is even possible.

PR #1795, opened 2021-10-27
mod: fix FT icon displaying upside down on 2.60b clients, refs #1794

Aciz:
The TGA had wrong field orientation (top-left), fixed by making it bottom-left.


rmarquis:
Thanks! Mystery solved.

rmarquis:
Do you mind quickly explaining where I could check this orientation info? Couldn't find anything (simply using paint.net here). So I don't introduce the same issue in new icons.

Aciz:
I personally use [XnView](https://www.xnview.com/en/xnview/) as a general image browser, and I can check it there by right clicking and checking properties. I have no clue about paint.net, but when saving a TGA image, if there is some sort of option dialog, you should check there if it allows you to set the field order. Photoshop does the correct thing by default, but for example Krita does it the wrong way, with no way to change it.

image

While saving image with XnView, bottom up needs to be checked:

image

Aciz:
Quickly looked at paint.net, doesn't seem like there is anything to change the order but it does Bottom-Left by default.

image

rmarquis:
Thank you for the detailed answer! I checked the new obj status icons with XnView, and there're correctly oriented.

I guess the FT issue was introduced because it's not possible to draw a triangle upside down by default in paint.net, so I draw it with the spike on top before flipping it.

Issue #1302, opened 2020-03-05
Exclude heavy weapons from the best accuracy award calculation


ETLegacyTeam:
Author: @alkhimey (artium)
Date: 2019-05-24
Redmine Issue: 1302, https://dev.etlegacy.com/issues/1302

---

This is a feature request for the ETL mod.

When calculating who is the player with best accuracy, exclude the hits made by heavy weapons.

Currently, someone using flamethrower will get the accuracy boosted but someone using mg will get it reduced. 
This makes the award at the end of the game meaningless.

---

Related issue(s): #478

---

Related commit(s):
96815f897a7f83b8bd7e461580bc28986322d266 - mod: exclude heavy weapons from accuracy computati..

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-05-24T19:55:26Z

---

Yeah agree on the idea. We probably should disable accuracy for all weapon that can't do headshot, -and maybe introduce a new awards for HW- (not needed, there is already a best soldier award).

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-06-12T22:21:49Z

---

Note: The "Topshot" stats windows only display bullet weapons now (since it's based on accuracy too). I don't think it is useful to have other weapon (landmine, flamethrower, artillery, etc.) listed there, but let me know if you think otherwise.
Knives are also excluded from accuracy stats (since they can't do headshot either, and they are mostly used to send players to limbo so accuracy is irrelevant here).

PR #1797, opened 2021-10-28
Show accuracy for all weapons but only include headshot weapons for rewards

Helldembez:
From the competitive community we have received feedback that people want to see their accuracy of grenades and rifle grenades. When looking at the code we noticed only weapons that can shoot headshots have their accuracy shown.

I have reverted most of this change except for the reward part. This should still be in line with the public vision and therefore should not require a cvar toggle.

I think I touched all spots, but only tested +stats / +topshots.

### DIFF
--- a/src/cgame/cg_servercmds.c
+++ b/src/cgame/cg_servercmds.c
@@ -2094,9 +2094,11 @@ void CG_topshotsParse_cmd(qboolean doBest)
 		{
 			CG_cleanName(cgs.clientinfo[cnum].name, name, 17, qfalse);
 			Q_strncpyz(ts->strWS[ts->cWeapons++],
+			           aWeaponInfo[iWeap - 1].fHasHeadShots ?
 			           va("%-12s %5.1f %4d/%-4d %5d %6d %8d  %s",
-			              aWeaponInfo[iWeap - 1].pszName,
-			              (double)acc, hits, atts, kills, deaths, headshots, name),
+			              aWeaponInfo[iWeap - 1].pszName, (double)acc, hits, atts, kills, deaths, headshots, name) :
+			           va("%-12s %5.1f %4d/%-4d %5d %6d           %s",
+			              aWeaponInfo[iWeap - 1].pszName, (double)acc, hits, atts, kills, deaths, name),
 			           sizeof(ts->strWS[0]));
 		}
 
@@ -2146,11 +2148,12 @@ void CG_parseWeaponStatsGS_cmd(void)
 	int          nRounds;
 	int          weaponMask;
 	int          skillMask, xp = 0;
-	int          totHits      = 0;
-	int          totShots     = 0;
-	int          totKills     = 0;
-	int          totDeaths    = 0;
-	int          totHeadshots = 0;
+	int          totHits             = 0;
+	int          totShots            = 0;
+	int          totKills            = 0;
+	int          totDeaths           = 0;
+	int          totHeadshots        = 0;
+	int          totHeadshotableHits = 0;
 
 	nClientID  = Q_atoi(CG_Argv(iArg++));
 	nRounds    = Q_atoi(CG_Argv(iArg++));
@@ -2191,6 +2194,11 @@ void CG_parseWeaponStatsGS_cmd(void)
 				totHits      += nHits;
 				totShots     += nShots;
 				totHeadshots += nHeadshots;
+
+				if (aWeaponInfo[i].fHasHeadShots)
+				{
+					totHeadshotableHits += nHits;
+				}
 
 				Q_strncpyz(strName, va("%-12s  ", aWeaponInfo[i].pszName), sizeof(strName));
 				if (nShots > 0 || nHits > 0)
@@ -2246,7 +2254,7 @@ void CG_parseWeaponStatsGS_cmd(void)
 			ptRatio        = (float)atof(CG_Argv(iArg++));
 
 			htRatio = (totShots == 0) ? 0.0f : (float)(totHits * 100.0f / (float)totShots);
-			hsRatio = (totHits == 0) ? 0.0f : (float)(totHeadshots * 100.0f / (float)totHits);
+			hsRatio = (totHits == 0) ? 0.0f : (float)(totHeadshots * 100.0f / (float)totHeadshotableHits);
 
 			Q_strncpyz(gs->strExtra[0], va(CG_TranslateString("Damage Given: %6d      Team Damage Given: %6d"), dmg_given, team_dmg_given), sizeof(gs->strExtra[0]));
 			Q_strncpyz(gs->strExtra[1], va(CG_TranslateString("Damage Recvd: %6d      Team Damage Recvd: %6d"), dmg_rcvd, team_dmg_rcvd), sizeof(gs->strExtra[0]));
@@ -2388,12 +2396,13 @@ void CG_parseWeaponStats_cmd(void(txt_dump) (const char *))
 	unsigned int nRounds;
 	unsigned int dwWeaponMask;
 	unsigned int dwSkillPointMask;
-	int          xp           = 0; // XP can be negative
-	int          totHits      = 0;
-	int          totShots     = 0;
-	int          totKills     = 0;
-	int          totDeaths    = 0;
-	int          totHeadshots = 0;
+	int          xp                  = 0; // XP can be negative
+	int          totHits             = 0;
+	int          totShots            = 0;
+	int          totKills            = 0;
+	int          totDeaths           = 0;
+	int          totHeadshots        = 0;
+	int          totHeadshotableHits = 0;
 
 	fFull = (qboolean)(txt_dump != CG_printWindow);
 
@@ -2456,6 +2465,11 @@ void CG_parseWeaponStats_cmd(void(txt_dump) (const char *))
 				totShots     += atts;
 				totHeadshots += headshots;
 
+				if (aWeaponInfo[i].fHasHeadShots)
+				{
+					totHeadshotableHits += hits;
+				}
+
 				Q_strncpyz(strName, va("^3%-10s: ", aWeaponInfo[i].pszName), sizeof(strName));
 				if (atts > 0 || hits > 0)
 				{
@@ -2502,7 +2516,7 @@ void CG_parseWeaponStats_cmd(void(txt_dump) (const char *))
 			ptRatio        = atof(CG_Argv(iArg++));
 
 			htRatio = (totShots == 0) ? 0.0 : (float)(totHits * 100.0 / (float)totShots);
-			hsRatio = (totHits == 0) ? 0.0 : (float)(totHeadshots * 100.0 / (float)totHits);
+			hsRatio = (totHits == 0) ? 0.0 : (float)(totHeadshots * 100.0 / (float)totHeadshotableHits);
 
 			if (!fFull)
 			{
--- a/src/cgame/cg_servercmds.c
+++ b/src/cgame/cg_servercmds.c
@@ -2191,12 +2191,12 @@ void CG_parseWeaponStatsGS_cmd(void)
 
 				totKills     += nKills;
 				totDeaths    += nDeaths;
-				totHits      += nHits;
-				totShots     += nShots;
 				totHeadshots += nHeadshots;
 
 				if (aWeaponInfo[i].fHasHeadShots)
 				{
+					totHits             += nHits;
+					totShots            += nShots;
 					totHeadshotableHits += nHits;
 				}
 
@@ -2461,12 +2461,12 @@ void CG_parseWeaponStats_cmd(void(txt_dump) (const char *))
 
 				totKills     += kills;
 				totDeaths    += deaths;
-				totHits      += hits;
-				totShots     += atts;
 				totHeadshots += headshots;
 
 				if (aWeaponInfo[i].fHasHeadShots)
 				{
+					totHits             += hits;
+					totShots            += atts;
 					totHeadshotableHits += hits;
 				}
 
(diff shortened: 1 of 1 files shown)


rmarquis:
I don't have much time to comment on this tonight, but it's overall a bad idea: it messes other useful overall stats. I'll give some details tomorrow.
What could be done instead is using another metric computed separately, just not together with accuracy.

Helldembez:
My attempt was to not influence the overall accuracy stats. At least in rewards. Other overall stats should include all weapons as this is part of the word "overall". If a number is needed to only indicate overall accuracy of headshot weapons it should be introduced as separate overall stat in my opinion. 

The messing actually happens if you don't include all weapons since it becomes an altered value.. You could decide when calculating overall acc to limit to 100% per weapon in the calculation, but individual weapon accuracy should show their actual accuracy.

rmarquis:
Just a little note: I'm very busy today, but I'll try to write something in the evening (accuracy is actually quite a complex topic). That last change seems to be a step in the good direction though.

rmarquis:
Let me start by saying I feel the way we represent accuracy in Legacy kinda sucks, but at the very same time it's also the less sucky it's ever been.
I completely understand that some players might feel that grenades (or more generally other splash damage weapons) aren't outputting the numbers they deserve, compared to their usual bullet weapons. But the average player doesn't understand much about stats, or at least never gave too much thought about it. Note I'd use below "splash damage weapons" or SDW as a general term, be it grenade, mines or airtrike since the same concept applies here (and also to flamethrower somehow, which kinda behave the same but in a more amplified manner). 

To understand the current status, the first important step is to define the reason we want accuracy stats in the first place. Do we want some pretty number after a match because why not? Of course not, we actually want some statistical metrics because they are useful to us. I see two major reasons we want accuracy statistics:
* accuracy can be monitored over time to see how much we (as player) progress over time, how good our aim becomes, how long we can keep it more or less consistent before we can't focus as much, or how a bad day affect us, etc,
* accuracy can be compared across players to determine how good the aim of a player is. Someone with 10% accuracy might be a noob, 40%+ might indicate a seasoned player, 50 and above tend to be seen suspiciously excellent.

Each weapon has its own stats, and the global accuracy stat (and HS percentage) is thus an aggregated metric that aims to provide a single, fast and easy way to get an idea of what the quality of the aim of the player is.

And now the first question arises: what is the "accuracy" of splash damage weapons, exactly? Let's imagine a player throws 10 grenades, and misses 9 times completely while the 10th successfully hits the enemy. Is the accuracy 10%? Intuitively, most people would agree it is indeed 10%. For each bullet weapons, your accuracy for 10 shots and 9 missed would indeed be 10%, so there shouldn't be different. And yet accuracy can be 10%, 50%, 100%, or 300%. In fact, *nobody* knows what the accuracy of SDW is when computed the same way as bullet weapon, because SDW can make more than a single hit, unlike bullet weapon. We literally had report of accuracy of 5-digit percentage for some weapon (flamethrower with 12000% accuracy) which is nothing short of ridiculous.

From there, you should start to understand why computing accuracy of grenade is difficult, and that using that value in global accuracy only makes the metric more useless than actually useful. You wouldn't even add information, but you'd do something like adding random noise in the result, rendering it harder to interpret and use.

So while I understand your original intent here, I can't disagree more with you when you say "if you don't include all weapons, then it becomes an altered value".

This is wrong for the same reason you can't lump oranges with apples, and pretend they are apple while evaluating the overall quality of your apples. It's not going to work. Incidentally, previous iterations of Legacy had what you proposed above, with an attempt at capping stats. It's indeed not working, because while the individual value displayed is 100% at most, all the rest of the stats calculation is *distorted*.

Simply said, accuracy as a metric for splash damage weapon doesn't make sense, especially not when computed the same way as bullet weapons. It is a least wholly incomplete, and like in IRL statistics, "accuracy" is quite a vague metric that is almost never used by itself alone (f.e. we use other metrics such as "precision", "recall", "balanced accuracy" if the data is parse, "F1-score", ROC curve and dozens more depending on the field and type of data).

At the end of the line, we have to remember what is the goal of the accuracy metric. If adding information makes is more useless, then the way that information is processed might be completely wrong. "Accuracy" itself is a misnomer when taking splash damage weapons into account, and chances are high that different people might disagree on what it is, or should be.

*If* accuracy were to be computed for them, then I'd personally define it as "number of launch that hit *at least* one enemy"/"total number of launch" (so 9 miss on 10 launches would be consistently 10%), which is something that is more complex to implement than we ever did. At the same time, that number alone wouldn't provide a good picture of what the result of that accuracy actually is, and we'd need to add a new metric (f.e., "average hits per launch") which would be a some positive value, not limited to a percentage like accuracy is supposed to be. And even if done that way, I am not sure if would be wise to lump it together with global accuracy in regards to the purpose of the accuracy metric (SDW are still oranges at the end of the day).

I hope the above is clear enough, and that you see that the topic of accuracy is quite deeper than what it initially seem, and that this post provides a good base to improve metrics we display in Legacy rather than the opposite. There's room for improvement, but seemingly not in that "easy" direction of going back to what we got previously. There is also a whole other chapter on how accuracy for bullet weapons is computed, because there are different ways to do it (huh), but we'll keep that other topic for another time.

zenixje:
I first thought, that it was a bug, because it seems to weird not to display accuracy of explosive weapons. Getting high accuracy with rifle, nades, arty, airstrikes or panzer actually takes skill and timing, perhaps even more than just aiming good with a mp40. Some rifles barely get above 40% riflenade accuracy, but skilled ones go above 80% and above that, and you can actually see if you are improving. 

As you said, the average player does not really care about stats, but competitive players really like to compare stats.  We don't care about the overall metric or the science behind it, we just want too see all accuracy stats for all weapons in topshots or individual stats. It does not matter either if the metric is not 100% correct, if we can compare it to other competitive players, it's good enough.

rmarquis:
Yes, the fact that is missing is counterintuitive and exactly why I am not pleased with the current status. However, I'd largely prefer it to not be there than to be wrong an irrelevant.

> It does not matter either if the metric is not 100% correct, if we can compare it to other competitive players, it's good enough.

Yes, this is exactly my point. If it is simply computed the same way bullet weapons are (hits/shots), it is not only "not 100% correct", but also completely wrong and incomparable.

Don't take it the wrong way, but you're perfectly illustrating the "average player" regarding to stats here. This has nothing to do with being a comp player.

zenixje:
Would it not be better to remove accuracy all over the board than? And show only the amount of bullets you hit and missed? Because the current metric for bullet weapons is also wrong, you can't really compare the bullet weapons, somebody with a garand will generally get higher acc than somebody with a mp40. Also playing style matters here, some players only play close combat and don't shoot a lot of bullets, but if somebody shoots lower acc but does a lot more damage on distance, aiming wise that is a lot harder.  Should that not be taken into account?

I am just saying, getting the right metric to account for all the different situations is pretty much an impossible task, because there are too just too many variables.  The reason this was implemented was to remove the explosive weapons from the awards. That that individual acc got removed is a side-effect and that should be fixed. https://github.com/etlegacy/etlegacy/issues/1302 We are letting you know that we think otherwise.

rmarquis:
Did you... did you even read my post above? There's some suggestion to actually implement individual stats back, just not by reverting to the previous behaviour that doesn't make any sense.

You're however right that having an single aggregated value to represent aim accuracy is not perfect (one might say "accurate" ha), but that's not a reason to make the stats wrong on purpose. We should also fix shot stats in regards to gibs too.

zenixje:
I have seen it, still don't see a reason why we can't bring it back for now, because the reason it was removed was because of the messing with the overal accuracy stats, but that is fixed. In your post you have some good ideas to improve it, but it doesn't mean it should not be fixed at this moment. Because fixing accuracy all across the board seems to be a big topic that will take time to fix. I don't know if that is worth the time to be honest.

This is not just me speaking, but I am also speaking on behave of the competitive community since I am a moderator there. I am posting here because we received several complaints from players about this.

rmarquis:
Fixing individual stats themselves might be not too hard, depending on what solution we settle on to display it. My suggestion is to first make it consistent with the number of misses, regardless of the number of enemies hit (10 grenades launched with 9 missed = 10%).

If that idea seems sound, then I'd suggest we prioritize this, so players will have their individual splash damage accuracy stats as soon as possible. The rest of the stats issues can then be incrementally fixed.

There won't be a snapshot release anyway for testing on gathers/scrims at least until the end of the cup anyway, so there's time to do this first step. I also happen to like the topic, so I don't mind working on it asap (just need to get that damn hotfix out first).

Helldembez:
> And now the first question arises: what is the "accuracy" of splash damage weapons, exactly? Let's imagine a player throws 10 grenades, and misses 9 times completely while the 10th successfully hits the enemy. Is the accuracy 10%? Intuitively, most people would agree it is indeed 10%. For each bullet weapons, your accuracy for 10 shots and 9 missed would indeed be 10%, so there shouldn't be different. And yet accuracy can be 10%, 50%, 100%, or 300%. In fact, nobody knows what the accuracy of SDW is when computed the same way as bullet weapon, because SDW can make more than a single hit, unlike bullet weapon. We literally had report of accuracy of 5-digit percentage for some weapon (flamethrower with 12000% accuracy) which is nothing short of ridiculous.

I understand your point. So I have a suggestion. Keep individual %'s of weapons visible. Introduce different total accuracy counters. 

1. Aggregated accuracy of SMGs. (The current total accuracy)
2. Headshot accuracy also only of SMGs (obviously)
3. Aggregated accuracy of non-SMGs ("spam" accuracy)
4. Aggregated accuracy of all weapons (Actual total accuracy)
5. Accuracy award is given for highest SMG accuracy. (Can also introduce new award for top non SMG acc)

This way all accuracy besides point 4 makes sense in their own context. For completeness and people who are curious point 4 is still interesting on a personal level. (Though this might be removed later depending on feedback of actual players)

> My suggestion is to first make it consistent with the number of misses, regardless of the number of enemies hit (10 grenades launched with 9 missed = 10%).

Not a fan. Splash damage weapons should represent what they are made to do. SPLASH. Who cares if shot 1 and 2 missed if they hit a triple for 100% acc. They know they missed because it says 1/3 hits. I think this is a fine road in the middle to prevent introducing many different numbers to exactly dissect what shot did what etc.

> So while I understand your original intent here, I can't disagree more with you when you say "if you don't include all weapons, then it becomes an altered value".

With this I was already hinting at the 5 categories I listed above. As in the real TOTAL accuracy is everything combined. The categories should just have corresponding names that indicate what they are.

rmarquis:
> Not a fan. Splash damage weapons should represent what they are made to do. SPLASH. Who cares if shot 1 and 2 missed if they hit a triple for 100% acc.

An accuracy above 100% doesn't make sense to begin with (nor from a mathematical perspective, nor intuitively). It's a percentage, so the value should be between 0 and 1.

But I do see your point about representing splash, thus my suggestion to have an average "splash value" (referred by "average hits per launch" above).

About your list:
* Point 1: By SMG, do you mean only SMG (Thomson/MP40) here, or all bullet weapon that can do HS? There's a whole category of bullet weapons that can do headshot, so I suggest we regroup all weapons of that category together (so we can compute HS% easily too, in your point 2)
* Point 3 could be the "splash value", in addition to the individual accuracy (between 0 and 1)
* Point 4: okay, but makes more sense if we do indeed compute accuracy consistently. Along Point 3, this can actually work!
* Point 5; Could have accuracy award, and a new "splash" award too?

Thanks for the suggestion! Finally feeling we're making progress here.
(do note I'm also trying to make it fit for all weapons of the game, rather than having a comp-only perspective, if that make sense).

Helldembez:
> An accuracy above 100% doesn't make sense to begin with (nor from a mathematical perspective, nor intuitively). It's a percentage, so the value should be between 0 and 1.
> 
> But I do see your point about representing splash, thus my suggestion to have an average "splash value" (referred by "average hits per launch" above).

So you mean the word "accuracy" should always be represented in a percentage between 0 and 100 and to keep the 300% numbers on splash weapons there will be a new to be determined name, like "splash value"? This "splash value" will be shown for each individual weapon? 

The only thing I am thinking about going forward is how the layout of the stats will be with this. Adding this as a separate column does not feel like the right solution. Maybe some mockups or poc implementations before we agree on the layout(can be in ASCII art as well)?

> Point 1: By SMG, do you mean only SMG (Thomson/MP40) here, or all bullet weapon that can do HS? There's a whole category of bullet weapons that can do headshot, so I suggest we regroup all weapons of that category together (so we can compute HS% easily too, in your point 2)

Yes, I do mean all bullet weapon that can do HS. I was just lazy to write it that way. The current code has the `fHasHeadShots` boolean that is being used to group this category. Also headshot % is already calculated and shown and even rewarded. I just thought I should mentioned it in the list for completeness sake.

rmarquis:
> So you mean the word "accuracy" should always be represented in a percentage between 0 and 100 and to keep the 300% numbers on splash weapons there will be a new to be determined name, like "splash value"? This "splash value" will be shown for each individual weapon?

Yes, exactly.

Like you, my biggest worry right now is to find enough space to display the information. That's a minor issue, but that always takes lot of time do properly. Drawing a mockup with an image editor might help too (but feel free to do ASCII :D).

> Yes, I do mean all bullet weapon that can do HS.

All right, perfect then!

Helldembez:
In that case I will revert the aggregation of all weapons into the current Total accuracy value. So this PR can be merged and create an issue with a link to the discussion of this PR. So future PR's regarding the total value discussion can be linked to that issue and further discussion can be done centralized.

Helldembez:
@rmarquis / anyone with rights. Are we not merging this PR while we work on the newly created issues?

rmarquis:
Apologies for the late reply, I've not had much time due to some unplanned IRL events - thx to Ara for merging. As a quick note: current status is WIP, still need to adjust code for airstrike/arty/flamethrower, and I haven't touched UI yet.

Issue #1796, opened 2021-10-28
Revise sniping experience


Helldembez:
Within the Competitive community we have received feedback regarding sniping. Right now it is almost impossible to snipe without standing still and not get hit. User awAz also made a comparison of the Legacy behaviour to ETPro.

See: 
ETL: https://streamable.com/su9k41
ETPro: https://streamable.com/2txfx7

Personally I think you should at least be able to scope and shoot relatively fast after moving. While getting hit should just provide some up recoil when scoped.

Aranud suggested the following:
Aranud: We could cap the move speed while scoping and remove the unscoping behaviour when speed is too high.

x0rnn:
I'll add my opinion here instead of opening another issue:

1. Movement while scoped should be capped to crouching speed, 80 UPS, that way you can still walk while scoped but not gain a speed advantage, it would essentially be like crouch sniping before.

2. You should unscope if swimming too fast, same as before, or at least cap it same as when walking.

https://github.com/etlegacy/etlegacy/pull/1798

x0rnn:
Rashomon will add his opinion and videos about re-adding binoc sniping (weapalt + fire with binoculars) soon, which Aranud said got removed (fixed) 4 years ago in 2.76, in the meantime I'd like to propose unscoping the weapon when holding the sprint key instead of currently capping the speed to 80 (only when sprinting, keep the 80 cap when walking), to avoid having to weapalt to unscope.

RASHOMONN:
Here I am, so: Here are a 2 frag clips (on Jaymod). Sorry for the watermarks and very basic editing, first time doing that.

1st clip: https://streamable.com/z34lfp     Here you can see the efficiency of quickly switching between binocs and sniper scope while jumping: I can be farther from the action and way harder to target. I'm using double jump except for the last hit but it's the same reasoning with single jump. I believe the in-air time with the sniper scope has been patched already, allowing for more time (don't know what the exact duration is on jaymod). 

2nd clip: https://streamable.com/5ra6jg   Here, this clip showcases the absolute necessity of being able to quickly switch between your binocs and your sniper scope (it is also needed in the first clip by the way). At the moment, when clicking mouse2 while using binocs on a Legacy server, nothing happens at all. 
[...]

PR #1798, opened 2021-10-29
mod: cap speed on weapon scoped, refs #1796

Aranud:
* Cap to half the run speed (128, initial speed limit where the weapon would be automaticly unscoped
* Don't allow sprint while scoped
* Add spread while mid-air / jump and scoped


BystryPL:
Let's see what happens!

Aranud:
Indeed ! Game breaker or good balance change ... let see

PR #1801, opened 2021-10-31
app: fixed version string with tags

rmarquis:
This is a simpler way to get either the complete version string for a snapshot (`v2.78.0-96-g96f9484`), or just the tag for release version (`v2.78.1`).

Do note the `v` is included in the version string already.

@rafal1137 This is untested on Android build, so feel free to have a closer look. I only tested how git outputs the string on regular commit or commit with a tag.


rafal1137 (approved):
I found no reasons to not approve this one. Tested on my local and it does work correctly

Issue #1802, opened 2021-10-31
Operating system taskbar panel appears over game window when in "Windowed Fullscreen" mode


porkbytes:
On GNU/Linux (OpenSUSE in my case) and KDE, when running the game in "Windowed Fullscreen" mode, the taskbar panel appears over the game window. This obscures part of the game window. Ideally, the game window should always appear on top of any operating system elements such as taskbars.

I suspect this could affect other window managers and maybe even other operating systems, but I have only been able to test on my Linux workstation.

rmarquis:
Instead of the proposed `SDL_WINDOW_ALWAYS_ON_TOP` which always catches the focus if present (breaking alt-tab), we might need to use the `SDL_SetWindowAlwaysOnTop()` function dynamically.

This requires SDL >= 2.0.16 (we use a mix of SDL 2.0.14 and 2.0.16 at the moment, see #1762).

porkbytes:
I'll keep investigating solutions. Is alt-tab behavior the only thing I should be testing for or did that flag break anything else?

rmarquis:
I only noticed this particular issue, but there might be others.

PR #1803, opened 2021-11-01
glimp: made windowed fullscreen always on top, refs #1802

porkbytes:
Adding the `SDL_WINDOW_ALWAYS_ON_TOP` flag ensures nothing else will appear above the game window.

Refs #1802

### DIFF
--- a/src/sdl/sdl_glimp.c
+++ b/src/sdl/sdl_glimp.c
@@ -749,7 +749,7 @@ static int GLimp_SetMode(glconfig_t *glConfig, int mode, qboolean fullscreen, qb
 	{
 		if (noborder)
 		{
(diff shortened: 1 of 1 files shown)


rmarquis:
Yes. According to the [doc](https://wiki.libsdl.org/SDL_WindowFlags), this flag only has an effect on X11 system.
Also tested on Windows just in case, and there isn't any side effect.

Thanks for the fix!

rmarquis:
I'm reverting the merge. I botched my testing somehow, but on further inspection it does impede on the behaviour of windowed full screen (on windows, and probably others platforms too).

Note to self: this is again a good reminder not to merge stuff hastily before a release...
Note to self2: Don't trust SDL doc blindly.

PR #1805, opened 2021-11-02
client: added cl_slashCommand cvar so the user can choose whether a

JumptoHistory:
slash is required before cmds in the console
Couldn't this improve user experience? Let me know if it could cause trouble or I need to fix something.

### DIFF
--- a/src/client/cl_console.c
+++ b/src/client/cl_console.c
@@ -601,7 +601,14 @@ void Con_DrawInput(void)
 
 	re.SetColor(con.color);
 
-	SCR_DrawSmallChar(SMALLCHAR_WIDTH, y, ']');
+	if (SLASH_COMMAND)
+	{
+		SCR_DrawSmallChar(SMALLCHAR_WIDTH, y, ']');
+	}
+	else
+	{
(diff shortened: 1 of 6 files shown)


ensiform:
Why changing the ] to \ for input prompt?

Also might be better to have the cvar init in common.c and defined in qcommon.h so as to avoid it being NULL ever.

JumptoHistory:
Well, I thought changing the prompt would make it easier to tell which mode you're on.
I'll look into it about the cvar init and definition.

rmarquis:
This looks like an over-engineered solution to a problem that doesn't really exist.
I never ever type the slash because autocompletion does it.

JumptoHistory:
Okay, I used to forget adding a slash before I implemented this in my custom build though.

ensiform:
If you don't put slash, it turns command into chat, does it not?

JumptoHistory:
Yeah, it's a bit embarrassing; not that I've given my referee password out in the global chat or anything but still.

rmarquis:
Closing, because not needed. Just need to make use of completion, which is faster to use anyway.

PR #1820, opened 2021-11-10
game: allow shoutcasters to use setviewpos

Aciz:
Allow shoutcasters to use `setviewpos` command so they can make binds/configs for fixed camera angles.

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -3647,7 +3647,7 @@ void Cmd_SetViewpos_f(gentity_t *ent)
 	char   buffer[MAX_TOKEN_CHARS];
 	int    i;
 
-	if (!g_cheats.integer)
+	if (!g_cheats.integer && !ent->client->sess.shoutcaster)
 	{
 		trap_SendServerCommand(ent - g_entities, va("print \"Cheats are not enabled on this server.\n\""));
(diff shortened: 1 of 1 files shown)


ensiform:
Suggestion: You may also want to make it so shoutcasters can use exact location, because when players use it, it is a teleport and you dont actually respawn in the exact spot.

Aciz:
> Suggestion: You may also want to make it so shoutcasters can use exact location, because when players use it, it is a teleport and you dont actually respawn in the exact spot.

Not sure if I follow you here, the command takes absolute position, not relative? And what does it have to do with respawning? The intent here is to allow shoutcasters to have fixed camera angles by using this command to set them to that location.

ensiform:
It teleports the player and you move forward a bit from the actual position you give, at least this is how it normally works.

Aciz:
I've never seen that it would move you forward a bit, only confusion regarding the position is that `viewpos` gives you viewheight, while `setviewpos` sets your origin, so there's a bit of z-pos discrepancy as a result (but we have added an extension to the command to account for this, so you can match the exact coordinates that you get from `viewpos` command).

Aciz:
All the above confusion was due to Q3 adding forward velocity after teleportation.

PR #1821, opened 2021-11-10
game: add cvar to swap suicide anims to regular death anims

Aciz:
This is meant as a temporary test to gather feedback, adds `g_altSuicideAnims` to use regular death animations as suicide animations. Many people believe this will lead to less confusion, so I think it's worth trying this out.

Edit: sorry about the commit message, brain froze while doing this apparently.


Aranud:
Let have a try and see what happen

Aciz:
Feedback regarding this has been very positive. Is anyone opposed to dropping the cvar and making this the default behavior?

PR #1822, opened 2021-11-10
cgame: move popup shadow from althud flags to a seperate cvar

Aciz:
I don't think many people realize that drawing shadow on popups is even a possibility, I didn't know about this either until like a month ago, especially since it's not even included in the options menu for althud flags. So therefore, removed `cg_althudFlags 8` in favor of `cg_popupShadow`, I don't really understand why it was hidden there in the first place.

### DIFF
--- a/etmain/ui/options_customise_hud_advanced.menu
+++ b/etmain/ui/options_customise_hud_advanced.menu
@@ -30,16 +30,17 @@ menuDef {
 
 // Popup //
 #define POPUP_Y 32
-	SUBWINDOW( 6, POPUP_Y, (SUBWINDOW_WIDTH_L), 54, _("POPUPS AND BANNER") )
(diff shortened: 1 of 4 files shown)


BystryPL:
neat

jackeri:
I would prefer moving away from the whole cg_althudflags thing (I originally added it way back when just to try to entice etpro players) but those flags / customizations / etc should be on the specific hud components. Not on global level imho.

Aciz:
Yes, the whole HUD system should be rewrote ideally and get rid of as many HUD related cvars, and do everything through custom HUDs with generic parameters (x/y/w/h/alpha/shadow etc) and leave element-specific customization to cvars. But this requires a user-friendly way to edit HUDs, which is not what the current system with custom HUDs is.

jackeri:
Which is why there needs to be an effort to add a basic hud editor (like in other mods / games).

Issue #1806, opened 2021-11-03
Directional sound issue with SDL2 backend


BystryPL:
From user "Abject" on ET: Legacy Competitive Discord server:

> Hey,
> 
> is SDL2 sound expected to be default? I've been playing few games so far and every time I was so confused with things being right next to me. Today I find out about SDL and OPENal settings and second one feels way more normal.
> SDL2
> https://streamable.com/csr3ef
> OPENAL
> https://streamable.com/1wualx
> 
> btw. could you please let me know what is the command for this settings? I want to set openal permanent for now.
>

rmarquis:
OpenAL used to be the default, but SDL2 has been reverted back as the default sound system in 2.78.0, as OpenAL has some issues in a few specific cases (some environmental audio too loud).

There might be some subjectivity here, but I also find OpenAL better for gameplay (more "natural" and immersive, and easier to evaluate distance of noise sources compared to SDL which sounds "flat"). OpenAL also has surround sound support, and a few bugfixes related to OpenAL were applied in 2.78.0 too.

The cvar is `s_initsound` , available in `Options > Settings > Audio > Sound system`.

Closing because that's not really an issue.

Aciz:
Ummh, what? There's clearly an issue here, the guy is throwing a grenade in front of him, yet the explosion comes from the left?

rmarquis:
Sorry, thought he just wanted OpenAL back. I didn't actually watched the video, my bad (long story short, no sound on my machine right now because driver update).

Aciz:
Another example, metal impact sound playing on the right with SDL2 when shooting a surface in front of me.

SDL2:
https://streamable.com/8fexha

OpenAL:
https://streamable.com/750bjm

Aciz:
Seems like it's always playing some sounds as if they are coming from East.

https://streamable.com/lyya28

Edit: this is the case in the Ice clip in the starting post as well.

rmarquis:
Is this happening on Windows, or Linux? (just to rule out SDL 2.0.14/16).

Aciz:
Windows for me, so 2.0.14.

Aciz:
Eh, the clips are actually from 2.0.16, as I tested with my mapping installation which I haven't updated yet (running 2.78.0), so that is actually 2.0.16. But, it does happen with 2.78.1 too, so 2.0.14 is also affected.

rmarquis:
All right, fixed my sound drivers.

[...]

PR #1823, opened 2021-11-12
client: fix directional sound issue with SDL2 backend, refs #1806

ryzyk-krzysiek:
The issue was transposition of `listener_axis` on `y` axis which later affected calculation of `rscale` and `lscale`.

https://github.com/etlegacy/etlegacy/blob/a1676ec0f54866153b3bcc20111b783ed5dfe26e/src/client/snd_dma.c#L571-L574

This could be done instead: 
```
listener_axis[0][0] = axis[0][0];
listener_axis[0][1] = axis[1][0];
listener_axis[0][2] = axis[2][0];
listener_axis[2][0] = axis[0][2];
listener_axis[2][1] = axis[1][2];
listener_axis[2][2] = axis[2][2];
```

But from what I tested this doesn't produce anything really, the `rscale` and `lscale` stays the same no matter with or without it, so there doesn't seem to be any benefit to doing it.

refs #1806

### DIFF
--- a/src/client/snd_dma.c
+++ b/src/client/snd_dma.c
@@ -1421,7 +1421,6 @@ void S_Base_Respatialize(int entnum, const vec3_t head, vec3_t axis[3], int inwa
 	VectorCopy(axis[0], listener_axis[0]);
 	VectorCopy(axis[1], listener_axis[1]);
 	VectorCopy(axis[2], listener_axis[2]);
-	mat3_transpose(listener_axis, listener_axis);
 
 	// update spatialization for dynamic sounds
 	ch = s_channels;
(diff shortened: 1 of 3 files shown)


Aciz:
Doesn't seem to fix anything for me, impact sounds are still coming from East.

https://streamable.com/gfspdd

ryzyk-krzysiek:
Added vanilla `vec3_rotate` function that should fix it. I was testing with it without realizing it was affecting the results.

jackeri:
While you are at it, could please add a brief documentation to the method so coders now that THAT function is the one that generates inverted results.

Issue #1777, opened 2021-10-13
Client crashes when dynamite explosives 


Minimuppet:
Large amount of clients are crashing to their desktop on certain map when dynamite explodes. eg erderberg when blowing second gun, goldrush I think second barrier and aldernest when controls are blown.

Helldembez:
Over on discord people tried to reproduce it by playing on the server, but they did not encounter it. This at least means it's not so frequent. This doesn't mean that it shouldn't be fixed asap, but at least gives some room for investigation. 

My first thought is this has something to do with dynamite chaining? However, on the competitive side I haven't seen any reports yet. Aciz will try to play through debugging client. If a consistent reproduction path can be found please let us know.

rmarquis:
I can't reproduce this issue when planting several dyna. I kinda doubt the issue is related to the latest dyna chaining changes, as the behaviour hasn't been really modified (extra dyna removed instead of exploded by default).

This however makes me think about the infamous oasis wall bug we had several years ago (#627), which was related to a sound issue when compiled with a specific compiler. For 2.78, the default sound engine has been changed from OpenAL to SDL again. Might be something here.

ryzyk-krzysiek:
First reports of something like this occurring were like 2-3 months ago, so changes to dynamite chaining shouldn't matter. Also I believe if it was a server side issue either server would crash or everyone on the server would.

This crashing never happens on scrim servers too and there are played all 3 mentioned maps regularly. My only suspicion before was `cg_drawEnvAwareness` as it is forced off on scrims, but recent changes probably exclude it as being the probable cause of crash.

rmarquis:
Potentially fixed by 0ff8a24065d368423758a421ca0b285a4c7334e7.

randomek:
I had it just 2 times but only after upgrade to 2.78 and it's very hard to reproduce. In logs it looks like normal: "Client shutdown" but today someone who was playing with me ask if someone experienced that issue exactly at that same time when my Client shutdown. Today it was 1 second after winning the game and dyna on fuel dump blow up. But first time wasn't with dyna related.

PR #1824, opened 2021-11-13
mod: fix client crash when dynamite explodes, refs #1777

ryzyk-krzysiek:
This reverts commit 31ee38561cafa4f77ea08cb3df98800c027ec1d9.

The crash occurs here: (demo: [2021-11-12-194858-etl_adlernest_v1 CRASH.zip](https://github.com/etlegacy/etlegacy/files/7530981/2021-11-12-194858-etl_adlernest_v1.CRASH.zip))

https://github.com/etlegacy/etlegacy/blob/a1676ec0f54866153b3bcc20111b783ed5dfe26e/src/cgame/cg_commandmap.c#L954

`effect1Time` is some huge number because the `cent` is wrong. Instead of being an actual objective related ent it's some event. In the case of the provided demo it is `EV_MG42BULLET_HIT_WALL`. What I think is happening is when there is a lot of players and a lot of entities, some are getting overwritten and `cent = &cg_entities[mEnt->data];` returns wrong ent.

I was able to reproduce the issue pretty consistently, with the fix after 20 tries I was unable to do it.

refs #1777

### DIFF
--- a/src/game/g_utils.c
+++ b/src/game/g_utils.c
@@ -909,10 +909,38 @@ void G_FreeEntity(gentity_t *ent)
 		return;
 	}
 
-	Com_Memset(ent, 0, sizeof(*ent));
-	ent->classname = "freed";
-	ent->freetime  = level.time;
-	ent->inuse     = qfalse;
+	// this tiny hack fixes level.num_entities rapidly reaching MAX_GENTITIES-1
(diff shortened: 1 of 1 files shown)


rmarquis:
Good catch! The `l_free` cvar and related comment can be removed too.

Issue #1825, opened 2021-11-14
Vote gets cancelled automatically near the end of the round


BystryPL:
An issue brought up by user hazz on #feedback channel on ET: Legacy Competitive Discord which I'm now transferring over here:

> this is about votes being automatically cancelled with the message in the screenshot when you vote something near the end of the round
> you can't vote cointoss already + you can't skip garbage time (so e.g. people end up attacking supply first stage with 3min left when when result is already clear and  people just wanna move to next map)

https://streamable.com/nhy3x0 

> it just cancels immediately and says that 3 people voted no
> but it only happens when there's not much time left which makes me think it was made like this on purpose

Issue #1386, opened 2020-03-05
Tallies votes results are confusing


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2020-01-07
Redmine Issue: 1386, https://dev.etlegacy.com/issues/1386

---

Tallies votes results are confusing, because the requirements for a vote to be accepted are not explicit: a minimum percentage of voters need to vote or a positive result will be dismissed.

Improve the UI with the required information, so people understand why a "successful" vote fails.

Aranud:
Need feedback to know if the message look too long or not !

PR #1826, opened 2021-11-14
game: fix vote timeout logic, refs #1386 #1825

Aciz:
Vote timeout was using wrong logic, resulting in votes being automatically timed out when there was less than 30s of round remaining.

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -3223,9 +3223,9 @@ qboolean Cmd_CallVote_f(gentity_t *ent, unsigned int dwCommand, qboolean fRefCom
 	{
 		level.voteInfo.voteTime = level.warmupTime - VOTE_TIME;
 	}
-	else if (g_gamestate.integer == GS_PLAYING && (level.startTime + (g_timelimit.integer * 60000) - level.time < VOTE_TIME))
+	else if (g_gamestate.integer == GS_PLAYING && (level.startTime + (g_timelimit.value * 60000) - level.time < VOTE_TIME))
 	{
-		level.voteInfo.voteTime = level.startTime + (g_timelimit.integer * 60000) - VOTE_TIME;
+		level.voteInfo.voteTime = level.startTime + (g_timelimit.value * 60000) - VOTE_TIME;
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


rmarquis:
I am not sure if this actually changed anything much (`a - b > c` is equivalent to `a > b + c`), apart from that `=` sign. 
But `level.voteInfo.voteTime` has been changed in bacade7892a1a2c1563c6897897e2854e17b6a6e, maybe there has been some side effect introduced here.

Aciz:
Yes, that change precisely is what made the logic fail, as `level.voteInfo.voteTime` was not necessarily the same as `level.time` anymore at the time the vote was called. This made the previous logic instantly evaluate to true when round/warmup had less than `VOTE_TIME` left. With the new logic, `level.voteInfo.voteTime` not being the same as `level.time` at the time of voting does not matter anymore, as we add `VOTE_TIME` to it.

rmarquis:
Yes, but my point here is that your change is equivalent to the old way. Algebra tells me the code logic didn't change much (only the `>=` sign to `>`, which shouldn't be relevant I think?).

`level.time - level.voteInfo.voteTime > VOTE_TIME` is the same as `level.time > level.voteInfo.voteTime + VOTE_TIME`

Am I missing something obvious here, or is my algebra that bad?

Aciz:
You are indeed correct, and my testing appeared to be faulty earlier today.

The real reason appears to be that we are using an integer value for `g_timelimit` (aka the cvar `timelimit`), but as we know, it is not necessarily an integer. We then multiply this by `60000` so the float to int "conversion", which means just dropping the decimal, can result in a massively incorrect `level.voteInfo.voteTime`.

https://github.com/etlegacy/etlegacy/blob/a0216cf88fb5269959d91c92bf5f751210f3ef8d/src/game/g_cmds.c#L3226-L3229

For example:

Time is set to 5:54 on 1st round of stopwatch. 5:54 -> 5.9 `timelimit`, 5.9 to int -> 5 * 60 000 = 300 000
Where as if timelimit was a float, the resulting time would be 5.9 * 60 000 = 354 000
`level.voteInfo.voteTime` is now 54s too early, and the timeout logic triggers.

image

Aranud:
neat

rmarquis:
That was nasty. Good job!

PR #1831, opened 2021-11-16
game: cancel start match vote if countdown starts

Aciz:
Cancel start match vote if all players ready up, and the countdown starts.


BystryPL:
neat

jackeri:
neat

PR #1833, opened 2021-11-20
mod: add g_extendedprone - prone animation transitions through crouching

ryzyk-krzysiek:
https://streamable.com/64dbq2

### DIFF
--- a/etmain/configs/defaultpublic.config
+++ b/etmain/configs/defaultpublic.config
@@ -33,6 +33,7 @@ init
 	set team_riflegrenades "1"
 
 	set g_pronedelay "0"
+	set g_extendedprone "1"
 	set g_shove "60"
 
 	set g_antiwarp "1"
--- a/etmain/configs/defaultpublic.config
+++ b/etmain/configs/defaultpublic.config
@@ -33,7 +33,6 @@ init
 	set team_riflegrenades "1"
 
 	set g_pronedelay "0"
-	set g_extendedprone "1"
 	set g_shove "60"
 
 	set g_antiwarp "1"
--- a/etmain/configs/legacy1.config
+++ b/etmain/configs/legacy1.config
@@ -126,6 +126,7 @@ init
 	setl g_campaignfile ""
 
 	setl g_pronedelay "1"
+	setl g_extendedprone "1"
 
 	// set this to disable wolfadmin or enable custom Lua files
 	//setl lua_modules ""
--- a/etmain/configs/legacy1.config
+++ b/etmain/configs/legacy1.config
@@ -126,7 +126,6 @@ init
 	setl g_campaignfile ""
 
 	setl g_pronedelay "1"
-	setl g_extendedprone "1"
 
 	// set this to disable wolfadmin or enable custom Lua files
 	//setl lua_modules ""
--- a/etmain/configs/legacy3.config
+++ b/etmain/configs/legacy3.config
@@ -127,6 +127,7 @@ init
 	setl g_campaignfile ""
 
 	setl g_pronedelay "1"
+	setl g_extendedprone "1"
 
 	// set this to disable wolfadmin or enable custom Lua files
 	//setl lua_modules ""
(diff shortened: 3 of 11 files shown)


rmarquis:
Great! But _<_insert another grumpy comment about yet another unnecessary cvar_>_

Aciz:
1st person view should still be synced.

https://streamable.com/fjz5ct

Aranud:
There the adjustment is done there: https://github.com/etlegacy/etlegacy/blob/b29b136a7c9f06ca22b9f941cc64bbbf9dd5d65d/src/cgame/cg_view.c#L708-L741

So, the value from `cg.duckTime` should be adjusted as well

EDIT : `PRONE_TIME` and `DUCK_TIME` should be checked, and `cgs.extendedprone` checked for adding or not the `EXTENDEDPRONE_TIME 400`

ryzyk-krzysiek:
> Great! But _<_insert another grumpy comment about yet another unnecessary cvar_>_

I agree on the cvar part, comp players will have it on, but public players will most likely not like it, they already dislike pronedelay itself hence the cvar.

> 1st person view should still be synced.
> 
> https://streamable.com/fjz5ct

The synced one feels sluggish imo and further extends the time where muzzle point is not synced with player view (muzzle point is instant) so player for longer time will think he can shoot that player while he will be shooting the curb. Current behaviour is etpro `b_extendedprone` like.

@Aranud Yea, probably something like this 

```
diff --git a/src/cgame/cg_view.c b/src/cgame/cg_view.c
index 49712af1d..fbb576be5 100644
--- a/src/cgame/cg_view.c
+++ b/src/cgame/cg_view.c
@@ -712,15 +712,22 @@ static void CG_OffsetFirstPersonView(void)
 		//origin[0] += forward[0] * 18;
 		//origin[1] += forward[1] * 18;
 
+		int pronetime = PRONE_TIME;
+
+		if (cgs.extendedprone)
+		{
+			pronetime += cg.pmext.extendProneTime;
+		}
+
 		if (timeDelta < 0)
 		{
-			cg.duckTime = cg.time - PRONE_TIME;
+			cg.duckTime = cg.time - pronetime;
 		}
-		if (timeDelta < PRONE_TIME)
+		if (timeDelta < pronetime)
 		{
-			cg.refdef_current->vieworg[0] += cg.deltaProne[0] * (PRONE_TIME - timeDelta) / PRONE_TIME;
-			cg.refdef_current->vieworg[1] += cg.deltaProne[1] * (PRONE_TIME - timeDelta) / PRONE_TIME;
-			cg.refdef_current->vieworg[2] -= cg.duckChange * (PRONE_TIME - timeDelta) / PRONE_TIME;
+			cg.refdef_current->vieworg[0] += cg.deltaProne[0] * (pronetime - timeDelta) / pronetime;
+			cg.refdef_current->vieworg[1] += cg.deltaProne[1] * (pronetime - timeDelta) / pronetime;
+			cg.refdef_current->vieworg[2] -= cg.duckChange * (pronetime - timeDelta) / pronetime;
 		}
 	}
 	else
```
Need to add `cg.pmext.extendProneTime` so the prone can be instant when you are already crouching, else you will be stuck for 400ms crouching and then go prone.

isRyven:
I'd vote for dropping the cvar and forcing the behavior. People are abusing the prone too often in fight, this should at least improve the situation.

rmarquis:
Yup. Prone is OP as hell. I'm not a fan of pronedelay in the sense it artificially freezes time to slow the behaviour (though I understand why it's done for comp purpose), but in this case, it seems like a visual improvement that also happens to make prone less OP. Default makes sense imho.

jackeri:
Yeah Id vote for dropping the cvar or at least making it cheat protected for testing purposes.

BystryPL:
I fully agree with you as well, it should be default without a cvar.

ryzyk-krzysiek:
Removed `g_extendedprone` cvar and adjusted first person view to be in sync with animation.

PR #1835, opened 2021-11-20
game: disable sudden death on second round of stopwatch, refs #1498

Exagone313:
Do not merge, it needs to be tested.

Check discussion in #1498 

---

It is now enabled on first round only (`g_currentRound == 0`).

If sudden death is used to win the first round, `g_nextTimeLimit` will be greater than the default `g_timelimit`.

### DIFF
--- a/src/game/g_main.c
+++ b/src/game/g_main.c
@@ -4296,7 +4296,8 @@ void CheckExitRules(void)
 				}
 				else
 				{
-					if (g_suddenDeath.integer && DynamiteOnObjective() && g_gametype.integer != GT_WOLF_STOPWATCH)
+					// sudden death is disabled on second round of stopwatch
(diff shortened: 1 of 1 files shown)


src/game/g_main.c
```diff
@@ -4296,7 +4296,8 @@ void CheckExitRules(void)
 				}
 				else
 				{
-					if (g_suddenDeath.integer && DynamiteOnObjective() && g_gametype.integer != GT_WOLF_STOPWATCH)
+					// sudden death is disabled on second round of stopwatch
+					if (g_suddenDeath.integer && DynamiteOnObjective() && (g_gametype.integer != GT_WOLF_STOPWATCH || g_currentRound.integer))
```
Exagone313: Note to self: actually I should move the `DynamiteOnObjective()` check after for optimization.

Issue #1433, opened 2020-04-18
Extend shoutcaster role with modern features


rmarquis:
This is the follow up of #1292, which currently implements the shoutcaster role in an equivalent / slightly better way than etpro implementation.

Some other discussed features requires deeper architecture changes, are aren't strictly necessary but are nice to have. Since bugs are likely to be introduced, these will be worked on later and I'm separating the ticket.

Extra features:
* display fireteam table when specing a specific player
* can see player health floating above them (in spec and following mode), always! (this require architectural changes, as health of all players is not sent to client - unless they are in the same fireteam)
* can see throwed nade path (see CS-Go), use `missileTrailFunc`.
* can see through wall to some extend (see CS-Go)
* can see banner prints that print objective information showing what objective has been done in what time and for the next round it also prints if that was slower or faster. For example: in the next round allies team captures the forward flag, and it prints "allies has captured the flag in 1:56 (faster on 30s than in previous round)" or something like that. See games doing speedrun competition.

rmarquis:
Note: the health floating bar is implemented in gaymod (sic), a jaymod derivative. It happens to also have a few nice ideas that make the UI clearer (despite some other crap). Worth a look.

ETPurple:
Heres a great!(terrible) paint screenshot of an example UI.
![2020-05-22-000016-oasis](https://user-images.githubusercontent.com/56918597/82613181-f84e5580-9bc4-11ea-9eb3-7c05c3144577.jpg)

ETPurple:
I've started on a starter UI based off this screenshot. There's a lot of features that can be added / removed but it's a start.
![2020-07-21-132358-oasis](https://user-images.githubusercontent.com/56918597/88049824-0fbea600-cb56-11ea-9b25-7be92d5d1b8e.png)

Issue #1282, opened 2020-03-05
Rework compass


ETLegacyTeam:
Author: @BystryPL (Bystry)
Date: 2019-05-06
Redmine Issue: 1282, https://dev.etlegacy.com/issues/1282

---

bq. The colours feature made me think a little bit more about compass usability and I started paying more attention to it and think about how it could be improved. With alternative HUD the compass displayed in the bottom part of screen is pretty small, the players icons are almost invisible with  big zoom and you can mainly use it for the "syringe" icon. While the command map is very good to track your teammates, it's very hard with the compass. I had a talk with hatcher about it  and we came to the conclusion that it could be made bigger by cutting off the edges with directions. Together with the colours it would greatly improve in general the compass functionality, especially for competitive gameplay. Hatcher made a prototype, you can see it on screenshots above (before and after). 

bq. The direction have they're usefulness too, but I'd agree it's too wide. A thinner one would be much cleaner imho.

---

Related issue(s): #1000, #1271

---

Related commit(s):
6630820a4bc78f345017386daa58c61c901d4dfd - cgame: increased compass size on alternate huds, r..

ETLegacyTeam:
From: @mittermichal (kimi)
Date: 2019-05-06T17:22:59Z

---

i had the exact same idea when i was working on custom HUD
i think it should be done with using componentStyle::STYLE_SIMPLE
so far STYLE_SIMPLE is only used for drawing skills

ETLegacyTeam:
From: @realkemon (keMoN)
Date: 2019-05-10T10:02:48Z

---

Improved graphics and used higher resolution. Also included .psd for future reference.
This simply makes the compass border thinner and replaces the direction with a single red indicator for North. 
Size of map needs to be adjusted in code.

ETLegacyTeam:
From: @realkemon (keMoN)
Date: 2019-05-10T14:06:44Z

---

Currently, Fireteam members are marked on the compass with a white icon. This is present even when those members are already in close enough vicinity to be drawn on the minimap.
I propose to use the player's icons instead. This can be combined with @bystry's suggestion of coloured player icons (#1271). In "2019-05-06-181627-goldrush-2.jpg" they are represented by the white icons. Fireteam members that are not in vicinity of the minimap are drawn on the border and receive an arrow indicating that they are out of bounds. Fireteam members in minimap vicinity are drawn as usual in their respective colour.
[...]

Issue #1000, opened 2020-03-05
Modernize HUD


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2018-02-21
Redmine Issue: 1000, https://dev.etlegacy.com/issues/1000

---

Based on Kemon early design, give a refresh to the player HUD.

* Use of rounded chargebar (as an option), by implementing Bezier curves
* Implement skill charges, to quickly know when a weapon/skill can be used. We currently implement a colored chargbar, but this requires to switch to the according weapon to know its status.
* Implement a good alternative HUD using the current HUD code.
Unlike the design picture, keep the compass as it is very useful.

The info on screen should be displayed in a more minimalist way, to lessen distraction and ease focus on game.

Remove all unnecessary lines and/or colors for:
* Fireteam window
* Scoreboard
* Stats, Weapons stats and Objective windows
* Build and location indicators
* Skill levels and skill/promotion popups
* Flags and other minor details
* Fireteam option menu
* In-game chat windows

Ensure consistency with the end of game debriefing screen.

Some early WIP screenshot attached.

---

Child issue(s): #1073, #1074

rmarquis:
Kemon's concept:

![concept_HUD](https://user-images.githubusercontent.com/162442/85201770-3fce1b80-b302-11ea-8c0d-e92fa7a539e8.png)

PR #1849, opened 2021-12-01
cgame: merge shoutcast minimap with compass and allow usage of square minimap, refs #1000 #1282 #1433

ryzyk-krzysiek:
Initialy I separated shoutcast minimap and compass because I didn't want to break compass as I barely knew the project. I did it however in mind of eventually adding it back and allowing using it for normal gameplay and this PR should be it.

- standard compass should be working as before
- shoutcast minimap should be working as before - it is still forced in both the size and placement
- enabled with `cg_drawCompass 2`

I would consider it a work in progress, standard `cg_altHud` settings will correctly draw the square minimap, but it is hardly usable. It's too small and icons are too big for such small size. As of now it is best used with custom hud, where players can set the size and location as they wish.

What I feel could and probably should be worked on in the future:

- improve the icons scaling - square minimap uses big icons so they are easily visible when max zoomed out but with small minimap dimensions it barely presents any information because of it. Compass on the other hand has **very** small icons when max zoomed out making it only usable with big zoom which reduces the overall needed "visibility".
- fix voice chat sprites - on square minimap they can be drawn out side of minimap (same as compass but it has bigger border). Position of sprites is also pretty bad - full zoom it is under the player icon, full zoom out it is way too far from player icon.
- to make square minimap work with current `cg_altHud`s it is also reduced in size just like compass which shouldn't be necessary imo. https://github.com/etlegacy/etlegacy/compare/master...ryzyk-krzysiek:square-minimap?expand=1#diff-09bf4aee5c64caee30f85eb55b903a2b2004f3461eb16040b223a25ff18523b8R1771-R1775

Custom hud example:
square minimap

refs #1000 #1282 #1433

### DIFF
--- a/etmain/scripts/legacy.shader
+++ b/etmain/scripts/legacy.shader
@@ -805,6 +805,20 @@ sprites/ready
 	}
 }
 
+sprites/cm_medic_icon
+{
+	nopicmip
+	nocompress
+	nomipmaps
+	{
+		map sprites/voicemedic.tga
+		depthFunc equal
+		blendfunc blend
+		rgbGen vertex
+		alphaGen vertex
+	}
+}
+
 ui/assets/mp_ammo_blue
 {
 	nopicmip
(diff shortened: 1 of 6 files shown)


mittermichal:
neat

BystryPL:
Extremely neat.

PR #1853, opened 2021-12-05
cmake: add DOCDIR variable

Flowdalic:
Some distributions, e.g. Gentoo, require a particular naming scheme
for /usr/share/doc. In case of Gentoo, the package version is also
included, so it would be for example /usr/share/doc/etlegacy-2.78.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ The following variables can be adjusted in CMake:
   * (optional) **`INSTALL_DEFAULT_MODDIR`**: Location for libraries and paks. Appended to `CMAKE_INSTALL_PREFIX`.
     Defaults to `lib/etlegacy` and then `legacy` is appended to it.
 
+  * (optional) **`DOCDIR`**: Location for documentation. Defaults to `INSTALL_DEFAULT_SHAREDIR/doc/etlegacy`.
 
 ### Linux
 
--- a/cmake/ETLInstall.cmake
+++ b/cmake/ETLInstall.cmake
@@ -2,6 +2,10 @@
 # Install
 #-----------------------------------------------------------------
 
+if(NOT DOCDIR)
+	set(DOCDIR "${INSTALL_DEFAULT_SHAREDIR}/doc/etlegacy")
+endif()
+
 # description file - see FS_GetModList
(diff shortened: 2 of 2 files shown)


rmarquis:
I've had a very quick look at the [lastest FHS spec](https://refspecs.linuxfoundation.org/fhs.shtml), and the `doc` directory isn't explicitly mentioned. On Arch, both version (with and without version string) seem to be used. I can't check on other distro though.

If there is no drawback (ie, if other distro don't explicitly require *not* to include the version string), I'd suggest to simply add the version as standard for the `doc` directory, as opposed to introducing yet another cmake variable.

Flowdalic:
> I'd suggest to simply add the version as standard for the `doc` directory, as opposed to introducing yet another cmake variable.

I am afraid that won't work, as Gentoo uses "/usr/share/doc/${PF}" where PF includes the package name and a potentially Gentoo refined version string, e.g. etlegacy-2.78.1-r1. Also 'docdir' is a common variable of [GNU standard installation directories](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), that is what allows Gentoo to tailor it to its needs.

rmarquis:
I see, guess we have to do it this way then. Do you mind also adjusting the doc in README file too?

Flowdalic:
Of course, but unfortunately I am not sure which 'doc' in https://github.com/etlegacy/etlegacy/blob/master/README.md you mean. There seems to be no reference to the docdir in this file.

Flowdalic:
Friendly ping. I'd really like to get ET legacy into Gentoo's main package repository. :)

Aciz:
Perhaps he means the part about `INSTALL_DEFAULT_SHAREDIR` under `COMPILE AND INSTALL`

https://github.com/etlegacy/etlegacy#compile-and-install

Flowdalic:
> Perhaps he means the part about `INSTALL_DEFAULT_SHAREDIR` under `COMPILE AND INSTALL`

That makes sense. Thanks. Pushed a new version.

Aranud:
@Flowdalic Thank you for your contribution !

Issue #218, opened 2020-03-05
Replace assets with our own


ETLegacyTeam:
Author: @jackeri (Jacker)
Date: 2013-12-29
Redmine Issue: 218, https://dev.etlegacy.com/issues/218

---

We should slowly start replacing assets with our own so that in the distant future we wont require the original ET pak0,pak1,pak2 files.

Assets such as models should be created fully in iqm (which i will implement to the renderers) as that model format has bone animation and its supported in many forms of applications.
http://sauerbraten.org/iqm/

When we have fully replaced the files we don't need players to have an preinstalled version of vanilla ET.

This step comes after the new renderer is on a decent functional level.

---

Related issue(s): #625
Child issue(s): #452, #817, #867, #992, #1002

ETLegacyTeam:
From: @adanski (Dragonji)
Date: 2013-02-26T16:50:53Z

---

What about backward compatibility in this case? I'm not an expert in programming stuff but even if the compatibility could be remained I can see some security reasons why not to make own assets. By security I mean "cheating" in this case.

Please correct me if I am wrong.

ETLegacyTeam:
From: @jackeri (Jacker)
Date: 2013-02-26T21:39:54Z

---

I don't see how cheating is an issue? Only packs that the server has will be loaded anyway. Anyways loading packs into memory would be the least effective cheat... ever. As for backward comp, simply if the player doesn't have a pack it will downloaded and used and as for mods there should be no impact.

ETLegacyTeam:
From: @adanski (Dragonji)
Date: 2013-03-02T10:21:50Z

---

My bad, I guess paks are being checked on server side so there is no easy way to replace something in there.

Jacker wrote:
> As for backward comp, simply if the player doesn't have a pack it will downloaded and used and as for mods there should be no impact.

Well, for (impatient) people with > 1 Mb/s connections it will be a pain in the ass to download around 200 MB while connecting to the server. Well, the size of the paks can be much smaller of course if sounds were compressed to Ogg Vorbis and textures to a better format than TGA (I guess JPG is not an option cause this format produces really low quality stuff).

ETLegacyTeam:
From: @jackeri (Jacker)
Date: 2013-03-02T18:24:53Z

---

[...]

PR #1861, opened 2021-12-15
mod: SVG weapon icons

WuTangH:
Adding new weapon icons in .SVG format, which can be seen in fireteams, shoutcaster mode or obituaries.
Icons have black strokes, but testing on bright/snowy maps would be good.

Can be also used for simple item icons maybe?

refs #1433 
refs #218


isRyven:
Why would you remove old weapon icons? They could serve as a fallback for vet clients.

WuTangH:
True, didnt think about vet players on legacy servers... VET throws warnings and uses large icons instead, which doesnt look good. 
Keeping both seems to work alright on both clients.

jackeri:
Screenshot comparisons would be really nice for tickets like these :)

WuTangH:
Ah right, added some pictures to the 1st post.

PR #1883, opened 2022-01-10
Hud scale/color configurability

Helldembez:
Still WIP, but you can look through the changes so far.

### DIFF
--- a/etmain/ui/huds.hud
+++ b/etmain/ui/huds.hud
@@ -8,33 +8,33 @@ hudDef {
 		// hudnumber value must be unique, if not then the hud will be discarded
 		// PS. 'x' should be between 0 and 640 and 'y' between 0 and 480
 		hudnumber 3
-		compass 44 393 84 84 0 1
-		staminabar 4 388 12 72 0 1
-		breathbar 4 388 12 72 0 1
-		healthbar 604 388 12 72 0 1
-		weaponchargebar 624 388 12 72 0 1
-		healthtext 580 415 0 0 0 1
-		xptext 48 476 0 0 0 1
-		ranktext 62 464 0 0 0 1
-		statsdisplay 24 385 0 0 1 1
-		weaponicon 538 424 60 32 0 1
-		weaponammo 598 470 0 0 0 1
-		fireteam 400 10 100 100 0 1
-		popupmessages 4 100 72 72 0 1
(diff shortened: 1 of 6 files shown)


jackeri:
Instead of jamming more stuff to the one line of hud definition why not branch out and create a proper component definition inside the braces {}. I thought that you had decided on doing it that way already? You could have just PC_PeakToken and checked for a { and that way kept backwards compatibility with the old format for now..

isRyven:
Could we perhaps just add `json`, `yml` or `ini` support, something that can be used to create and parse the nested structures, without the need to constantly reinvent the wheel. I would also be interested in seeing one of the listed formats being whitelisted on client side instead of labeling everything as `dat`, it would then be easier to write generators that can output the file as it is - without the extension changes, which could be then properly edited by the editors with full syntax support. Also in this case, you could just then drop the idea of extending legacy hud format, and just add the new one defined in the `json` or other format, something like `/huds/my_awesome_hud.json`.

Helldembez:
> Instead of jamming more stuff to the one line of hud definition why not branch out and create a proper component definition inside the braces {}. I thought that you had decided on doing it that way already? You could have just PC_PeakToken and checked for a { and that way kept backwards compatibility with the old format for now..

Yeah, I did decide on changing the configuration formatting. But after working with the code more I am convinced there needs to be a whole other way to configure hud elements. Now we have a system that doesn't make sense for how it is used. There is a hudcomponent struct with the fields we know. But the way the values of the fields are used differs per component. Some components don't use the w/h at all, or have overlap with cvars like visibility and cg_drawtime. I've even seen width get used as a scale.

So for now I am just extending the current format to support color and scale for text components, but will make it optional so existing hud files don't break. As for a new structure, like Ryven mentions we could use a standard, though I don't think we want to use a library for it and only supporting the basics means you don't support the standard anyway. What we need is a way to define each individual component with its own unique fields/values. Something like a map maybe.

isRyven:
Well you can use library for other means, not just for the hud, internally you can use it for other configurations, one idea was to use this for weapon table for example. Having something that is widely known and has editor support or even natively supported by some languages would help to improve the format. This also means building the hud editor (as external app) would be a much easier job. Besides, it should be easy enough to just document the accepted fields for the components. But in general i am just looking forward to see more whitelisted extensions by the client :p

PR #1887, opened 2022-01-15
game: fix throwable missiles infinite bouncing

ryzyk-krzysiek:
- Grenades will now behave like other missiles and will drop if the ground they were on is gone (etpro like)
- Adjusted the code where in some occasions missiles would bounce forever and were making a lot of noise and players were unable to interact with them. This can still happen to my knowledge (but less than before) at least with doors when they are entirely stuck in them, but players will be able to interact with them unlike before.

### DIFF
--- a/src/game/g_missile.c
+++ b/src/game/g_missile.c
@@ -122,7 +122,7 @@ void G_BounceMissile(gentity_t *ent, trace_t *trace)
 
 		// check for stop
 		//if ( trace->plane.normal[2] > 0.2 && VectorLengthSquared( ent->s.pos.trDelta ) < Square(40) )
-		if (trace->plane.normal[2] > 0.2f && VectorLengthSquared(relativeDelta) < 1600) // Square(40)
+		if ((trace->plane.normal[2] > 0.2f && VectorLengthSquared(relativeDelta) < 1600) || !trace->fraction) // Square(40)
 		{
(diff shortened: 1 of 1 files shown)


ryzyk-krzysiek:
Not sure if you noticed or not yet but this has a serious issue when it comes to rifle grenades. When hitting a moving player (especially that is running towards you) missiles do not bounce as I expected and go inside of the player which causes this code (`!trace->fraction`) to trigger. It's fine for everything beside rifle grenade because of this code (which I totally misunderstood first after reading the comment):

https://github.com/etlegacy/etlegacy/blob/e647e237a9e390f097f6675f9ca84703b07ad7ab/src/game/g_missile.c#L135-L139

On my branch I fixed it just by excluding rifle grenades from the additional check for stop bouncing so the faster explosion doesn't trigger. Rifle grenades or even nades were never really an issue so something like it should be fine (other way could probably be to check `if (ent->r.contents == CONTENTS_CORPSE)` to exclude both grenades and rg).

https://github.com/ryzyk-krzysiek/etlegacy/commit/96a37fc5f204287613b63257e33b6d87946f99a0

Aciz:
Yeah I was just looking into this today, I think only including throwables `GetWeaponTableData(ent->s.weapon)->firingMode & WEAPON_FIRING_MODE_THROWABLE` is another option as that's what the fix is initially for.

PR #1890, opened 2022-01-17
game: apply spread to rifles and scoped weapons while in air

ryzyk-krzysiek:
As of now rifles (including covert ops rifle) and scoped weapons have no spread applied like every other weapon while player is in air. Smgs spread while in air is `800` and pistols is `1200`, rifle spread on the other hand is always the same `250` or no spread at all for scoped weapons, meaning players can jump as much as they want to without getting penalized like other weapons.

This PR would mean weapons spread in air would be:
- rifle `500`
- scoped rfile `1400`
- scoped fg42 `400`

https://streamable.com/rk3fht


ryzyk-krzysiek:
¯\\_(ツ)_/¯

PR #1891, opened 2022-01-17
cgame: add extra cvars to control popups more thoroughly

Helldembez:


jackeri:
Why add the popuptime? Its the delay right?

Helldembez:
Yes, so if many things happen you have some time to see them instead of disappearing behind the spam. I personally don't use it, but I don't see a reason to not add it.

Issue #1799, opened 2021-10-31
Rework gibs stats accuracy


rmarquis:
As a follow up of #478 and the discussion about individual stats weapon in #1797 and #1800, see if reworking the way accuracy involving gibs can be done too. Currently these hits and shots are completely ignored.

> Fix gib stats. Neither shots or hits when forcing into limbo are taken into account. See G_Damage() and G_addStats(). NQ does count only shots, so accuracy decrease when forcing into limbo. ETPub allows a cvar for this (g_stats).

Maybe display a "real accuracy" along an "adjusted accuracy"? The current value is the latter one.

This should be handled after the rework for individual stats weapon is done (see #1800).

PR #1895, opened 2022-01-24
game: fix accuracy calculation for splashdamage weapons

Aciz:
The hack in accuracy calculation that's intended to not count shots/hits for a weapon when you use it to gib someone was causing splash damage weapons to calculate accuracy wrong, as the reduction in shots would be called for every hit, meaning if you hit 2 wounded enemies with a grenade, your total "shot" grenades would go down by 2. This PR removes this hack completely for non-hitscan weapons, since you could argue that you don't really purposefully aim them anyway (except maybe flamethrower). Fixing this "for real" to work properly on splash damage weapons would be close to impossible I think (yes, ETPro has this exact same bug).

### DIFF
--- a/src/game/g_match.c
+++ b/src/game/g_match.c
@@ -291,13 +291,19 @@ void G_addStats(gentity_t *targ, gentity_t *attacker, int damage, meansOfDeath_t
 	{
 		if (attacker && attacker->client)
 		{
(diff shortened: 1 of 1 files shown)


Helldembez:
#1799

PR #1897, opened 2022-01-27
cgame: introduce sharetimer command

Helldembez:
With this command you can share the spawntimer you have set for when you think the enemy spawns. So `roundtimer - current spawntimer` e.g. `14:30 - 23 = enemy spawns at 7`. This is also configurable with `${nextspawn}` and `${enemylimbotime}`.

Please take a look at how buffer and text are handled. I would like it to be cleaner.

### DIFF
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -2320,7 +2320,7 @@ static void CL_FrameHandleVideo(int *msec)
 		//clc.aviVideoFrameRemainder = frameDuration + msec;
 	}
 	else if ((!cl_avidemo->integer && CL_VideoRecording())
-			 || (cl_avidemo->integer && (cls.state != CA_ACTIVE || !cl_forceavidemo->integer)))
+	         || (cl_avidemo->integer && (cls.state != CA_ACTIVE || !cl_forceavidemo->integer)))
 	{
 		CL_StopVideo_f();
 	}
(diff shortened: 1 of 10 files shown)


mittermichal:
neat

Helldembez:
Forgot to add how it looks: https://streamable.com/85kpfi

mittermichal:
should also add sharetimer and sharetimer_buddy to controls menu
Also consider that If custom text was stored in cvar it could be then configured through menu

mittermichal:
or if there are no arguments then it would check cvar, if cvar is also empty, use default message

Helldembez:
> should also add sharetimer and sharetimer_buddy to controls menu
> Also consider that If custom text was stored in cvar it could be then configured through menu 
> or if there are no arguments then it would check cvar, if cvar is also empty, use default message

I've implemented it. The two commands are added to the advanced controls under misc. But I am not sure where to add the custom text field in the menu. Since it's not a control, maybe under view -> advanced and create a misc category as well?

mittermichal:
imo options -> game

jackeri:
Instead of doing this in such a style, you could take a look how ezQuake handles "macros", those would be pretty nice to have in ET as well.

Helldembez:
> Instead of doing this in such a style, you could take a look how ezQuake handles "macros", those would be pretty nice to have in ET as well.

Like this? http://www.ezquake.com/docs/guides/macros.html

I can check for `%s` and `%l` in the command code, but incorporating that into the whole text system might be a bit more tricky. Like some information is only available client side and vice versa.

jackeri:
mean the $lastip etc

ensiform:
Chat tokens are pretty easy to parse in G_Say processing on the server side

src/client/cl_main.c
```diff
@@ -3013,6 +3013,7 @@ void CL_Init(void)
 	Cvar_Get("cg_autoReload", "1", CVAR_ARCHIVE);
 	Cvar_Get("cg_weapaltReloads", "0", CVAR_ARCHIVE);
 	Cvar_Get("cg_weapaltSwitches", "1", CVAR_ARCHIVE);
+	Cvar_Get("cg_sharetimerText", "", CVAR_ARCHIVE);
```
ensiform: Very little reason to pollute newly added mod-specific cvars in the engine tbh.

Helldembez: Thanks for checking. I just added the cvar in places where the cvar I used as example was.

PR #1902, opened 2022-02-04
cgame: cg_drawCompass 2 fixes and improvements

Aciz:
Fix and improve several aspects of `cg_drawCompass 2`. These changes don't affect the shoutcaster minimap at all (apart from `cg_drawCompassIcons` icons not being affected by map zoom), they're only fixes for when you used the square minimap while playing.

* Fixed icon scaling to match regular compass
* Fixed slightly incorrect out of bounds check for icons
* Icons are now drawn when they're inside the map and respect map zoom
* Added indicators for N, S, E, W

Additionally, I also removed grid drawing from the regular compass, I don't think it serves any purpose and just adds visual clutter. Grid is still drawn on expanded automap.

### DIFF
--- a/src/cgame/cg_commandmap.c
+++ b/src/cgame/cg_commandmap.c
@@ -52,13 +52,22 @@ qboolean ccInitial = qtrue;
 #define FLAG_TOPFRAC                0.7421875f // 95/128
 #define SPAWN_SIZEUPTIME            1000.f
 
-// square minimap values
-#define AUTOMAP_PLAYER_ICON_SIZE_NEW    20
-#define CONST_ICON_NORMAL_SIZE_NEW      17
-#define CONST_ICON_EXPANDED_SIZE_NEW    20
-#define CONST_ICON_LANDMINE_SIZE_NEW    8
-#define FLAGSIZE_EXPANDED_NEW           98
-#define FLAGSIZE_NORMAL_NEW             82
+// shoutcaster minimap values
+#define AUTOMAP_PLAYER_ICON_SIZE_SC     20
+#define CONST_ICON_NORMAL_SIZE_SC       17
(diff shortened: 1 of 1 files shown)


ryzyk-krzysiek:
> Fixed icon scaling to match regular compass

Probably it is just me, but I always imagined that minimap would be used the same way as minimaps in most other games, that is you can see the whole map (max zoomout in ET).
That's how I always used it myself and changing scaling to regular compass makes it unusable (icons so small you can't really see them, example below).




My proposal for a solution of icons would be an idea I had for a bit of time to maybe introduce dynamic icon scaling that I remember from Dota 2 from few years back - scaling based on other players proximity.
`cg_dynamicIcons` dynamic player icon scaling on/off (default off)
`cg_dynamicIconsDistance` distance between players when scaling down starts
`cg_dynamicIconsSize` player icon size. command map defaults: compass 5, minimap/sc 20
`cg_dynamicIconsMaxScale` max scale
`cg_dynamicIconsMinScale` min scale

Short example with just the defaults I came up with: 
https://streamable.com/vrovnp
Code:
https://github.com/ryzyk-krzysiek/etlegacy/commit/db7746471661076e4dd7d935ee52bf1bb5cfe9ec

My solution although is pretty basic allows to pretty much control player icons how anyone would like I imagine, even just changing icon size from right now hardcoded default and disabling scaling by setting min and max scale to 1.
If ever something like this would be to be added, then I guess defaults should be done based compass since it is most used and other players (like me) could set it to something else.

Aciz:
Very cool concept! Few people have actually asked if it would be possible to set the icon scale manually, and I agree it would probably be a nice addition to allow changing the icon sizes. Could probably be added alongside this.

PR #1903, opened 2022-02-05
cgame: add cvars to control fireteam alpha values

Aciz:
Cvar to allow adjusting the background alpha of fireteam overlay. Full fireteam alpha is unfortunately not possible (at least in a simple way) since we use images in the overlay, so they get their alpha values from shaders.

### DIFF
--- a/etmain/scripts/legacy.shader
+++ b/etmain/scripts/legacy.shader
@@ -847,6 +847,28 @@ sprites/cm_voicechat_icon
 	}
 }
 
+sprites/uniform_allied_hud
+{
+	nocompress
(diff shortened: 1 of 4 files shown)


isRyven:
use setcolor trap to add alpha to images

Aciz:
> use setcolor trap to add alpha to images

How exactly? I can't think of any example in code where this is done, besides that would then mess up the image colors, wouldn't it?

Aciz:
Added overall alpha adjustment too via `cg_fireteamAlpha`, thanks for the tip @isRyven

Issue #1957, opened 2022-04-16
Gun sounds / step sounds gets doubled 


BystryPL:
https://clips.twitch.tv/NurturingZealousMonitorJebaited-GGdFzdx2SRpuoiQb

The gun sounds and step sounds gets doubled on a regular basis. This intensifies when the load on the network increases (streaming or playing on overpopulated server such as TM public).

Aciz:
The fix in #1968 seems to be working well and there are no side effects reported, closing.

PR #1968, opened 2022-04-24
mod: change handling of player events in client game, refs #1957

ryzyk-krzysiek:
The issue of doubled sound occurs when client miss predicted events causing them to play again. For example server sometimes predicts 2 events a `footstep` and `fireweapon` compared to client only predicting fireweapon and if footstep comes before fireweapon, both will be played.

The proposed solution in PR to this is to ignore all predictable server events (except non-predictable server only generated ones). So in above case we will no longer get a secondary fireweapon event and additional footstep that we miss predicted. However nothing changes for server and nothing changes for how other players are perceived as it's server that decides what events they get, so other players will still hear a footstep and fireweapon.

To do this, client will no longer save events to `playerstate_t` but to `pmoveExt_t` and will ignore all predictable events coming from server in playerstate. For prediction we will only check for new non-predictable events and if we find them we will do full prediction as before so we can correctly play them.

refs #1957

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1018,8 +1018,6 @@ typedef struct
 // all cg.stepTime, cg.duckTime, cg.landTime, etc are set to cg.time when the action
 // occurs, and they will have visible effects for #define STEP_TIME or whatever msec after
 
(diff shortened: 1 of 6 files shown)


Aranud:
Neat

PR #1969, opened 2022-04-27
cgame: colorize bots ping column in scoreboard

Aciz:
Color the `BOT` text in ping column on scoreboard yellow for bots to better differentiate them from human players.


Aranud:
Very neat

PR #1971, opened 2022-04-29
engine: improve VM_Create errors

Aciz:
Improved `VM_Create` errors so average user might have some idea on what's actually wrong.

I'm not 100% sure if the method I used to check for x86 platforms is the most optimal, but at least it works, let me know if it should be done some other way.

### DIFF
--- a/src/client/cl_cgame.c
+++ b/src/client/cl_cgame.c
@@ -187,7 +187,7 @@ qboolean CL_GetSnapshot(int snapshotNumber, snapshot_t *snapshot)
 	for (i = 0 ; i < count ; i++)
 	{
 		snapshot->entities[i] =
-		    cl.parseEntities[(clSnap->parseEntitiesNum + i) & (MAX_PARSE_ENTITIES - 1)];
+			cl.parseEntities[(clSnap->parseEntitiesNum + i) & (MAX_PARSE_ENTITIES - 1)];
 	}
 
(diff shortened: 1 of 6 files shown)


isRyven:
I just wonder if throwing an error from within the `VM_Create` could reduce the duplication. 

Also have you thought about using `ARCH_STRING` instead of `CPUSTRING`? seems a bit more generic from my point of view: https://github.com/etlegacy/etlegacy/blob/00b35a7e14db6f0b73dd6fbf5fdfe7436a8ed776/src/qcommon/q_platform.h#L200

There is also "boolean" macro:
https://github.com/etlegacy/etlegacy/blob/00b35a7e14db6f0b73dd6fbf5fdfe7436a8ed776/src/qcommon/q_platform.h#L52

Aciz:
From what I can see, `ARCH_STRING` is not `i386` on win-x86? So I'd still need to do two checks, one for linux-i386 and another for win-x86?

https://github.com/etlegacy/etlegacy/blob/00b35a7e14db6f0b73dd6fbf5fdfe7436a8ed776/src/qcommon/q_platform.h#L124-L142

isRyven:
Oh that windows guy is so annoying! 

You can also make own macros like `#define ARCH_X86 1` wrapped with `#if defined(_M_IX86) || defined(__i386__)` which will consolidate the platforms.

PR #1985, opened 2022-05-08
ui: fix warning when starting dedicated server

mittermichal:
it is possible to connect to dedicated server on same machine


mittermichal:
forgot to edit etlegacy_mod.pot

mittermichal:
I don't have to bother with *.po files, right?
Interestingly in czech translation it's already correct: "Pro připojení k serveru musíš hru spustit znovu." -> To connect to server, you have to run the game again.

PR #1987, opened 2022-05-08
cgame: change minimap icons to follow same logic as vanilla compass

ryzyk-krzysiek:
https://streamable.com/l8gh85


Aranud:
Neat

PR #1988, opened 2022-05-09
cgame: fix incorrect zoom when trying to zoom with FG42

Aciz:
In original code, the check here explicitly checked for scoped K43/Garand, but it was now simply checking for scoped weapons, which allowed FG42 to access the `CG_AdjustZoomVal` function. This made it bypass `cg_zoomDefaultSniper` and use the `55` zoom value from weapon table instead when trying to zoom in/out.


Aranud:
Bit-wise master

PR #1990, opened 2022-05-09
game: add spread for snipers when scoped while in air, refs #1796

Aciz:
Adds spread to snipers while scoped in air, currently they have 0 spread and can be shot with 100% accuracy while jumping since we don't scope out anymore.

### DIFF
--- a/src/game/g_weapon.c
+++ b/src/game/g_weapon.c
@@ -3334,7 +3334,8 @@ void EmitterCheck(gentity_t *ent, gentity_t *attacker, trace_t *tr)
  */
 void Bullet_Endpos(gentity_t *ent, float spread, vec3_t *end)
 {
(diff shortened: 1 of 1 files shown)


cakeET:
Would it be better to make this adjustment in pmove since this is a positional/movement based effect on spread? Here the effect is hidden from the client because it only randomizes the end point and doesn't show up in the spread bar.

You could do the same check for ground contact in PM_AdjustAimSpreadScale and just set aimSpreadScale and aimSpreadScaleFloat to AIMSPREAD_MAXSPREAD.

x0rnn:
I guess this can get closed now?

Aranud:
Patch to reduce speed to 80 on standinng position (same as crouch) and unscope the weapon while in air (jumping, falling ...)
[0002.patch](https://github.com/etlegacy/etlegacy/files/10279062/0002.patch)

PR #1991, opened 2022-05-11
cgame: fix incorrect levelshot in mapvote menu for missing maps

Aciz:
When a client doesn't have a map downloaded and they select it in the mapvote list, the UI was showing the levelshot of the previously selected map, instead of `levelshots/unknownmap` shader like on serverlist.


mittermichal:
cool :ok_hand:

PR #1993, opened 2022-05-11
game: fix a case where players were warping and don't antiwarp bots

ryzyk-krzysiek:
1. As far as I know the part of the antiwarp design was to start allowing all incoming user commands if the player stopped moving, because - as I understand - he can't warp if he is not moving. This is done by checking user move commands in `G_CmdScale`, but the issue is player doesn't come to a stop as soon as those commads stop, which was causing player to warp for the remaining commands.
2. Antiwarping bots makes no sense.
3. If I understand correctly, as of now and just like in etpro, players are allowed to warp for the 75ms worth of commands `LAG_MAX_DELTA`. From what I tested warping is a bit visible still because of it. Could be maybe adjusted in the future but for now I left it be, because I'm not 100% sure if current fix in PR will be good and if changing that 75ms to lower value will not make playing less enjoyable for players with bad connections.


Aranud:
Neat

Issue #1921, opened 2022-02-26
Wallbugging


x0rnn:
Is wallbugging fixed in ETL or is it possible to fix it at all?

Example: https://youtu.be/rR1im0oYNmA

Basically you set com_maxfps 43 and in some walls/doors you can wallbug through. Binding screenshotJPEG to a key and pressing it sometimes gives you like a 1000 UPS boost when stuck in a wall, which aids greatly in being able to wallbug through certain walls. Perhaps if there is no other way to fix it, preventing screenshotJPEG from giving you such a big UPS boost for some reason when stuck would be at least a 50% of a fix for this.

Issue #1208, opened 2020-03-05
Not possible to crawl close to wall and objects


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2019-02-21
Redmine Issue: 1208, https://dev.etlegacy.com/issues/1208

---

It's not possible to crawl close to objects or walls, unlike vanilla. It is still possible to get close by crawling sideways, after which crawling forward directly towards the wall will push the player backward.
This *might* be related to fixed realHead box (very wild guess).

---

Related issue(s): #503, #1115

---

Related commit(s):
8c5b0c59036abde79acc41ab56771e48dbadfe2d - game: prevent rotation while proning if head go so..
 991c32d9c522d4e64c0d68455418f7cb10dad567 - game: add g_debugPlayerHitboxes 4 to display hitbo..
 683e7e735a0fa23f25c72b6ea48673f0abc57953 - cgame: forward view origin when prone to match hea..
 b2b08a5af1bc2aacd413ae6069d69f82a656bbab - cgame: don't play prone transition view when crouc..
 a4ab1a39612188e7e906144631e5bb9f28bb7ae5 - game: improve debug players hitboxes collision, re..
 8f306555761055b64f95c69224895f726d527282 - game: adjust collision detection on prone, refs 1..
 b22f27eb31d8f8c475da514df1c00d67b087eb6d - game: adjust Z hitbox collision on prone, refs 12..
 164b964c3682c0098363470e4c279d6dceff4d7b - game: removed unused variable
 3083f98b53cd8123e89df2cc486007ccfbf1b30c - mod: fix head/legs collision on steppable elements..

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-03-16T14:44:23Z

---

-I can reproduce as far as 2.72. This is much likely due to the readhead code, which was introduced in 2.71.- Can also reproduce on 2.71rc3, the issue might be even older.

And the current behaviour is correct, the player shouldn't be allowed to crawl too far... because the model then crawl inside the wall/objects.
It is however possible to crawl "from the side" (since the headbot doesn't interfere) and then rotate the body. See screenshot below.

Might be a good idea to prevent that rotation somehow.

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-11-13T13:38:23Z

---

[...]

PR #1994, opened 2022-05-12
game: fix prone collision (stuck in walls/going through walls), refs #1208 #1921

ryzyk-krzysiek:
Afaik vanilla issue is that (at least one part) during `PM_UpdateViewAngles` there are traces for just legs which would make body BBox enter into solid in some occasions. This was party fixed in legacy by also tracing for body BBox by using `PM_TraceAll` instead. But the `mins` and `maxs` were still not set properly server side (in pmove it is done after `PM_UpdateViewAngles`). Client side it is kept from frame to frame in `cg_pmove`.

refs #1208, #1921


Aciz:
Neat

PR #1998, opened 2022-05-13
server: fix sv_ipmaxclients acting like sv_maxclients

Aciz:
`sv_ipMaxClients` was incorrectly blocking connections when there were more total clients than it was set to, due to `clientIP` being always set to `theirIP`, probably due to some pointer or overflow error happening in `Net_AdrToString` (honestly didn't care to figure out :D) . It will now correctly limit connections per IP address.


Aranud:
Neat

PR #2000, opened 2022-05-14
cgame: clear medic/ammo request voicesprites on respawn

Aciz:
Reset voicechat sprite time on respawn if current voice chat sprite is medic/ammo request sprite. Medic sprite will always be cleared, but ammo sprite won't be if player gets revived.


isRyven (approved):

mittermichal:
neat

PR #2008, opened 2022-05-18
etmain: reorganize legacy mod shaders

Aciz:
Reorganized the shader files shipped with the mod. Spread the shaders out into many more shader files, based off the shader name, following etmain convention. Nothing has actually changed content wise, this is simply a formatting/structural change.


Aranud:
Trusted

PR #2010, opened 2022-05-19
fix #313 removes spread growth cap for mouse moves

### LINKED ISSUE #313: Mouserate/FPS dependant turnspread
Author: S3ti (S3ti)
Date: 2013-06-09
Redmine Issue: 313, https://dev.etlegacy.com/issues/313

---

If FPS are higher than mouse polling rate, turning spread reduces.
If you get twice the FPS than your mouse polling rate (e.g 250FPS/125Hz) you get 0 added spread by moving your mouse.
Overall turning spread seems to be somewhat FPS dependant.
Maybe it's fixable or made selectable through a server cvar?

http://dev.etlegacy.com/boards/3/topics/513

cakeET:
Removes the cap on spread growth for rapid mouse movements and with the +left and +right commands.

This resolves issue #313 with FPS and mouse Hz and also stops an exploit where there is no spread growth when using a 180 turn script. Absolute maximum spread (255) is unchanged.

A side effect of this change is there is no longer an engine related advantage to making very fast mouse movements. Before this change, to control spread growth when playing, it was best to either move the mouse slowly or very quickly (to minimize the number of frames that the maximum amount of spread growth would be applied). There is still a general strategic benefit to moving quickly: the sooner you stop moving, the sooner your spread will settle.

### DIFF
--- a/src/game/bg_pmove.c
+++ b/src/game/bg_pmove.c
@@ -3082,10 +3082,6 @@ void PM_AdjustAimSpreadScale(void)
 		{
 			viewchange = 0;
 		}
-		else if (viewchange > (AIMSPREAD_VIEWRATE_RANGE / wpnScale))
-		{
-			viewchange = AIMSPREAD_VIEWRATE_RANGE / wpnScale;
-		}
 
 		// now give us a scale from 0.0 to 1.0 to apply the spread increase
 		viewchange = viewchange / (AIMSPREAD_VIEWRATE_RANGE / wpnScale);


cakeET:
Some reference info on spread change each frame at 100 FPS when turning at a steady rate:
>> with current code 
    180 deg/sec:  +1.87
    360 deg/sec:  +4.67 (this is the clamped maximum)
    720 deg/sec:  +4.67 (this is the clamped maximum)
>> without the growth cap
    180 deg/sec: +1.87
    360 deg/sec: +9.09
    720 deg/sec: +23.47

PR #2015, opened 2022-05-23
Allow changing `cg_swingSpeed` when cheats are disabled

### LINKED ISSUE #1992: Adjust default `cg_swingspeed` to better reflect actual player direction
[ET (like RTCW) defaults to `cg_swingSpeed 0.1`.](https://github.com/etlegacy/etlegacy/blob/b41b3c965ce8aba9e330d374f07b37fe092dc7d6/src/cgame/cg_main.c#L442) Quake 3 (and most other derivative games) used `cg_swingSpeed 0.3`. This cvar is used to smooth out third-person direction changes. Higher values result in more instant visual feedback, but (logically) look less smooth.

I've always found player animations to look a bit strange in ET when quickly changing directions (either when strafing during fights, or when players are sprinting to exit their spawn). I don't get the same impression in other Quake 3-based titles, which makes me think the swing speed value is too low.

Maybe we could consider using an intermediate value like `0.15` or `0.2`?

**Note:** `cg_swingSpeed` is cheat-protected, so you need to start a local server with `/devmap` or `/sv_cheats 1` to adjust it. Use `/cg_thirdPerson 1` then move around to see the effect of adjusting this cvar.

Calinou:
**Note:** Untested, since I don't have a local buildsystem setup.

This can be used to adjust how fast third-person player animations adjust to players' direction and rotation changes.

The value can be adjusted between 0.1 and 0.3. 0.1 remains the default value (same as vanilla ET and RTCW), while 0.3 is the value used in Q3A.

This closes https://github.com/etlegacy/etlegacy/issues/1992.

### DIFF
--- a/src/cgame/cg_players.c
+++ b/src/cgame/cg_players.c
@@ -1745,6 +1745,7 @@ static void CG_PlayerAngles(centity_t *cent, vec3_t legs[3], vec3_t torso[3], ve
 	else
 	{
 		float clampTolerance;
+		float swingSpeed = Com_Clamp(0.1f, 0.3f, cg_swingSpeed.value);
 
 		legsAngles[YAW] = headAngles[YAW] + cent->currentState.angles2[YAW];
 
@@ -1762,7 +1763,7 @@ static void CG_PlayerAngles(centity_t *cent, vec3_t legs[3], vec3_t torso[3], ve
 		}
 
 		// torso
-		CG_SwingAngles(torsoAngles[YAW], 25, clampTolerance, cg_swingSpeed.value, &cent->pe.torso.yawAngle, &cent->pe.torso.yawing);
+		CG_SwingAngles(torsoAngles[YAW], 25, clampTolerance, swingSpeed, &cent->pe.torso.yawAngle, &cent->pe.torso.yawing);
 
 		// if the legs are yawing (facing heading direction), allow them to rotate a bit, so we don't keep calling
 		// the legs_turn animation while an AI is firing, and therefore his angles will be randomizing according to their accuracy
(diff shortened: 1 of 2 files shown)


src/cgame/cg_players.c
```diff
@@ -1761,6 +1761,8 @@ static void CG_PlayerAngles(centity_t *cent, vec3_t legs[3], vec3_t torso[3], ve
 			clampTolerance = 60;
 		}
 
+		Cvar_CheckRange(cg_swingSpeed, 0.1, 0.3, qfalse);
```
Aciz: `Cvar_CheckRange` is an engine-only function, can't use that here. We have `Com_Clamp(min, max, value)` to use in mod code for range checking.

```cpp
Com_Clamp(0.1f, 0.3f, cg_swingSpeed.value);
```

Aciz (changes requested):
Something to note is that we have server side MDX model support, which has an adaptation of `CG_PlayerAngles` function. This uses a hardcoded swing value of `0.1f`. I'm not familiar with this code myself, but we probably need to test how this functions with this change.

https://github.com/etlegacy/etlegacy/blob/1679f1bc482f65b65e1e84a6dd8ebca5c09288ee/src/game/g_mdx.c#L1915-L2150

src/cgame/cg_players.c
Aciz: So, what you should do is for example this, at the start of this else block

```cpp
float swingSpeed = Com_Clamp(0.1f, 0.3f, cg_swingSpeed.value);
```

And then you replace the `cg_swingSpeed.value` parameters at the `CG_SwingAngles` function calls with this `swingSpeed` variable.

Aciz (approved):

Aciz:
👍 need to make sure this works fine with server side mdx, but other than that LGTM!

PR #2016, opened 2022-05-23
allows per pixel crosshair position adjustment

cakeET:
Currently the effect size of changing cg_crosshairX and cg_crosshairY depends on the client resolution and per pixel adjustments are not possible on higher resolutions.

This patch moves the translational transform to after CG_AdjustFrom640 so that cg_crosshair<X|Y> behaves consistently and allows single pixel adjustments at all resolutions.

### DIFF
--- a/src/cgame/cg_camera.c
+++ b/src/cgame/cg_camera.c
@@ -284,11 +284,12 @@ void CG_CameraEditorDraw(void)
 		                  0, 0, ITEM_TEXTSTYLE_SHADOWED, &cgs.media.limboFont2);
 
 		// render crosshair
-		x = cg_crosshairX.value;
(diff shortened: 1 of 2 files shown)


cakeET:
existing: cg_crosshairY moves the crosshair ~2.2px per unit at 1920x1080
patch: cg_crosshairY moves the crosshair 1px per unit at all resolutions

Crosshair placement is still off-center by default at most resolutions but this will allow for players to make a correction on the client until it is fixed in the source.

cakeET:
Now also allows for sub-pixel crosshair position adjustment by using the float value of the cvars instead of int.

Aranud:
Thx !

PR #2018, opened 2022-05-24
mod: fix drawing of several cursorhints

Aciz:
This fixes several cursorhints having drawing issues by changing the texture map stage to use `clampMap`. Unfortunately due to shader parse order weirdness, I had to make new shaders for these and register under different names.

The fixes are basically eliminating these lines from certain cursorhints due to the texture not being aligned properly.
image

### DIFF
--- a/etmain/scripts/legacy_gfx_2d.shader
+++ b/etmain/scripts/legacy_gfx_2d.shader
@@ -45,6 +45,112 @@ gfx/2d/net
 
 
 //========================================//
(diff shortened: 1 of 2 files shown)


isRyven (approved):

suburbski (approved):

Aranud (approved):

Aranud:
Council of reviewer approved this Pull Request

Issue #714, opened 2020-03-05
Fix flamethrower


ETLegacyTeam:
Author: @saukko28 (Saukko)
Date: 2014-12-13
Redmine Issue: 714, https://dev.etlegacy.com/issues/714

---

The demo attached shows how buggy this weapon is. Ungiven kills, falme shader bug, unresponsive hits onto heal needing enemy.

-*S

---

Related issue(s): #248, #455, #507, #668, #876

---

Related commit(s):
75d073b3ac3caf00e8e4a555d5b94b88bb7fd33f - game: Fix flamer selfkill message, adjust flamer s..
 b02e18bf0bc23f62b7cbf897cc6e0167c9761916 - mod: synced FLAME_START_MAX_SIZE - flamer fire act..
 feb772ccb7e1cb7fe481b0b59dca4c51f3cd06c8 - cgame: capped global accuracy stats, refs 478

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2015-01-25T10:30:09Z

---

Kills are now correctly given, -but the flamethrower is  currently very overpowered- (hu, no, that seems correct) (see last demo).
It's also possible to make more "hits" than "shots", resulting in crazy accuracy statistics.

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2015-01-25T14:21:46Z

---

And also to be noted that if the player is halfway dead and lying on the ground, flamer isn't always so responsive. As you might see, the player doesn't seem to get any damage on the ground (you can't see the flamer shader). Or in the other hand if it takes damage and you can see the flamer shader on the lying body, it doesn't gib him from the first shot.

-*S

Helldembez:
There seems to be a bug with proning on a slope as well. Making the flame invisible while still getting hit.

Noticed it in caen, flamethrower was proning on the slope towards documents building and the flames were invisible. Might also have to do with the nearby corner? Like when you run around the corner into an already fired flame that doesn't render?

Vorschreibung:
> There seems to be a bug with proning on a slope as well. Making the flame invisible while still getting hit.
> 
> Noticed it in caen, flamethrower was proning on the slope towards documents building and the flames were invisible. Might also have to do with the nearby corner? Like when you run around the corner into an already fired flame that doesn't render?

Verified, this seems to happen as `muzzlePoint` ↓
[...]

PR #2019, opened 2022-05-24
cgame: fix flamethrower invisible flame animation, refs #714

ryzyk-krzysiek:
Fixes cases where when start point of flame was in solid, the flame animation would not be shown.

Old: https://streamable.com/plyb3d
New: https://streamable.com/37772j

refs #714


Aranud:
Neat

PR #2021, opened 2022-05-26
mod: allow loading loose ttf/otf fonts in pure mode

Aciz:
Allow loading `.ttf` files in pure mode, so ETL clients can replace the default fonts with custom ones if they want to.

### DIFF
--- a/src/cgame/cg_loadpanel.c
+++ b/src/cgame/cg_loadpanel.c
@@ -299,10 +299,14 @@ void CG_DrawConnectScreen(qboolean interactive, qboolean forcerefresh)
 
 	if (!bg_loadscreeninited)
 	{
+		char *font1, *font2;
 		trap_Cvar_Set("ui_connecting", "0");
 
-		RegisterFont("ariblk", 27, &cgs.media.bg_loadscreenfont1);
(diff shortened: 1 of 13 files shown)


Aciz:
Added `com_customFont1/2` cvars so you don't need to rename your fonts explicitly to `ariblk.ttf` and `courbd.ttf`.

mittermichal:
neat. Add some documentation by menu tooltips or wiki?

Aciz:
> neat. Add some documentation by menu tooltips or wiki?

Added a cvar description via `Cvar_GetAndDescribe`, once it's merged and released into stable build, I'll wriite up a wiki page explaining usage :+1:

Aciz:
Moved the cvars to cgame, since this requires support from mod as well as engine. Cvars are only registered for ETLegacy clients.

Also added support for `.otf` fonts as they are supported by FreeType as well.

TODO: ensure fallback to default fonts in case invalid font is specified, and a simple command to list available fonts.

Aciz:
This should be ready now. So the complete feature set is:

* loose TTF/OTF files are now allowed in pure mode
* `cg_customFont1/2` cvars to change fonts
* `listfonts` command to list available fonts, which includes "documentation" on usage
* exclusive to ET: Legacy clients because well, needs FreeType support

PR #2022, opened 2022-05-28
remove extra randomness from SMG type weapons

cakeET:
Replaces the random element from the amount of spread added to each shot when firing an SMG type weapon with a higher base amount. Increases the base amount of spread for each affected weapon when firing to reflect the current average value with random.

Affected weapons: 
MP40
THOMPSON
MP34
STEN
FG42
FG42_SCOPED

This should make the outcome of fights with an SMG determined more by player skill than good or bad luck with a random number generator.

Note: The bullet end point for SMGs is still randomized based on total spread like most other weapons. This was a second kind of randomness that seems unnecessary.


mittermichal:
maybe this should be play tested in 3v3 or 6v6 match?

cakeET:
I'm all for testing but I don't think the effect size of this change is large enough that it would be noticeable is a few test matches. The effect of random bullet endpoint spread makes it this change hard to spot. There's no reason not to do it but I don't think it'd be very useful.

To get a feel for the change I'd suggest loading up a test server both with and without this patch and shooting at a stationary target at the same distance. With the patch you should notice that for continuous fire (trigger held down) the number of shots that land is slightly more consistent than with the current code.

Over the long term players might notice that they have had fewer encounters where their gun was unusually inaccurate and fewer times where it felt more accurate than usual.

The best case scenario for 4 shots (no extra random spread) and worst case (maximum extra random spread) should both happen about 1 in 10,000 times. Most of the time, the extra spread with the current code and the patch will be approximately the same. The patch just fixes the improbable but inevitable edge cases that pop up.

HeDo88TH:
I think it's not a good idea to touch on the basic mechanics of the game. The impact of this change could be substantial and significantly change the style of play. I believe the competitive community should vote on this before merging this PR.

cakeET:
Based on your comment I don't think you understand what this patch would do. This patch standardizes the growth in spread per shot from SMG type weapons, it does not get rid of spread for SMGs or get rid of end point randomness for SMGs.

Try it yourself and see if you can even tell a difference.

cakeET:
If this change is implemented it would allow for moving crosshair pulsing and the sniper spread scale to use client generated values for more real-time feedback on spread instead of waiting for a server response. The current delay in spread feedback gives an advantage to lower ping players.

Without this random element the spread calculation on the client will be able to match the server calculation except during packet loss.

HeDo88TH:
Now I see what you mean. It could be good to standardize it. We should actually test it before merging, I know that 0.5 is not that much but it's important to evaluate the impact on the gameplay.

Helldembez:
If we want to proceed with this we need this version deployed on a server with comp config. Then the comp community can play test it. Not sure how we would build fork branches though. I've build it locally and will try to set up a practice match.

Helldembez:
We did a play test and some people noticed the bullets don't go where they used to, for now I think its better to stay with what we have unless someone else can bring new insights.

Aranud:
Close for now, as it unfortunatly change the gameplay (Feel-play) too much.

PR #2025, opened 2022-05-30
game: make head collision bbox same size as hitbox

ryzyk-krzysiek:
This should help rotating when prone on most slopes.

refs #1208


Aciz:
Very neat

PR #2029, opened 2022-05-31
game: fix prone leg and head collision when rotating, refs #1208

ryzyk-krzysiek:
Note: this ( https://github.com/etlegacy/etlegacy/pull/2025 ) is an important part, else it will still be pretty bad for head collision.

refs #1208


Aciz:
Extremely neat

PR #2040, opened 2022-06-09
Tweak teamkill notification message to avoid being fakeable by players

Calinou:
Previously, a player called "^1TEAMMATE ^7something" could make it look like you killed a teammate, even if they are an enemy.

The new form colors the "You killed" part as well, so the capitalization of "teammate" was removed as it felt redundant.

**Note:** Translations will need to be updated to follow this change.

## Old kill notification

*It looks like a teamkill notification…*

image

## What it looked like in the console

*…but it's not a teamkill. Actual teamkill messages are already fully written in red (even before this PR).*

image

### DIFF
--- a/src/cgame/cg_event.c
+++ b/src/cgame/cg_event.c
@@ -213,7 +213,7 @@ static void CG_Obituary(entityState_t *ent)
 			}
 			else
 			{
-				s = va("%s %s", CG_TranslateString("You killed ^1TEAMMATE^7"), targetName);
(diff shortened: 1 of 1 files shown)


u1ps:
another possiblity is adding teamkill in front of line:
TEAMKILL! You killed TEAMMATE neu-rah
TEAMKILL! TEAMMATE neu-rah was killed by TEAMMATE neu-rah

PR #2042, opened 2022-06-10
cmake: generate linux binaries with extension

Aciz:
Append `.x86` or `.x86_64` suffix to Linux binaries, indicating the architecture. Linux is the only platform with both x86/64 binaries, and in order to have both installed, you currently need to either install twice, in separate directories, or manually rename the binaries.

The check for "Linux" seems very stupid, but I couldn't find any other way to detect. Let me know if there's a better way and I'll edit.

Also, there are probably scripts and stuff that need to be updated now, I will take a look at those later.

### DIFF
--- a/cmake/ETLBuildClient.cmake
+++ b/cmake/ETLBuildClient.cmake
@@ -95,6 +95,14 @@ set_target_properties(etl PROPERTIES
 	MACOSX_BUNDLE_INFO_PLIST ${CMAKE_SOURCE_DIR}/misc/Info.plist
 )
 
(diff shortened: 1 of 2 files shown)


rafal1137 (commented):
Was it not possible making `SUFFIX` outside of set_target_properties ?

Aciz:
> Was it not possible making `SUFFIX` outside of set_target_properties ?

Not sure, from what I understood it needs to be set in `set_target_properties`, and you can't call it multiple times, but I might be wrong.

Aciz:
Seems like I was wrong indeed, can I can simply do another `set_target_properties` call. Simplified the detection.

PR #2043, opened 2022-06-11
client: reduces latency, improves debugging output, creates error correction flags

cakeET:
Two bit flags were added to the user command packet to notify the server that the client sense of time was adjusted. This allows for special handling of `serverTime` in the next command received. Temporary debugging output was added to mod code in bg_pmove.c with g_developer 2. This change gives legacy and other mods the information needed to properly respond to incorrect `serverTime` values caused by automatic adjustments to `serverTimeDelta` on the client.

---

Creates`cl_extrapolationMargin` with a default value of 1 and allowed range 0 to 10 to replace a magic number 5 that added unnecessary latency at some client frame rates. This change will typically reduce the client latency by up to 4 ms at the default value. Servers that want to restrict client control can use `sv_cvar` to force default.

When the client sense of time in incremented is now more cautious to accommodate this change instead of always moving forward every snapshot.

### DIFF
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -585,7 +585,7 @@ static cvarTable_t cvarTable[] =
 
 	// Engine mappings
 	
-	{ &int_cl_extrapolationMargin, "cl_extrapolationMargin",     "0",           CVAR_ARCHIVE,                 0 },
+	{ &int_cl_extrapolationMargin, "cl_extrapolationMargin",     "1",           CVAR_ARCHIVE,                 0 },
 	{ &int_cl_maxpackets,          "cl_maxpackets",              "125",         CVAR_ARCHIVE,                 0 },
 	{ &int_cl_timenudge,           "cl_timenudge",               "0",           CVAR_ARCHIVE,                 0 },
 	{ &int_m_pitch,                "m_pitch",                    "0.022",       CVAR_ARCHIVE,                 0 },
@@ -824,17 +824,6 @@ void CG_UpdateCvars(void)
 					else if (cg_errorDecay.value > 500.0f)
 					{
 						trap_Cvar_Set("cg_errorDecay", "500");
-					}
-				}
-				else if (cv->vmCvar == &int_cl_extrapolationMargin)
-				{
-					if (int_cl_extrapolationMargin.integer < -2)
-					{
-						trap_Cvar_Set("cl_extrapolationMargin", "-2");
-					}
-					else if (int_cl_extrapolationMargin.integer > 5)
-					{
-						trap_Cvar_Set("cl_extrapolationMargin", "5");
 					}
 				}
 			}
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -281,7 +281,6 @@ vmCvar_t demo_teamonlymissilecam;
 vmCvar_t cg_predefineddemokeys;
 #endif
 
-vmCvar_t int_cl_extrapolationMargin;
 vmCvar_t int_cl_maxpackets;
 vmCvar_t int_cl_timenudge;
 vmCvar_t int_m_pitch;
@@ -584,8 +583,7 @@ static cvarTable_t cvarTable[] =
 #endif
 
 	// Engine mappings
-	
-	{ &int_cl_extrapolationMargin, "cl_extrapolationMargin",     "1",           CVAR_ARCHIVE,                 0 },
+
 	{ &int_cl_maxpackets,          "cl_maxpackets",              "125",         CVAR_ARCHIVE,                 0 },
 	{ &int_cl_timenudge,           "cl_timenudge",               "0",           CVAR_ARCHIVE,                 0 },
 	{ &int_m_pitch,                "m_pitch",                    "0.022",       CVAR_ARCHIVE,                 0 },
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -281,6 +281,7 @@ vmCvar_t demo_teamonlymissilecam;
 vmCvar_t cg_predefineddemokeys;
 #endif
 
+vmCvar_t int_cl_extrapolationMargin;
 vmCvar_t int_cl_maxpackets;
 vmCvar_t int_cl_timenudge;
 vmCvar_t int_m_pitch;
@@ -583,6 +584,8 @@ static cvarTable_t cvarTable[] =
 #endif
 
 	// Engine mappings
+	
+	{ &int_cl_extrapolationMargin, "cl_extrapolationMargin",     "0",           CVAR_ARCHIVE,                 0 },
 	{ &int_cl_maxpackets,          "cl_maxpackets",              "125",         CVAR_ARCHIVE,                 0 },
 	{ &int_cl_timenudge,           "cl_timenudge",               "0",           CVAR_ARCHIVE,                 0 },
 	{ &int_m_pitch,                "m_pitch",                    "0.022",       CVAR_ARCHIVE,                 0 },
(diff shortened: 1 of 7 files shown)


cakeET:
`CL_AdjustTimeDelta()` works by rolling the sense of time forward 1ms every snapshot that didn't require extrapolation and backward 2ms for snapshots that were extrapolated.

So, without other network lags or delays, the client sense of time will tend to cycle over three snapshots with a sense of present changing by -2ms, +1ms, +1ms.

The issue with the existing code is a 5ms offest is added that triggers the extrapolation flag even when no extrapolation actually occurred. So, the actual snapshots senses of time are offsets by -7, -6, -5.

With `cl_extrapolationMargin 0` the offsets are -2, -1, 0. (One snapshot 2ms behind, one snapshot 1ms behind, one snapshot is right on time.)

With `cl_extrapolationMargin -1` the offsets are -1, 0, 1. (One snapshot 1ms behind, one snapshot right on time, one snapshot is 1ms extrapolated.)

With `cl_extrapolationMargin -1` the offsets are 0, 1, 2. (One snapshot right on time, one snapshot is 1ms extrapolated, one snapshot is 2ms extrapolated.)

localhost at 100 FPS:
margin

Zoom way in to see the sporatic single extrapolated frames with a margin of -1 and -2.

src/client/cl_input.c
```diff
@@ -1046,6 +1046,8 @@ void CL_FinishMove(usercmd_t *cmd)
 	}
 }
 
+#define MASK_CGAMEFLAGS_SERVERTIMEDELTA 0b00000110
```
ensiform: Binary literals are not part of C standard yet FYI. This is a gnu/clang extension if it works.

src/cgame/cg_main.c
```diff
@@ -823,6 +826,17 @@ void CG_UpdateCvars(void)
...
+				{
+					if (int_cl_extrapolationMargin.integer < -2)
+					{
+						trap_Cvar_Set("cl_extrapolationMargin", "-2");
+					}
+					else if (int_cl_extrapolationMargin.integer > 5)
+					{
+						trap_Cvar_Set("cl_extrapolationMargin", "5");
+					}
+				}
```
ensiform: Don't bother adding these to cgame, you don't use it here. Cvar_CheckRange() call after the Cvar_Get in cl_main is better option for range checking an engine cvar for all mods.

Example: https://github.com/etlegacy/etlegacy/blob/master/src/renderer/tr_init.c#L1066

ryzyk-krzysiek:
The default `cl_extrapolationMargin` value of `0` could be causing unwanted behaviour on different settings - `com_maxfps 125` `sv_fps 40`, it gets even worse on `sv_fps 100`. https://streamable.com/3p5m7q

cakeET:
I think this is a real effect and not just something to do with the lagometer representation though it is important to make comparisons at the same resolution for the same level of detail.

It kind of makes sense though, if the snapshot hasn't arrived before the next frame is run, it can't be used for interpolation. When the server frame time and client frame times don't sync well some overrun/mismatch is expected.

Here's a rough estimate of what I think you're seeing at `sv_fps 40` and `com_max 125` with `cl_extrapolationMargin 0`:
```
snapshot A -- 0000
    ...
snapshot B -- 0025
    client frame  1 -- 0000-0008 AB
    client frame  2 -- 0008-0016 AB
    client frame  3 -- 0016-0024 AB
    client frame  4 -- 0024-0032 AB  << this may show as 7 ms of extrapolation in lagometer
snapshot C -- 0050 << arrives at ~25ms
    client frame  5 -- 0032-0040 BC  << probably prediction errors to roll off here
    client frame  6 -- 0040-0048 BC
    client frame  7 -- 0048-0056 BC << this may show as 6 ms of extrapolation in lagometer
snapshot D -- 0075 << arrives at ~50ms
    client frame  8 -- 0056-0064 CD  << probably prediction errors to roll off here
    ...
```

Now the same setup but with the old fixed value, `cl_extrapolationMargin 5`:
```
snapshot A -- 0000
    ...
snapshot B -- 0025
    client frame  1 -- 0000-0008 AB
    client frame  2 -- 0008-0016 AB
    client frame  3 -- 0016-0024 AB
snapshot C -- 0050  << arrived at ~20 ms (because of the 5ms cushion)
    client frame  4 -- 0024-0032 AC  << no extrapolation in lagometer
    client frame  5 -- 0032-0040 BC
    client frame  6 -- 0040-0048 BC
snapshot D -- 0075 << arrives at ~45 ms (because of the 5ms cushion)
    client frame  7 -- 0048-0056 BD
    ...
```
This effect still occurs in the existing code with `sv_fps 20` and `com_maxFPS 125`, just much less often. You need to run at fairly high resolution to see it in the lagometer but it happens anytime the server and client frametimes are ever desynched by more than 5ms.

Since 125 FPS is still somewhat standard **it is probably best to leave the default at 5ms** then and let people change it to suit their extrapolation taste and compatibility with their `com_maxFPS` setting. It might also be good to extend the upper limit to 10ms, that would accommodate  players who never want any extrapolation ever (at 100 FPS and above anyway). It might be trying to have the client re-run the frame if a new snapshot arrives after it is ready but before it is used when the client is spinning in `Com_Frame()` and the frame was extrapolated.

The higher the `sv_fps` the less tweaking  `cl_extrapolationMargin` probably matters in general since overall latency from interpolation is greatly reduced. Higher values (> 5) may even be preferable (similar to source engine cl_interp_ratio).

src/cgame/cg_main.c
cakeET: Makes sense.

src/client/cl_input.c
cakeET: Yeah it does work and I personally like the clarity/verbosity of it but I'll just make it a hex and an end of line comment to get the same effect within the standard.

cakeET:
@ryzyk-krzysiek Your testing help me understand the original purpose behind the fixed margin of 5 ms. Below I'll explain why it existed, where it succeeded/failed, and how I've updated the pull to make it adjustable without triggering the issues you found.

In the vanilla (existing) code, non-synchronous clients (when the client/server frame rates are not multiples) tend to overrun when the `serverTimeDelta` is incremented every snapshot. The margin of 5 ms ensures that when there is no network lag the client generally doesn't actually extrapolate at the default `sv_fps 20` for most client frame rates. Synchronous frame rates (20, 40, 100, 200) generally stay 5 ms behind while other common client frame rates eat into the margin to varying amounts. The 5 ms margin isn't quite enough at `sv_fps 40` (Try the existing code with `com_maxFPS 43` and see the extrapolation in the lagometer.)

I've updated the pull to handle non-synchronous clients slightly more intelligently. The worst case time over-run for a given server/client frame rate combo can be calculated and then used as threshold to decide when it is safe to increment the client sense of time.

**The default value for** `cl_extrapolationMargin` **is now set at 1** with an allowed range of 0-10. In general, this maintains the experience for non-synchronous clients and reduces latency by about 4 ms for synchronous clients. Players on a LAN should probably use `cl_extrapolationMargin 0` for the lowest latency. On most servers, the default value of 1 will (hopefully) give a consistent non-extrapolated render with minimal latency. On connections with high server to client jitter, players will experience less extrapolation by using a higher value of `cl_extrapolationMargin`


--------

I also reworked the debugging output for `cl_showTimeDelta` again:

`& 1` show for each snapshot without any line breaks

`spareTime (adjustmentMessage | serverTimeDelta deltaDelta drift)`

`& 2` also prints the spareTime for every frame to see how much the client over-ran the available data

`spareTime` is color coded such that:
```
white --- excess time to spare
green --- exactly on target
yellow -- margin in use
red ----- margin exhausted
```
When working correctly and no or little network lag, the `spareTime` will generally be white or green.

`& 4` adds a line break after the output for each snapshot

```
 1457950 +17 +09 +01 +18 (-2 ms | -191 18 20)
 1457975 +12 +04 +21 (none | -191 21 20)
 1458000 +13 +05 +22 (none | -191 22 20)
 1458025 +14 +06 +23 (none | -191 23 20)
 1458050 +15 +07 +24 (none | -191 24 20)
 1458075 +16 +08 +25 (none | -191 25 20)
 1458100 +17 +09 +26 (+1 ms | -190 26 21)
 1458125 +17 +09 +26 (+1 ms | -189 26 22)
```

Aranud:
Thanks you very much!

PR #2046, opened 2022-06-16
updates unix clocksource to prefer monotonic time

cakeET:
This patch replaces `gettimeofday` with `clock_gettime` inside `Sys_Milliseconds` and uses a monotonic clock if available on the system. 

The old clock getting function `gettimeofday` which was marked obsolete by POSIX in 2008. The recommended replacement `clock_gettime` is supposed to be more accurate and should be available on most systems. All systems with `clock_gettime` will have `CLOCK_REALTIME` and most will have `CLOCK_MONOTONIC`.

The Windows clocksource is already monotonic so this will give parity and eliminate rare issues caused by NTP or manual system time changes.

### DIFF
--- a/src/sys/sys_unix.c
+++ b/src/sys/sys_unix.c
@@ -158,37 +158,52 @@ void Sys_Chmod(const char *file, int mode)
 	Com_DPrintf("chmod +%d '%s'\n", mode, file);
 }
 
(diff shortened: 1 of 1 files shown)


Aranud:
Thanks !

PR #2081, opened 2022-07-28
cgame: fix CG_AddPMItem function definition

suburbski:


Aranud:
Wouldn't be better to fix the issue here : https://github.com/etlegacy/etlegacy/blob/e7af93ae82b1376f083e6e64d006d99124a75a60/src/cgame/cg_local.h#L3945 ?

suburbski:
nope because `listItem->color` is of type `vec3_t` and thus the fourth coordinate unused

https://github.com/etlegacy/etlegacy/blob/e7af93ae82b1376f083e6e64d006d99124a75a60/src/cgame/cg_popupmessages.c#L451

Aranud:
Understand, the alpha is computed on PM draw for fade effect ... so the value isn't set on init. Neat !

PR #2096, opened 2022-08-05
[WIP] general: flash window when match is about to start

mittermichal:
using extension system ported from ETe


mittermichal:
@ensiform :wave: am I doing this right?

mittermichal:
acording to ensiform its all bad, i am adding WIP to title

ensiform:
https://gist.github.com/ensiform/014869907c5e821960399c21c25b43f5

Untested but should work.

PR #2123, opened 2022-08-27
cgame: add missing commands to gamecommand list

Aciz:
`start_match`, `reset_match` and `swap_teams` were not included in game commands list after the list was refactored, which lead to them not being tab completed.

refs cf833ac1ffe9aa4eaa17b4394eb5eb8b9871c03c


src/cgame/cg_consolecmds.c
```diff
@@ -2399,6 +2399,96 @@ static const char *gameCommand[] =
...
 };
 
+/**
+ * @var gameConsoleCommand
+ * @brief Game console command list
+ *
+ * @todo List the server command
+ * find a more elegant / dynamic way to sharesonsole command list
+ */
+static const char *gameConsoleCommand[] =
```
Helldembez, line 2825: Would like to see this table be reusable instead of copied if possible

Issue #2075, opened 2022-07-25
Decals vertex modulation can go out of range


suburbski:
Occasionally happens when breakables are destroyed

```c
etlegacy/src/renderer/tr_decals.c:954:32: runtime error: -689.219 is outside the range of representable values of type 'unsigned char'
etlegacy/src/renderer/tr_decals.c:955:32: runtime error: -689.219 is outside the range of representable values of type 'unsigned char'
etlegacy/src/renderer/tr_decals.c:956:32: runtime error: -689.219 is outside the range of representable values of type 'unsigned char'
etlegacy/src/renderer/tr_decals.c:957:32: runtime error: -689.219 is outside the range of representable values of type 'unsigned char'
```

https://github.com/etlegacy/etlegacy/blob/ace50f1b7087903b04d7150970c6ee8dd08a07e8/src/renderer/tr_decals.c#L954-L957

PR #2144, opened 2022-09-15
renderer: fix decals modulation going out of range, refs #2075

ryzyk-krzysiek:
This is quite easy to reproduce by using PVS which is probably the biggest reason the issue exists too. When decal is of out PVS it will not be drawn (I think because of viewCount? iirc it is used as a way to know if something is in PVS or not in current frame, but not exactly sure). 

Because of that decal's `fadeEndTime` can go below `tr.refdef.time` and on the first frame where it goes back into PVS the `fade` will be negative and only at the end decal will be freed. 

Not exactly sure if freeing at the start is the best way to fix it but I think it's fine.

refs #2075

### DIFF
--- a/src/renderer/tr_decals.c
+++ b/src/renderer/tr_decals.c
@@ -933,6 +933,13 @@ void R_AddDecalSurface(decal_t *decal)
 		return;
 	}
 
+	// free temporary decal
+	if (decal->fadeEndTime < tr.refdef.time)
+	{
+		decal->shader = NULL;
+		return;
+	}
+
 	// get decal surface
 	srf = &tr.refdef.decals[tr.refdef.numDecals];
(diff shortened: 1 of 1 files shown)


ensiform:
This breaks player shadows when I added it to ETe fwiw.
And given that the decal code is the same in legacy it probably also applies here too.

ryzyk-krzysiek:
Yes it breaks etl too. At the time I didn't realize shadows are decals too so I haven't tested them at all, but I checked last night and noticed they are broken (dunno why though because it doesn't make much sense to me at first thought but I haven't looked into it). I think clamping `fade` variable could be done instead in this case, probably in 0-1 range is fine as I think that's what it is essentially. But I won't be dealing with that personally anymore, because I'm done with this game.

Issue #2072, opened 2022-07-24
vehicle shadow disappears


mittermichal:
Vehicle shadow disappears, but reappears when looking downwards. It works fine on 2.60b.
![2022-07-24-131052-etl_supply](https://user-images.githubusercontent.com/4035800/180644700-56e41974-0f8f-469a-a310-38a377d03262.jpg)
![2022-07-24-131050-etl_supply](https://user-images.githubusercontent.com/4035800/180644703-62f472f6-2f04-4ecd-8243-29f3bc5b429c.jpg)

rafal1137:
It does also flicker when you aim at specified angle and move around.

suburbski:
ETe has a wip branch for this, see [here](https://github.com/etfdevs/ETe/issues/74)

mittermichal:
nice, thanks!

PR #2146, opened 2022-09-16
renderer: fix decal projector culling, refs #2072

ryzyk-krzysiek:
The issue is caused by https://github.com/etlegacy/etlegacy/commit/efc95a2a77d39417e7dafe668d0902b31aaac2a4 I believe. Bringing back part of VET code fixes the issue, I think the change broke frustum culling of decals in `R_RecursiveWorldNode` - decalBits were changed without adjusting `R_RecursiveWorldNode`.

There were more changes so I'm not really sure if this is compatible with the rest of changes, especially `if (numDecalProjectors == MAX_USED_DECAL_PROJECTORS)` check (and like with the previous decals fix I'm not too familiar with the renderer logic), but it seemingly didn't break anything after some testing.

refs #2072

### DIFF
--- a/src/renderer/tr_decals.c
+++ b/src/renderer/tr_decals.c
@@ -1024,8 +1024,8 @@ void R_CullDecalProjectors(void)
 			*dp                                           = temp;
 		}
 
-		decalBits |= (1 << numDecalProjectors);
-		numDecalProjectors++;
+		numDecalProjectors = i + 1;
+		decalBits         |= (1 << i);
 
 		// bitmask limit
 		if (numDecalProjectors == MAX_USED_DECAL_PROJECTORS)


src/renderer/tr_decals.c
```diff
@@ -1024,8 +1024,8 @@ void R_CullDecalProjectors(void)
 			*dp                                           = temp;
 		}
 
-		decalBits |= (1 << numDecalProjectors);
-		numDecalProjectors++;
+		numDecalProjectors = i + 1;
```
rafal1137, line 1027: Its just same as `numDecalProjectors++` but its just shorter one

`decalBits |= (1 << numDecalProjectors);`
Did you checked if we are hitting the limit of MAX_USED_DECAL_PROJECTORS because for me it looks like the culrplit here?
With your change it gets its value from `i` ?

rafal1137:
Since my review got somehow doubled/removed.
I will repost it here.

`numDecalProjectors = i + 1`
Its just same as `numDecalProjectors++` but its just shorter one

`decalBits |= (1 << numDecalProjectors);`
Did you checked if we are hitting the limit of MAX_USED_DECAL_PROJECTORS because for me it looks like the culrplit here?
With your change it gets its value from i ?

ryzyk-krzysiek:
`numDecalProjectors++` is different. 

For the testing I was doing which was generating player shadows (3 projectors) + truck shadow (1 projector) (bug reproducing is also quite easy achieved with noclip toggling - when there is no player shadow anymore - truck shadow works all the time).

As an example: player shadow in my case was `i` 0, 1, 2. 0 and 1 was often culled here (player feet shadow):
```
if (R_CullPointAndRadius(dp->center, dp->radius) == CULL_OUT)
{
	continue;
}
```
And torso shadow (`i = 2`) would not be culled and so the decalBits would be `(1 << 2)` while on current master it would be `(1 << 0)` (`1 << numDecalProjectors`).
Truck would be `(1 << 3)` instead `(1 << 1)` overall resulting in decalBits = 12 (this fix) instead of 3 (master).

`numDecalProjectors` would be first set to 2 + 1 = 3 (on current master set to 1) and on last projector (truck) would be set to 3 + 1 = 4 (on current master set to 2). So current master would equal to 2 and this change would equal to 4 making check for `MAX_USED_DECAL_PROJECTORS` completely wrong I think.

ensiform:
@zturtleman was originally the author of the decal changes FYI.

ryzyk-krzysiek:
If he or someone else wants to give it a try and fix it properly then I'm all for it.

ensiform:
I was just pinging to show that it may need fixing in other projects that might have also merged the code originally too. And may have comments related to.

PR #2147, opened 2022-09-17
renderers: fix wrong decal projectors being culled, refs #2072

zturtleman:
Fix for #2072 based on https://github.com/etlegacy/etlegacy/pull/2146. It compiles but I haven't visually checked it due to unrelated issues running ETLegacy.

----

The wrong decal projectors may be culled if not all are visible and there are less than or equal to 32 decal projectors.

decalBits and numDecalProjectors were always set as if the active decal projectors were moved to the beginning of the list. However this only happened if there was more than 32 decal projectors.

The issue was introduced in commit 410d717e91ff7008430751cdce46842a66d3a841, "renderers: make decals work with multiple world scenes".

### DIFF
--- a/src/renderer/tr_decals.c
+++ b/src/renderer/tr_decals.c
@@ -1015,22 +1015,30 @@ void R_CullDecalProjectors(void)
 			continue;
 		}
 
-		// put all active projectors at the beginning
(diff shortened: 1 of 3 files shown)


ryzyk-krzysiek:
> It compiles but I haven't visually checked

Fixes the issue on my end. Thanks!

PR #2155, opened 2022-10-12
general: update uncrustify config to v0.75.1

Aciz:
No actual formatting changes, this simply takes the 0.75.1 default config and modifies it to match our old config.


Aranud (approved):
The config file is compliant with the last uncrustify executable (V0.75.1)

PR #2156, opened 2022-10-19
qcommon: prevent overwriting default configs

Aciz:
Don't allow `writeconfig` to write configs called `default.cfg`, `default_left.cfg` or `default_android.cfg`.


Aranud (approved):
I approve

PR #2165, opened 2022-10-25
game: fix g_corpses value checking logic

Aciz:
Fixes `g_corpses` value to act as a boolean, as it is checked to be `== 0` when deciding if a corpse is put into body queue or if a new entity is created. Since the initialization was checking for `> 0`, it was possible to init the body queue but not put any corpses to it by setting it to negative value.

### DIFF
--- a/src/game/g_client.c
+++ b/src/game/g_client.c
@@ -284,7 +284,7 @@ void InitBodyQue(void)
 	gentity_t *ent;
 
 	// no need to init when dyn BQ is set
-	if (g_corpses.integer > 0)
+	if (g_corpses.integer)
 	{
 		return;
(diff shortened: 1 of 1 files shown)


mittermichal:
i'm curious whether fix any particular bug that was noticed in-game

Aciz:
Doubt it, only "bug" this would cause is server crashing 8 entities earlier with `G_Spawn: no free entities` if `g_corpses` is set to negative value, which would have resulted in body queue being reserved but not utilized. But since most maps are nowhere near entity limits, and corpses start sinking after 10s anyway, I don't think it's very likely that this has ever caused any issues.

PR #2175, opened 2022-10-27
cgame: fix viewheight calculation for other leaning players

Aciz:
Viewheight for other leaning players was calculated from head origin, which is an absolute position in the world space, rather than an offset from the players origin, which caused leaning to be invisible to others depending on your location in the map.


ryzyk-krzysiek:
refs #1408

PR #2176, opened 2022-10-28
cgame: fix integer overflow in CG_DrawGrid

mittermichal:
This happens on android and Raspberry Pi and causes console to be spammed with:
"ERROR: Com_sprintf output length 3 too short, require 4 bytes."

### DIFF
--- a/src/cgame/cg_commandmap.c
+++ b/src/cgame/cg_commandmap.c
@@ -600,7 +600,8 @@ static void CG_DrawGrid(float x, float y, float w, float h, mapScissor_t *scisso
 	}
 	else
 	{
-		char   coord_char[3], coord_int;
+		char   coord_char[3];
+		signed char coord_int;
 		float  text_width, text_height;
 		vec2_t textOrigin;
 


src/cgame/cg_commandmap.c
```diff
@@ -600,7 +600,8 @@ static void CG_DrawGrid(float x, float y, float w, float h, mapScissor_t *scisso
 	}
 	else
 	{
-		char   coord_char[3], coord_int;
+		char   coord_char[3];
+		signed char coord_int;
```
Aranud, line 604: ~Something look wrong on line below (line 652)~
```c
		for (coord_int = -1; grid_y < dim_y[1]; grid_y += step[1], coord_int++)
		{
			if (coord_int >= 0)
{ [...]
```

~The variable `coord_int` is now signed, this means it can't be negative anymore.
The implicit conversion will change the value -1 to 255.
This made the next "if" statement useless and the first call of the code below the "if" statement isn't excuted anymore.~

Explicit signed variable look fine. I'm curious how it resolved the thing.

Aranud (dismissed):

src/cgame/cg_commandmap.c
mittermichal: I guess it has something to do with optimization on arm architecure, I have no idea how that works

PR #2179, opened 2022-11-01
cgame: add spawnpoint visualization

Aciz:
Adds option to visualize spawnpoints in the map with `cg_drawSpawnpoints`. Displays major spawnpoint name & number, and individual spawnpoints where players spawn. If `id` key is present in the spawnpoint, it's visualized as well. Only spawnpoints available for your team are drawn. Available in warmup, or if cheats are enabled.

2022-11-01-161515-etl_adlernest
2022-11-01-161536-etl_adlernest
2022-11-01-161559-radar

### DIFF
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -353,6 +353,8 @@ vmCvar_t cg_drawBreathPuffs;
 vmCvar_t cg_customFont1;
 vmCvar_t cg_customFont2;
 
+vmCvar_t cg_drawSpawnpoints;
+
 typedef struct
 {
 	vmCvar_t *vmCvar;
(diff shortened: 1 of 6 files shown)


src/cgame/cg_main.c
```diff
@@ -597,6 +599,8 @@ static cvarTable_t cvarTable[] =
 	{ &cg_activateLean,            "cg_activateLean",            "0",           CVAR_ARCHIVE,                 0 },
 
 	{ &cg_drawBreathPuffs,         "cg_drawBreathPuffs",         "1",           CVAR_ARCHIVE,                 0 },
+
+	{ &cg_drawSpawnpoints,         "cg_drawSpawnpoints",         "0",           CVAR_ARCHIVE,                 0 },
```
jackeri, line 603: Should this be a cheat cvar? Or only active in the developer mode?

Aciz: No, I think it's perfectly reasonable to have it available in warmup.

Aranud (approved):

PR #2187, opened 2022-11-05
game: don't let missiles fly through horizontal, upwards facing skyboxes

Aciz:
Notably this fixes being able to throw nades/airstrike/mortar etc. through the 2nd Axis spawn of karsiah_te2, since it lacks a tracemap. Tracemap check was removed in abf688c362c29c786be4039c49c28033d6a0b842 to allow mortar/riflenades (and other missiles for that matter) to be shot through skyboxes over the map, since in the absence of tracemap, they would be immediately removed upon hitting skybox.

Also simplified the check for missilecamera.

### DIFF
--- a/src/game/g_missile.c
+++ b/src/game/g_missile.c
@@ -648,12 +648,14 @@ void G_RunMissile(gentity_t *ent)
 
 	if (tr.fraction != 1.f)
 	{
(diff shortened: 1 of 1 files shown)


Aciz:
Need to do some additional testing regarding some edge cases, to make sure this doesn't break anything.

Aciz:
Better fix in #2193

PR #2190, opened 2022-11-07
cgame: add editcomponent command, #refs 1967

Aranud:
Add `editcomponent` command which allow user to customize from console or bind the whole new HUD component field by field.


jackeri:
Sorry, the rebase just totally broke this, just going to do a local merge to master.

PR #2194, opened 2022-11-10
game: add func_fakebrush drawing with railbox when debugging

Aciz:
Use `G_Railbox` to visualize `func_fakebrush` entities when debugging ~~with either `g_scriptDebug` or `developer`~~ by setting `g_debugHitboxes func_fakebrush`


Aciz:
Should move this to `G_DrawEntBBox`

Aciz:
Now drawn using `G_DrawEntBBox`.

PR #2195, opened 2022-11-10
cgame: use different filename for custom hud

mittermichal:
So it won't conflict with old file used by older versions of mod.


mittermichal:
issue fixed in https://github.com/etlegacy/etlegacy/commit/4f8e2ec8bd445b015157d7f746463b033bd3cb68

Issue #1967, opened 2022-04-24
HUD Rework


Helldembez:
So with the recent suggestions for HUD features we have reached a stage where it does more harm than good to add them in the current HUD code. Which leads to the necessity to refactor the HUD code to facilitate modern features and be user friendly to configure as well.

Since this is a big task to do right, we want to formulate a well thought out plan and structure. The start of this issue will contain WIP notes regarding said plan and structure.

## General actions

- [X] - Remove althudflags altogether
- [ ] - Clean shoutcaster checks through all code since we use shoutcaster hud class?
- [X] - Centralize draw functions
- [x] - Think about a way to prevent repeating LMS checks
- [ ] - Colour-blindness / impaired?

## New general HUD customization structure
HUD elements can be divided into two categories:
- **Text-based elements** (centerprint, clock, chat etc.)
- **Image-based elements** (compass, stamina bar, weaponicon etc.)

The goal is to allow generic/specialized customization of every single HUD element without having to add a cvar for each of these generic parameters. The current custom HUD system closely represents this already, and is ultimately what we should aim for with every single HUD element. Special customization, specific to a certain HUD element, stays as a cvar (e.g. `cg_numPopups`). Based off these 2 categories, we can define some generic element values for each of them.

### Text-based elements
- Generic parameters:
- [X]   - Position (x/y)
- [X]   - Size/Scale (w/h)
- [X]   - RGBA color (text / background / border)
- [x]   - Shadow
- [x]   - Style (if applicable, think of `cg_drawCrosshair` - every value is a style)
- [x]   - Left/Right align (could perhaps be merged with style)
- [X]   - Visible (Y/N) (all / background / border)
- [X]   ~~Font~~ (cg_customFont1/2 instead ?)

### Image-based elements
- Generic parameters:
- [X]   - Position (x/y)
- [X]   - Size/Scale (w/h)
- [x]   - Alpha
- [x]   - Style (if applicable, think of `cg_drawCrosshair` - every value is a style)
- [X]   - Visible (Y/N) (all / background / border)

These generic parameters are simply adjusted in a HUD file, without needing a cvar.

Some things to consider:

- [X] - Each HUD element should have a clear anchor point, preferably in the very top-left of the element (with some exceptions such as chat and popups which make more sense to have it at bottom left due to them expanding upwards vertically). When resized, the top/bottom left should stay exactly where it is, unlike now with certain HUD elements
[...]

PR #2197, opened 2022-11-16
cgame: add json hud sample file, refs #1967

Aciz:
Similar to `huds.hud` included in pk3, add a sample `huds.json` file for easy starting point for creating JSON huds by hand, if people wish to do so instead of using the in-game editor.


jackeri:
This already old :) Mby add an example just before release, so the format is locked down.

PR #2199, opened 2022-11-17
Spaces swapped with tabs

BreakinBenny:
I took the liberty of replacing many spaces with tabs, to make it look more presentable (at least I think they do).

If there are no conflicts between the other branches and this, this can be merged!


morsik:
> make it look more presentable (at least I think they do).

The problem with tabs is that they are dependent on user settings. You can't make them presentable for all people because all "presentability" will break when someone has tabs set to 2, 4 or any other number you hadn't while doing this.

Also, answer yourself one question: what problem does this solve? ;)

I'm highly against such change, mainly, because I did something very similar here of the beginning of this project - reformat code - and I regret it until today and I'll regret that until the end of my life because it made everyone life's harder and merging other code is PITA.

etmain/characters/temperate/axis/fieldops.char
```diff
@@ -1,15 +1,15 @@
 characterDef
 {
-	skin					"fieldops"
+	skin	"fieldops"
 
-	mesh					"models/players/temperate/axis/fieldops/body.mdm"
```
morsik, line 5: This change seems outside of scope of this issue?

morsik (changes requested):

etmain/characters/temperate/axis/fieldops.char
BreakinBenny: It should be fixed now. I must've been half-asleep given much of this was in a rush.

morsik: No worries! ;)

BreakinBenny:
It doesn't really fix *anything*, it mainly frees up bytes.

morsik:
@BreakinBenny well, it's inside ZIP anyway, so doesn't matter at gameplay as this will compress very nicely.
And in terms of development… well, no comment on that ;) We have a lot more stuff that takes so much more disk space.

Also, I don't want to start spaces vs. tabs flamewar, but well… spaces are better for vertical alignment because they make everything well… aligned vertically :P While tabs doesn't unless you have set exact specific tab width (in this case 8 when most developers have tat set to 4 or sometimes even 2).

Aranud:
We close this PR for now, as the gain is not significant and could bring more issue regarding indentation usage (which are currently handled by IDE itself)

PR #2201, opened 2022-11-25
qcommon: add -no-defaults option to writeconfig

Aciz:
Allows omitting default values from being written with `writeconfig` by specifying `-nd` or `-no-defaults` option.


jackeri:
I see no issue with this…

PR #2202, opened 2022-12-02
Mouse movement scaling while in UI

jackeri:
Scale ui mouse movement to match the window manager mouse movement. This patch should somewhat match that of windows or other window manager mouse movement.


Aciz (approved):
Seems good, menus are actually usable even with something like 2000dpi.

PR #2203, opened 2022-12-03
cgame: don't load hud from old mod path

mittermichal:
In case that user has old non-json hud.dat there, which will be common case, it will print error and load profile hud file. Error print would be confusing then.

The case, where user has new hud format in mod folder, should not be taken into account, because saving there was never released

### DIFF
--- a/src/cgame/cg_hud_io.c
+++ b/src/cgame/cg_hud_io.c
@@ -36,7 +36,6 @@
 #include "cg_local.h"
 #include "json.h"
 
-#define HUDS_DEFAULT_PATH "ui/huds.hud"
 #define HUDS_USER_PATH "profiles/%s/hud.dat"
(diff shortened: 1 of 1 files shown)


mittermichal:
@jackeri

jackeri:
This doesn't seem right, you removed the reference to the default huds file which is shipped inside the legacy.pk3.

jackeri:
The mod can only load .dat files from the mod path. The .hud files are specifically inside .pk3 files.

mittermichal:
oh right didn't think about that.
What will happen if user has old hud.dat in `$homepath/legacy` then?

mittermichal:
well I can also try it myself

jackeri:
Nothing, the old path, is ignored.

mittermichal:
yes and also the default path is `ui/hud.dat` not `hud.dat` I completelly missed that. Thanks

PR #2205, opened 2022-12-05
cgame: add help description for style comp cmd, refs #1967

Aranud:


src/cgame/cg_consolecmds.c
```diff
@@ -2317,15 +2317,36 @@ static void CG_ShowEditComponentHelp()
 	CG_Printf("\n\nAvailable ^3<field> ^7:\n\n%s\n", str);
 }
 
+static void CG_ShowEditComponentStyleHelp(const hudComponentFields_t *compField)
```
Helldembez: It should mention the current value as well when printing the help. Besides that I noticed something that is not part of this PR, but should be looked at. When not supplying a value e.g. `editcomponent compass style` it shows the help correctly, but it also shows a argument parsing error `int field component needs at least 1 argument <value>`. But when supplying a value out of range or wrong type(string vs int) it does not give a warning or error.

src/cgame/cg_consolecmds.c
```diff
@@ -2167,7 +2167,7 @@ static qboolean CG_SetRectComponentFromCommand(int *argIndex, hudComponent_t *co
 	char      token[MAX_TOKEN_CHARS];
 	rectDef_t *value = (rectDef_t *)((char *)comp + offset);
 
-	if ((trap_Argc() - *argIndex) < 4)
+	if ((trap_Argc() - *argIndex) <= 4)
```
Helldembez, line 2229: Off by one errors?

PR #2211, opened 2022-12-18
client: added support for ttf ingame console fonts

jackeri:
Added a new default console font (JetBrainsMono-SemiBold) and an icon font for the custom ET font icons.


src/renderercommon/tr_common.h
```diff
@@ -269,4 +269,11 @@ void R_LoadPNG(const char *name, byte **pic, int *width, int *height, byte alpha
 void R_LoadTGA(const char *name, byte **pic, int *width, int *height, byte alphaByte);
 void R_LoadSVG(const char *name, byte **pic, int *width, int *height, byte alphaByte);
 
+/*
+=============================================================
+IMAGE WRITERS
+=============================================================
+*/
+void RE_SaveTGA(const char *filename, byte *data, int width, int height, qboolean withAlpha);
```
Helldembez, line 277: Does not seem to be used?

src/client/cl_scrn.c
```diff
@@ -117,22 +117,66 @@ void SCR_DrawPic(float x, float y, float width, float height, qhandle_t hShader)
...
 
-	re.DrawStretchPic(x, y, w, h, fcol, frow,
-	                  fcol + size, frow + size,
-	                  cls.charSetShader);
+		if (!info)
+		{
+			return;
+		}
+
+		scale = cls.consoleFont.glyphScale;
```
Helldembez, line 158: Scaling still seems to be a bit off. Default size is bigger than before now. But also on different resolutions. Is that what the FIXME is about?

src/renderercommon/tr_common.h
jackeri: Was used by the tr_font.c as means to output the generate font canvas. Its commented out currently but I used it in debugging. Not hurting to have it declared on the header tho (for debugging).

src/client/cl_scrn.c
jackeri: Yes, the scaling is not 1 to 1 matching, and It can never really be so. The original font is squashed and mangled to fit a 16x16 image grid. Hard to match that perfectly and don't know if really needed.
This PR is meant as the final, ultimate and best solution.

Helldembez: Unless we already have a way to control the font size/points that I am not aware of a CVAR for it would be helpful in the ultimate solution 👍 Besides that, is there a way to reload the font for users, besides that `cl_font` that was meant for debugging?

PR #2217, opened 2023-01-03
mod: fix pmove_fixed 1 with g_antiwarp 1 and cg_optimizePrediction 1, refs #1379

ryzyk-krzysiek:
`pmove_fixed` probably could also be deleted instead too I guess. 

refs #1379


Aciz:
Marked this as draft for now since this will break things, @ryzyk-krzysiek you should push the fixes you made at some point.

ryzyk-krzysiek:
Updated with proper `pmove_fixed` fix and some related and unrelated playerstate transition bug fixes.

Issue #1842, opened 2021-11-27
Count health/ammo packs thrown/picked up


x0rnn:
This is a feature request. Count health & ammo packs thrown and picked up by teammates in the weaponstats/accstats. I guess packs picked up by yourself or enemies shouldn't count. Not as accuracy percentage, just as a number; "Health packs thrown/picked up: 20/10" for example.

PR #2232, opened 2023-01-26
game: added medic statistics to scores

MrWaggel:
Keep track of revives and picked up medkits in the /scores dump.

console_scores_norating
scores_statson

- **RG** Revives Given
- **RR** Revives Received
- **THG** Total Health Given (by medkits and revives)
- **THR** Total Health Received
- **EHG** Enemy Health Given (enemies picked up medkits)
- **EHR** Enemy Health Received

Tested it with bots, seemed to work.


src/game/g_match.c
```diff
@@ -478,7 +532,14 @@ void G_createStatsJson(gentity_t *ent, void *target)
 		cJSON_AddNumberToObject(tmp2, "gibs", ent->client->sess.gibs);
 		cJSON_AddNumberToObject(tmp2, "self_kills", ent->client->sess.self_kills);
 		cJSON_AddNumberToObject(tmp2, "team_kills", ent->client->sess.team_kills);
-		cJSON_AddNumberToObject(tmp2, "team_gibs", ent->client->sess.team_gibs);
```
Aranud, line 481: Unwanted stat deletation it seems !

x0rnn:
Can you add it to the weaponstats/accstats too? By default Alt or Ctrl key. Would be nice to also count ammo packs then.

Aranud:
refs #1842

src/game/g_match.c
MrWaggel: v46 mybad

Aranud: Fix it for you !

jackeri:
Doesn't this just add somewhat of a bloat to the dump? I mean, if we add medic specific stats to it, why not fops specific, or covi?

HeDo88TH:
> Doesn't this just add somewhat of a bloat to the dump? I mean, if we add medic specific stats to it, why not fops specific, or covi?

This is a valid point. I think medic is the most played class so he went for it. Anyways if we go down this route it would be interestng to see the ammo packs given / received. For covert you could show the number of stolen uniforms. And so on...
I think this would be interesting but we have a decision to make before merging this.

x0rnn:
Maybe just Revives Given/Received would make sense then.

rmarquis:
I'd agree this is too much for the dump summary - we need to draw a line and this goes beyond it imho.

However, this could be something to integrate in a more general stats dump later on, such as #2703.

Closing.

PR #2241, opened 2023-01-30
cgame: show player status as color in ft and allow forced white names

Helldembez:
add cg_fireteamNameColorless to clean teammate names color in fireteam and cg_fireteamStatusColors to show teammates health status better through name coloring or full row background color

See: https://streamable.com/8ycuew

### DIFF
--- a/src/cgame/cg_fireteamoverlay.c
+++ b/src/cgame/cg_fireteamoverlay.c
@@ -476,7 +476,7 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 			locwidth = 0;
 		}
 
-		if (cg_fireteamNameColorless.integer || (cg_fireteamStatusColors.integer == 1 && (ci->health <= 0 || ci->ping >= 999)))
+		if (comp->style & BIT(2) || (comp->style & BIT(3) && (ci->health <= 0 || ci->ping >= 999)))
 		{
 			Q_strncpyz(name[i], ci->cleanname, sizeof(name[i]));
 		}
@@ -631,12 +631,12 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 			break;
 		}
 
-		if (cg_fireteamStatusColors.integer)
+		if (comp->style & BIT(3) || comp->style & BIT(4))
 		{
 			fireteamMemberStatusEnum_t status = CG_FireTeamMemberStatus(ci);
 
 			Com_Memcpy(&nameColor, CG_FireTeamNameColor(status), sizeof(vec4_t));
-			if (cg_fireteamStatusColors.integer == 2 && status != NONE)
+			if (comp->style & BIT(4) && status != NONE)
 			{
 				vec4_t rowColor;
 				Com_Memcpy(&rowColor, nameColor, sizeof(vec4_t));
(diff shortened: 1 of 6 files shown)


src/cgame/cg_fireteamoverlay.c
```diff
@@ -368,6 +368,12 @@ clientInfo_t *CG_SortedFireTeamPlayerForPosition(int pos)
 
 #define FT_SPACING 4.f
 
+vec4_t textWhite  = { 1.0f, 1.0f, 1.0f, 1.0f };               // regular text
+vec4_t textYellow = { 1.0f, 1.0f, 0.0f, 1.0f };               // yellow text for health drawing
+vec4_t textRed    = { 1.0f, 0.0f, 0.0f, 1.0f };               // red text for health drawing
+vec4_t textOrange = { 1.0f, 0.6f, 0.0f, 1.0f };               // orange text for health drawing
```
Aciz: These should probably be `static const` variables?

src/cgame/cg_fireteamoverlay.c
```diff
@@ -770,6 +793,42 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
...
+		return &textOrange;
+	case WOUNDED:
+		return &textYellow;
+	case DEAD:
+		return &textRed;
+	case NONE:
+	default:
+		return &textWhite;
+	}
+}
```
Aciz: These could be static too

src/cgame/cg_local.h
```diff
@@ -3775,10 +3778,20 @@ void CG_DrawUISelectedSoldier(void);
...
+    DEAD,
+    NONE
+} fireteamMemberStatusEnum_t;
+
 // Fireteam stuff
 fireteamData_t *CG_IsOnFireteam(int clientNum);
 fireteamData_t *CG_IsOnSameFireteam(int clientNum, int clientNum2);
 fireteamData_t *CG_IsFireTeamLeader(int clientNum);
+fireteamMemberStatusEnum_t CG_FireTeamMemberStatus(clientInfo_t* ci);
+vec4_t * CG_FireTeamNameColor(fireteamMemberStatusEnum_t status);
```
Aciz: These don't really need to be here at all do they? Since you only ever use them in `cg_fireteamoverlay.c`, might as well declare them there.

Helldembez: Yes, was just doing it for consistency sake. Since nothing else was declared in the .c file.

src/cgame/cg_fireteamoverlay.c
```diff
@@ -548,11 +566,12 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
...
-	iconColor[3]  *= comp->colorMain[3];
-	textWhite[3]  *= comp->colorMain[3];
-	textYellow[3] *= comp->colorMain[3];
-	textRed[3]    *= comp->colorMain[3];
+	FT_select[3]   *= comp->colorMain[3];
+	iconColor[3]   *= comp->colorMain[3];
+	colorWhite[3]  *= comp->colorMain[3];
+	colorYellow[3] *= comp->colorMain[3];
+	colorRed[3]    *= comp->colorMain[3];
+	colorOrange[3] *= comp->colorMain[3];
```
Aranud: You are changing the alpha of global color which are used all around the code !!
You must copy the color first and then change the alpha from the copy.

jackeri:
Shouldn't this be a fireteam component flag configurable, instead of a cvar configurable?

Helldembez:
> Shouldn't this be a fireteam component flag configurable, instead of a cvar configurable?

Done. Please check if uncrustify did its job correctly. It touched more lines than I changed.

PR #2243, opened 2023-02-02
mod: fix func_fakebrush prediction, fixes #2242

### LINKED ISSUE #2242: func_fakebrush prediction broken
`func_fakebrush` collision prediction doesn't work properly, because client never gets the mins and maxs from server. The `solid` encoding doesn't work here (I don't know what it's used for, but from the looks of it it's not meant to be used with arbitrarily sized boxes)

Example that spawns a `func_fakebrush` close to flag on supply for easy testing.
```
create
{
	scriptname "collisiontestonsupplynonsymmetric"
	classname "func_fakebrush"
	origin "-250 2300 250"
	contents 1
	mins "-100 -100 0"
	maxs "25 25 100"
}
```
Edit: pasted wrong example

Adding another example to show it won't work even if mins and maxs are symmetric, if they are above 255:
```
create
{
	scriptname "collisiontestonsupplybig"
	classname "func_fakebrush"
	origin "-250 2000 250"
	contents 1
	mins "-300 -300 0"
	maxs "300 300 100"
}
```

oheikk:
Inspired by etjump, this fix repurposes origin2 and angles2 to send mins and maxs for prediction. It's a bit hacky - is there a better way?

### DIFF
--- a/src/cgame/cg_predict.c
+++ b/src/cgame/cg_predict.c
@@ -216,27 +216,6 @@ static void CG_ClipMoveToEntities(const vec3_t start, const vec3_t mins, const v
 		else
 		{
 			if (ent->eFlags & EF_FAKEBMODEL)
-<<<<<<< HEAD
-			{
-				// repurposed origin2 and angles2 to receive mins and maxs of func_fakebrush
-				VectorCopy(ent->origin2, bmins);
-				VectorCopy(ent->angles2, bmaxs);
-			}
-			else
-			{
-				// encoded bbox
-				x  = (ent->solid & 255);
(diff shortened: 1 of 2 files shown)


ryzyk-krzysiek:
`bmaxs[2]` will get overwritten
```
// client-side hitbox prediction code
if (ent->eType == ET_PLAYER && cg.bulletTrace)
{
    bmaxs[2] = CG_ClientHitboxMaxZ(ent, zu);
}
else
{
    bmaxs[2] = zu;
}
```

Aranud:
Thank you for your contribution!

Issue #2247, opened 2023-02-06
Clientinfo out of bounds access


suburbski:
On second suicide:

```c
etlegacy/src/cgame/cg_players.c:1883:36: runtime error: index 65 out of bounds for type 'clientInfo_t [64]'
```

https://github.com/etlegacy/etlegacy/blob/0ce90b2535cdab3e871a73cdb9bc293186f2786e/src/cgame/cg_players.c#L1883

PR #2259, opened 2023-02-10
cgame: fix clientinfo out of bounds access, refs #2247

ryzyk-krzysiek:
refs #2247

### DIFF
--- a/src/cgame/cg_players.c
+++ b/src/cgame/cg_players.c
@@ -1880,7 +1880,7 @@ static void CG_PlayerAngles(centity_t *cent, vec3_t legs[3], vec3_t torso[3], ve
  */
 static void CG_BreathPuffs(centity_t *cent, refEntity_t *head)
 {
-	clientInfo_t *ci = &cgs.clientinfo[cent->currentState.number];
+	clientInfo_t *ci;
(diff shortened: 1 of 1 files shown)


src/cgame/cg_players.c
```diff
@@ -1912,6 +1912,8 @@ static void CG_BreathPuffs(centity_t *cent, refEntity_t *head)
 		return;
 	}
 
+	ci = &cgs.clientinfo[cent->currentState.number];
```
Aranud, line 1915: I think I'm missing something ... but how moving the assignment here would resolved the out of bound access ?

Aranud: Corpse early outs here `if (cent->currentState.eFlags & EF_DEAD || cent->currentState.eType == ET_CORPSE)` ... everything is good

ryzyk-krzysiek: Because it happens for `cent->currentState.number >= 64` which is `ET_CORPSE` which early outs here:
https://github.com/etlegacy/etlegacy/blob/8f97f0332b18e29c02967f61b508a8eb38974ca1/src/cgame/cg_players.c#L1898-L1901

PR #2278, opened 2023-02-26
game: antilag and hitreg optimization

ryzyk-krzysiek:
Server will keep track in snapshot history who was added to it in `clientbits` field. Added engine extension (`trap_wasInPVS`) for accessing that information from mod. Trap call will return if specified client was in ent's last acknowledged snapshot (will also return `qtrue` if nums are incorrect).

When bullet trace is done we use this to look into the past and know if certain clients were in clients PVS when they shot. Based on that entities are excluded from being antilagged and from computing animations and creating head hitboxes which can save time.

It does not seem possible to fully trust PVS when it comes to using it for puting clients back in time or not because of ping (PVS is too strict in some places on some maps). Before the usercmd packet gets to server client could already be in PVS and past the corner, so he has to be put back in time even though player that shot couldn't see him. This step could be omitted and keep antilagging every single player on server for every shot, but instead I added check for players viewangle (I'm not sure if doing that is worth it though). If player is in 55-10 degree angle scaled over 30-500 distance then player will be treated as in PVS even if he is not.

PVS check should be safe for adjusting player's height or creating head bboxes regardless of ping.

Because this requires server changes it would only work on ETL server engine.

The goal is to reduce amount of time whole process of hit registration needs. For example if there is 30 players alive and player `A` shoots, 29 other players will be adjusted back in time and head bboxes will be created for them even if only just 1 player is in PVS of player `A` (so rest 28 players are impossible to hit). This change tries to reduce the amount of clients that need to be processed to that 1 player.

refs #1408

### DIFF
--- a/src/game/g_antilag.c
+++ b/src/game/g_antilag.c
@@ -520,6 +520,12 @@ static void G_AdjustClientPositions(gentity_t *skip, int time, qboolean backward
 
 		if (backwards)
 		{
+			// don't adjust if client wasn't in PVS at the time
+			if (!G_ClientWasInPVS(skip, list, qtrue))
+			{
+				continue;
+			}
+
 			G_AdjustSingleClientPosition(list, time);
 		}
 		else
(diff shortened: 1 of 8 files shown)


ryzyk-krzysiek:
Changed to draft because I need to fix `trap_wasInPVS` being called when it is not needed in `G_AdjustClientPositions` (dead players, spectators etc.), which is only an issue because of viewangle check. Also because if it is deemed good enough to be merged I would rather see it happen after 2.81.1 patch.

ryzyk-krzysiek:
Since I don't play the game for more than 6 months I won't be able to playtest it myself on real servers and I'm not 100% confident that the idea won't have issues I've decided to close this PR.

PR #2287, opened 2023-03-06
Prepare simplification for multi-arch installations

pemensik:
Remove duplicated local copies and instead prepare those files or scripts during build. Use BIN_SUFFIX variable to prepare architecture dependent builds.


pemensik:
After playing a bit with my [etlegacy RPM package](https://copr.fedorainfracloud.org/coprs/pemensik/games/), I think current way of supporting multiple binaries in single system is quite unfortunate. I think I have prepared similar installation with much cleaner CMake scripts and easier to extend potentially to more than PC platforms.

It does not keep the paths exactly same, because I think ``$ARCH`` variable is better suited than windows-like platform id on unices. But spirit should be the same and allow installation of both 32-bit and 64-bit binary in single system.

jackeri:
Looks good, merging for the release.

PR #2289, opened 2023-03-12
Find cjson on Fedora, where cmake config is missing

pemensik:
Current fedora packages do not ship cmake config on cjson-devel package in any release. That should get fixed, but allow locating of cjson via pkg-config in that case.

### DIFF
--- a/cmake/ETLSetupFeatures.cmake
+++ b/cmake/ETLSetupFeatures.cmake
@@ -383,7 +383,7 @@ endif()
 if(NOT BUNDLED_CJSON)
 	if(NOT ANDROID)
 		find_package(cJSON REQUIRED)
-		target_link_libraries(etl_json INTERFACE ${CJSON_LIBRARY})
+		target_link_libraries(etl_json INTERFACE ${CJSON_LIBRARIES})
 	else()
 		#It fails with linking CJSON_LIBRARY at Android NDK Path do it manually
 		target_link_libraries(etl_json INTERFACE ${CMAKE_ANDROID_NATIVE_LIB_DIRECTORIES}"cjson")
--- a/cmake/FindcJSON.cmake
+++ b/cmake/FindcJSON.cmake
@@ -1,8 +1,8 @@
 # Simple wrapper for locating cJSON package even if cmake config is not available
+# that is missing on Fedora cjson-1.7.14-7.fc38 for example
 find_package(PkgConfig)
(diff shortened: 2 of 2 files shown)


pemensik:
A drawback of this change is it prefers pkg-config information over cJSON cmake module, even when it is present. I am not sure how that could be avoided and FindcJSON would use it only when normal config is missing.

jackeri:
Wonder if this custom findcjson actually breaks the build on some distros (arch for instance). It would be nice to get some feedback from users of other distros on this. Also, just static linking the cjson from the bundled libs is not such a big deal as it is just a single file…

pemensik:
Number of files does not really matter. Fedora guidelines are very clear about this and cjson has its own package. I do not see a good reason for bundling it. But I have found a way to have it maximally compatible and prefer CONFIG mode, if it works. It can still use pkg-config or plain library search

runlevel5:
@pemensik is this only impacting F38?

pemensik:
It affects Fedora 38 and older. cjson-devel on [Fedora 39](https://koji.fedoraproject.org/koji/rpminfo?rpmID=34263281) already contains cmake module, starting with version [1.7.15](https://koji.fedoraproject.org/koji/buildinfo?buildID=2187237).

It can of course affect also other distributions, but [Debian](https://packages.debian.org/bullseye/amd64/libcjson-dev/filelist) had it present also on previous oldstable. I have made no extensive search across distributions how they are built. If they are built with cmake, cmake module is made. Depends what package maintainer has chosen.

pemensik:
@jackeri It is not because the size of additional code. But using static library requires all dependent packages rebuilt if serious bug or even security vulnerability were found. With shared library just the library is rebuilt and unless it modifies ABI, that's it. I think that is primary reason why Fedora requires all libraries in shared form if possible.

pemensik:
Of course [EPEL builds](https://koji.fedoraproject.org/koji/rpminfo?rpmID=29911269) are affected as well. Not sure if anyone is compiling etlegacy on RHEL or their clones.

Aciz:
> Wonder if this custom findcjson actually breaks the build on some distros (arch for instance). It would be nice to get some feedback from users of other distros on this. Also, just static linking the cjson from the bundled libs is not such a big deal as it is just a single file…

Not sure what the status of this is and if this is still being considered or what, but FWIW it doesn't seem to break anything on Arch.

PR #2290, opened 2023-03-15
cgame: fix crosshair name revealing disguised enemy covert ops

oheikk:
Since adding color support for crosshair names, the name of a disguised enemy covert ops is shown, instead of the player whose uniform was stolen.

I'm not sure if it makes sense to check it the way I do in this PR, but it works. And from what I can tell, `CG_DrawCrosshairNames` is meant to be refactored and moved to `cg_draw_hud.c` eventually anyway, so maybe this can act as a hotfix until then..?


Aranud:
Thx !

PR #2312, opened 2023-04-23
cgame: Abbreviate large scores for long time players.

winrid:
Ref: https://github.com/etlegacy/etlegacy/issues/2313

People have been playing Enemy Territory forever. Personally, I've played it since like 2005 :)

Some folks have really high XP in some servers which makes the scoreboard look messy, and some new people might look at it and think it's broken. We can make the scoreboard look nicer and maybe better show off the high XP :)

Some examples:

XP: 4500000:
etlegacy-abbreviated-scores

XP: 45000:
etlegacy-abbreviated-scores-45k

Low XP:
etlegacy-abbreviated-scores-500

Out of scope:
- The XP in the bottom left of the client.
- The XP of each player in the Server Info window.

I will address these in separate PRs.

### DIFF
--- a/src/cgame/cg_draw_hud.c
+++ b/src/cgame/cg_draw_hud.c
@@ -1503,7 +1503,7 @@ void CG_DrawXP(hudComponent_t *comp)
 		clr = comp->colorMain;
 	}
 
-	str = va("%i XP", cg.snap->ps.stats[STAT_XP]);
+	str = va("%s XP", Com_ScaleNumberPerThousand((float) cg.snap->ps.stats[STAT_XP], 2));
 
 	CG_DrawCompText(comp, str, clr, comp->styleText, &cgs.media.limboFont1);
 }
--- a/src/cgame/cg_scoreboard.c
+++ b/src/cgame/cg_scoreboard.c
@@ -560,22 +560,6 @@ static void WM_DrawClientScore_Medals(int x, int y, float scaleX, float scaleY, 
 	}
 }
 
-/**
- * @brief Shortens a score like "4560000" to "4.56M" for the scoreboard.
- * @details Two decimal places are used, unless the displayed digits after the decimal would be zeros.
- * @param[in] int
- * @return
- */
-static char *AbbreviateScore(int score) {
-    if (score > 1000000) {
-        return va("^7%6gM", Com_RoundFloatWithNDecimal(((float) score) / ((float) 1000000), 2));
-    }
-    if (score > 1000) {
-        return va("^7%6gK", Com_RoundFloatWithNDecimal(((float) score) / ((float) 1000), 2));
-    }
-    return va("^7%6i", score);
-}
-
 static void WM_DrawClientScore_Score(int x, int y, float scaleX, float scaleY, const score_t *score)
 {
 #ifdef FEATURE_RATING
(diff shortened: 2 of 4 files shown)


HiddenEyeDev:
Allows for decimal places such as 4.5k or 3.2k ( Yes I want to know if I have 200 xp there )

https://gist.github.com/HiddenEyeDev/87227fce0b3e86707e83f3a7eb68888e

winrid:
@HiddenEyeDev The existing version shows up to 2 decimal points if they are not zeros (at least it should, is it not for you?)

winrid:
@HiddenEyeDev 

I guess my test data was confusing because it was exactly 45_000 - here's a screenshot of 45_120 as an example - you'll see it renders 45.12k!

etlegacy-abbreviated-scores-4512k

Aranud:
Hey ! Your concept is nice and I think it would be great to make it reusable for new purpose on other function. 

And to ensure it fit with a maximum of usage, it should be put somewhere in common part (q_shared.c) and may look like this : 

```c
/**
 * @brief Shortens a values by thousands and keeping wanted decimal values (i.e  "4560000" to "4.56M")
 * @param[in] value The number to shorten
 * @param[in] numDecimal The number of decimal to be displayed
 * @param[in] numDecimal The number of decimal to be displayed
 * @return Down scaled value string with suffixed unit
 */
char *Com_ScaleNumberPerThousand(float value, int numDecimal)
{
	static const char *units[] = { "", "k", "M", "G", "T" };
	unsigned int      i        = 0;

	while (value > 1000 && i < sizeof(units))
	{
		value /= 1000;
		++i;
	}

	return va("%g%s", Com_RoundFloatWithNDecimal(value, numDecimal), units[i]);
}```

Aciz:
This does absolutely nothing in context of legacy mod as it will never display total amount of XP in scoreboard. This is mod code, it will not change things for other mods that show total XP on scoreboard. In legacy mod, the scoreboard shows per map or per campaign XP, total XP is only shown in `+stats` screen, so unless you're somehow racking up 1000+ XP in a map or campaign, this is basically useless, and 4 digits will fit there just fine anyway.

winrid:
Oh okay. I will fix my cross compilation and test with some non etlegacy mod servers.

winrid:
@Aranud Thanks! Will update and test.

winrid:
- Updated logic and moved to common.
- Updated HUD too while I had time.

ksnip_20230429-074159

rmarquis:
I assume it's not possible to do that for other mods without having their source code, as the XP system is implemented at the mod code level.

rmarquis:
Interesting. What is the rational of this PR being merged, considering the above changes are completely useless? Just being curious here.

Aranud:
I would like to use it to format file size loaded (shader, image, sound, etc ...) for debug purpose.
That's why I asked to create a global function to handle it.

rmarquis:
I see. Do you mind removing the call in the dashboard? It seems quite ugly considering it's just extra computation for no purpose.

As a side note: it's always a good idea to document the reasoning when merging code. Even if that takes a few extra seconds, it helps others (as well as the future you) to understand why it's been done, in case you need to clean up or debug in the future.
Edit: Or you can even edit the title of the PR, to make it less misleading.

Issue #2411, opened 2023-11-14
vchat/vtchat/vbchat lua implementation is broken, custom message does not work


x0rnn:
`vtchat VoiceOnly ClientNum 50 Vsay-String X-Location Y-Location Z-Location Vsay-Number \"Custom-Message\"`

Custom message does not work, and the default message is always colored ^2.

oheikk:
I would argue the implementation is not broken, but the Lua documentation is wrong. There's no support for `Vsay-Number` behaviour in the game at all, and the `Custom-Message` behaviour is done server-side before sending the `vchat` command to the client. The documentation also doesn't say what the magic-looking number `50` is doing in the examples. It's the colour code in decimal ASCII (the ASCII value of '2' is 50 which is why you get green).

You can get the exact same behaviour with Lua, if you replicate what the server does. For example, `vtchat 0 7 53 enemydisguised 123 -456 789` would send the correctly coloured `vtchat` "Enemy in disguise!" as if sent by client 7.

To get the `Custom-Message` behaviour, you need to do something like:
```lua
-- color 50 here doesn't matter but need to pass in something
et.trap_SendServerCommand(-1, "vchat 1 3 50 cheer")
et.G_Say(3, et.SAY_ALL, "Yippee!")
```
That's a more or less direct Lua equivalent of what the server does when a client sends `vsay cheer Yippee!`

The Lua documentation for `vchat` etc. needs to be fixed, but before that is done, I'd argue the current `vsay` etc. implementation as a whole needs a little bit of work done on it. If I recall, etjump does it quite well with custom messages, vsay number etc. Could just use their implementation and adapt the Lua on top of that.

Aciz:
FWIW our custom vsay implementation in ETJump is not perfect either, it is straight up broken with fireteam vsays since the beginning because they can have anywhere between 3-7 params IIRC, and our code doesn't parse that correctly.

x0rnn:
The previous code example worked with ETPro, which is why I based my assumption the implementation is wrong (my etpro lua code does not work on legacy).

The extra parameter that is missing on legacy (Vsay-Number) selects the vsay variant if it has multiple sounds. Is it random on legacy or always the 1st?

oheikk:
> The extra parameter that is missing on legacy (Vsay-Number) selects the vsay variant if it has multiple sounds. Is it random on legacy or always the 1st?

It's randomized on the client's side, like all vchats in legacy.

x0rnn:
[...]

PR #2412, opened 2023-11-17
mod: improve vsay commands, refs #2411

oheikk:
Improved existing features:
 - Server passes custom vsay message to client instead of using `G_Say`, which helps with log clutter and gives the yellow `:` after the vsayer's name instead of the white one
 - Randomized on the server, so all clients hear the same variant of the vsay
 - Custom messages will work when sent through Lua as well, exact syntax below

Added feature:
 - Clients can choose which variant of the vsay they want to play. Everyone hears the chosen variant.

New syntax for clients:
- `vsay(_team) <vchatNum> <voiceLine> <customMessage>`
- e.g. `vsay 0 cheer ^1Oh YIIISS!` will always play the "Oh yes!" variant of the cheer on Axis, with that cringy custom message in red
- `vsay_buddy <whitelistClass> <numWhitelistClients> <whitelistCno 1> <whitelistCno2> ... <vchatNum> <voiceLine> <customMessage>`, works similarly as above, and the whitelist filters work the same as they did before
- e.g. `vsay_buddy 1 2 5 7 1 fthealme ^6HIIL MIIII` will play the hilarious variant (on Axis) of the `fthealme` vsay to medics in the fireteam if their client number is 5 or 7, with a stupid custom message

New Lua syntax:
- `vchat <voiceOnly> <chattingClient> <voiceLine> <randomNum> <vchatNum> \"<customMessage>\"`
- `vtchat/vbchat <voiceOnly> <chattingClient> <voiceLine> <X> <Y> <Z> <randomNum> <vchatNum> \"<customMessage>\"`
- replace `vchatNum` with -1 and generate a random number in [0,1) for `randomNum` to get a randomized vsay

Lua example snippet:

```lua
et.trap_SendServerCommand(0, "vtchat 0 0 IamMedic 1 1000 1 " .. math.random() .. " -1 \"^gI am medic test\"")
```

### DIFF
--- a/src/cgame/cg_servercmds.c
+++ b/src/cgame/cg_servercmds.c
@@ -1659,7 +1659,7 @@ void CG_LoadVoiceChats(void)
  * @param[out] chat
  * @return
  */
-int CG_GetVoiceChat(voiceChatList_t *voiceChatList, const char *id, sfxHandle_t *snd, qhandle_t *sprite, char **chat)
+int CG_GetVoiceChat(voiceChatList_t *voiceChatList, const char *id, sfxHandle_t *snd, qhandle_t *sprite, char **chat, double randomNum, int vsayNum)
 {
 	int i, rnd;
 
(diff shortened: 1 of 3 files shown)


x0rnn:
`et.trap_SendServerCommand(0, "vtchat 0 0 IamMedic 1 1000 1 " .. math.random(0,1) .. " -1 \"^gI am medic test\"")`

This will only work if math.random generates 0, doesn't work for any other number (1 or if vsay has more than 2, I mean).

Edit: math.random number must be a float, not an integer.

oheikk:
> `et.trap_SendServerCommand(0, "vtchat 0 0 IamMedic 1 1000 1 " .. math.random(0,1) .. " -1 \"^gI am medic test\"")`
> 
> This will only work if math.random generates 0, doesn't work for any other number (1 or if vsay has more than 2, I mean).
> 
> Edit: math.random number must be a float, not an integer.

Yep you speficially want math.random() without arguments in this case. More specifically, you want to pass a random float between 0.0 and 0.99999... to get the desired effect. It should work as intended that way.

(see https://www.lua.org/pil/18.html for details)

Issue #2381, opened 2023-11-03
improve server browser bot filter icon


hkly-1245:
### Description of the issue (current behavior, expected/desired behavior)

The current version of the server browser features a space invader like icon to filter bots, but with a confusing tooltip about human players. 

![image](https://github.com/etlegacy/etlegacy/assets/133721553/2cf8c64c-2dcd-4aab-9b4a-2ced82530282)

Most icons design, tooltips couple describe accuratly the behavior in this menu, there is a match between them. 

Solutions could be:
- Change the icon with one involving human player representation 
Or
- Change the behavior of the browser to only keep servers with bots
Or 
- Change both the icon and the tooltip, with an icon involving both a robot and human representation.

### Motivation and use case for changing the behavior
User Interface must be as meaningfull to the user as it can get.

hkly-1245:
I am not a designer, but you get the idea: 
![image](https://github.com/etlegacy/etlegacy/assets/133721553/237d7208-ccb7-4aae-8730-301d2e8d4435)

winrid:
I think the icon is too small to illustrate a human and a bot. The button has three states, off, bots only, humans only. The tooltip seems pretty descriptive to me. 

Probably the text should just be changed. When you select "exclude bots" it says "Don't Show..." Which is confusing.

Instead we can just make it say "Bots Only" or "Humans Only" depending on checkbox state.

winrid:
Gonna take a stab at this this Sunday

winrid:
@hkly-1245 I think the core issue behind this is resolved? LMK: https://www.youtube.com/watch?v=HErivAFFd_8

hkly-1245:
> @hkly-1245 I think the core issue behind this is resolved? LMK: https://www.youtube.com/watch?v=HErivAFFd_8

Yes, except for the bot icon graphical design

PR #2417, opened 2023-11-19
etmain: Server Browser - Easier to Understand Filters

winrid:
Demo video:

https://www.youtube.com/watch?v=HErivAFFd_8

Instead of just showing "include only" or "exclude only", we can show much better short descriptions that still fit fine.

Refs https://github.com/etlegacy/etlegacy/issues/2381

Verified the filters against the vars from the servers for correctness.


winrid:
Investigating how to update the other languages...  (edit - added in translation keys for all other languages...).

PR #2424, opened 2023-11-22
server: cap client rate to avoid div0

ghost:
Fixes division by zero in `SV_RateMsec()` when `sv_minRate` is `0`, which happens when a message is sent to a client who's `client_t` has been previously zeroed. I think it happens when a client is disconnected, though I'm not sure why. Anyway, I know from the stacktrace that this is what makes the server crash.


Aranud:
Thank you!

PR #2425, opened 2023-11-24
Openjk's custom entity loading

wolf6542:


src/qcommon/cm_load.c
```diff
@@ -456,15 +456,51 @@ void CMod_LoadBrushSides(lump_t *l)
...
+
+	COM_StripExtension(name, noext);
+	filename = va("%s.ent", noext);
+
+	entFileLen = FS_FOpenFileRead(filename, &file, qtrue);
+
+	if (file && entFileLen > 0)
+	{
+		cm.entityString = (char*)Hunk_Alloc(entFileLen + 1, h_high);
+		cm.numEntityChars = entFileLen + 1;
```
Aciz: `numEntityChars` is commented out in `clipMap_t` struct in ETL, this does not compile without uncommenting that.

https://github.com/etlegacy/etlegacy/blob/f1e9d75914cb983b57d7f98c04f11ce860f9fb99/src/qcommon/cm_local.h#L166

src/qcommon/cm_load.c
```diff
@@ -456,15 +456,51 @@ void CMod_LoadBrushSides(lump_t *l)
...
+		FS_Read(cm.entityString, entFileLen, file);
+		FS_FCloseFile(file);
+		cm.entityString[entFileLen] = '\0';
+		Com_Printf(va("Loaded entities from %s\n", filename));
```
Aciz: `va` should not be used for `Com_Printf`, it can handle it's formatting on it's own. Using `va` causes

```warning: format string is not a string literal (potentially insecure) [-Wformat-security]```

File: src/qcommon/cm_load.c [hunk 3]
```diff
@@ -456,15 +456,51 @@ void CMod_LoadBrushSides(lump_t *l)
...
 {
+	if (CMod_LoadCustomEntityString(name))
+	{
+		return;
+	}
+
 	cm.entityString = Hunk_Alloc(l->filelen, h_high);
 	//cm.numEntityChars = l->filelen;
 	Com_Memcpy(cm.entityString, cmod_base + l->fileofs, l->filelen);
+	Com_Printf(va("Loaded entities from %s\n", name));
```
Aciz: Same here as in the other `Com_Printf` call.

Aciz (changes requested):
After fixing compilation it does work for point entities. Spawning anything that has a bmodel will cause a crash with `CM_InlineModel: bad number`, but other than that it can be used to add spawnpoints just fine for example.

src/qcommon/cm_load.c
```diff
@@ -456,15 +456,51 @@ void CMod_LoadBrushSides(lump_t *l)
...
+ * @brief CMod_LoadCustomEntityString
+ * @param[in] name
+ */
+qboolean CMod_LoadCustomEntityString(const char* name) {
+	fileHandle_t file;
+	int entFileLen = 0;
+	char* filename;
+	char noext[MAX_QPATH];
+
+	COM_StripExtension(name, noext);
```
Aciz: Forgot about this, `COM_StripExtension` takes a size argument too
```void COM_StripExtension(const char *in, char *out, int destsize);```

Aciz (approved):
Didn't originally mean that you'd completely remove `cm.numEntityChars` but looking at both JKA and ETe implementation, it's not actually used for anything, so doubt it really needs to be there anyway.

This does have limited usage, since you can always use mapscript to spawn things, but it is good for:

* mods that don't support entity spawning via mapscript
* spawning entities on maps that don't have any entities with a spawn function (a lot of trickjump maps for example)

wolf6542:
> Didn't originally mean that you'd completely remove `cm.numEntityChars` but looking at both JKA and ETe implementation, it's not actually used for anything, so doubt it really needs to be there anyway.
> 
> This does have limited usage, since you can always use mapscript to spawn things, but it is good for:
> 
>     * mods that don't support entity spawning via mapscript
> 
>     * spawning entities on maps that don't have any entities with a spawn function (a lot of trickjump maps for example)

Yep, I looked at it briefly, didn't realize the var was useless before so got rid of it.

Issue #2435, opened 2023-12-06
Remember HUD before switching to Shoutcaster


x0rnn:
Related to https://github.com/etlegacy/etlegacy/issues/2376

Once you go Shoutcaster and back to normal, it will set HUD to 0 instead of the one you had set previously. The game should remember your actual HUD. Same if you quit the game as a Shoutcaster, the next time you launch ET it should set it to your actual HUD.

PR #2436, opened 2023-12-10
cgame: add cg_shoutcasterHud (#2435)

ghost:


Aranud:
It work fine, but I think we can go further : 
* Add a cvar to choose the shoutcaster hud, so player doesn't depend on "Shoutcaster" one and remove the hard code.
* Ensure `cg_selectedAltHud` is cleared if player is no more shoutcaster or the hud has been restored. So it doesn't restore to `cg_selectedAltHud` because player has selected shoutcaster hud as current HUD.

Also, just to be picky, I think the cvar `cg_selectedAltHud` is a bit missleading with `cg_althud`. It could be renamed. 

Your thoughts ?

ghost:
Good point, changed it.

Aranud:
Neat, thank you !

PR #2439, opened 2023-12-10
game: add backward compatible log timestamp formats for parsers

ghost:


src/game/g_main.c
```diff
@@ -3812,7 +3814,39 @@ void QDECL G_LogPrintf(const char *fmt, ...)
...
+	{
+		string[0] = 0;
+	}
+	else if (g_logTimestamp.integer == 2) // startup relative time (mmm:ss)
+	{
+		int timestamp = trap_Milliseconds();
+		int m = timestamp / 6e4;
+		int s = (timestamp - m * 6e4) / 1e3;
+
+		if (m < 1e3)
```
Aranud: Could it be possible, for the sake of concistency in code, change the value format to plain integer instead of power of ?
I know this format is more sexy but it would be a shame to start write numeric values differently.

Aranud:
Thank you very much !

Issue #2422, opened 2023-11-21
VRAM usage grows after continuous vid_restarts


Aciz:
Continuously vid_restarting or changing the map causes a memory leak, each reload adds up ~70MB of VRAM usage (at 1440p, lower/higher depending on resolution). ~Same issue is present on Linux with [ETe](https://github.com/etfdevs/ETe), but not on Windows, indicating potentially an issue with SDL2 integration (ETe uses SDL2 on Linux, but not on Windows). The leak is much more prominent in ETL as compared to ETe (\~3x faster grow).~

![image](https://github.com/etlegacy/etlegacy/assets/14221121/8d7e112d-0092-4a37-8e9e-4df1b5011bcb)

Aciz:
Seems this is not the same issue as which ETe has, so we can probably forget about SDL2 being the issue here.

However I did notice from a quick test that this seems to be related to `r_fbo`. With `r_fbo 1`, the VRAM usage keeps growing, however setting `r_fbo 0` will stop it (it won't free up the extra allocated VRAM though).

Edit: `r_ext_multisample 0` will also fix it, so it's probably related to that too.

PR #2444, opened 2023-12-21
renderer: fix VRAM memory leak, refs #2422

ryzyk-krzysiek:
I don't know if the fix is exactly correct, but it works for me.

refs #2422


rafal1137:
It does crash on me when in Release.
Debug is fine

```
3x vid_restart
1457 => 1552

Ryzyk_fix
Debug Build
1457 => 1552
```
As seen there is no much of a difference

PR #2451, opened 2024-01-08
game: remove unused variable and fix possibly corrupted stack

ghost:
This somehow seems to be fixing corrupted stack around the `cmd` variable on 32-bit Windows.

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -4954,11 +4954,10 @@ void Cmd_SelectedObjective_f(gentity_t *ent, unsigned int dwCommand, int value)
  */
 void Cmd_Ignore_f(gentity_t *ent, unsigned int dwCommand, int value)
 {
-	char cmd[MAX_TOKEN_CHARS], name[MAX_NAME_LENGTH];
+	char name[MAX_NAME_LENGTH];
 	int cnum;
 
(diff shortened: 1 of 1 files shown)


src/game/g_cmds.c
```diff
@@ -4954,11 +4954,10 @@ void Cmd_SelectedObjective_f(gentity_t *ent, unsigned int dwCommand, int value)
  */
 void Cmd_Ignore_f(gentity_t *ent, unsigned int dwCommand, int value)
 {
-	char cmd[MAX_TOKEN_CHARS], name[MAX_NAME_LENGTH];
+	char name[MAX_NAME_LENGTH];
 	int cnum;
 
-	trap_Argv(0, cmd, sizeof(cmd));
-	trap_Argv(1, name, sizeof(cmd));
```
winrid, line 4961: already said in discord but I'm guessing the "some reason" is that this line should be `sizeof(name)`, since these arrays are different sizes [0]. So probably fixing that is sufficient, but if `cmd` isn't used then sure, let's remove it!

[0]
```
#define MAX_NAME_LENGTH            36
#define MAX_TOKEN_CHARS     1024 
```

ghost: Yes, I figured, just didn't originally realize the function always writes the zero byte to `dest - 1` regardless of the source length.

PR #2452, opened 2024-01-08
game: suppress fireteam invites and proposals from ignored clients

ghost:


ghost:
Updated to preserve the submitted application message.

PR #2453, opened 2024-01-10
cgame: make weapon icons in popup messages point towards the victim

ghost:
With weapon icon style obituaries, the icon always points from right to left regardless of the `Swap V<->K` setting. This change addresses that and flips the icon appropriatelly.


Aranud:
Neat, thanks you !

PR #2459, opened 2024-01-17
server: enforce temporary ban on both GUID and IP (fixes #2457)

### LINKED ISSUE #2457: Tie kick to both IP and GUID
Currently kick is tied to the IP, changing it gets you unbanned. How about linking it to the GUID also?

Sure you can generate a new one too, but it's more work and will at least stop some from bypassing the kick.

ghost:
As per the referenced issue, also adds `cleartempbans` command and fixes very edge case bug where the timeout isn't correctly converted to minutes when all the banlist slots are exhausted.

### DIFF
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -362,6 +362,7 @@ typedef struct
 typedef struct tempBan_s
 {
 	netadr_t adr;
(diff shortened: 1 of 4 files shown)


Aranud:
Neat, thank you very much

PR #2463, opened 2024-01-26
Fix build on FreeBSD (amd64 and aarch64)

alonsobsd:
I have added games/etlegacy to FreeBSD ports tree and those are some changes I applied for build on amd64 and aarch64. Take a look at https://www.freshports.org/games/etlegacy


rafal1137 (changes requested):
@alonsobsd I see its more about case sensitivity here
Can you add additional entry for small one ?
```cmake
"${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "amd64")
```

alonsobsd:
> @alonsobsd I see its more about case sensitivity here Can you add additional entry for small one ?
> 
> ```cmake
> "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "amd64")
> ```

Done

jackeri:
Work on the platform detection is already on the way on the android branch.. Wait for it to merge (maybe later this weekend).

src/renderercommon/tr_font.c
```diff
@@ -202,7 +202,7 @@ static glyphInfo_t *RE_ConstructGlyphInfo(int imageSize, unsigned char *imageOut
 
 		rc = FT_Load_Glyph(face, index, flags);
 
-		if (rc)
```
jackeri, line 205: This seems a bit of a strange change

PR #2465, opened 2024-01-27
client: use fs_homepath if provided and default home path cannot be resolved

ghost:
After uninstalling OneDrive and nuking its folder, the client refuses to start as `Sys_DefaultHomePath()` returns an error. That's *an* issue, but what bothers me more is the game ignoring the `fs_homepath` value, insisting on knowing what the default home path is.

This change lets the client start if the default home path could not be resolved as long as `fs_homepath` is set. I'm not sure whether the default in that case should be the base path or current working directory, but since we do have the `fs_homepath`, it doesn't really matter.

### DIFF
--- a/src/qcommon/files.c
+++ b/src/qcommon/files.c
@@ -4443,7 +4443,15 @@ static void FS_Startup(const char *gameName)
 #ifdef DEDICATED
 		homePath = fs_basepath->string;
 #else
-		Com_Error(ERR_FATAL, "FS_Startup: Default home path is empty.");
+		if (Cvar_Flags("fs_homepath") == CVAR_NONEXISTENT)
+		{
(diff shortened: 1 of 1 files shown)


rafal1137 (commented):
Back in a day someone decided to keep fs_homepath and fs_basepath separate on Window OS for a reason when multi users installation is available. So different users have their own space with cfgs same as in Linus OS from the beggining. Isn't it just telling game to use basepath as their homepath ?

ghost:
It's not. It's telling the game to respect whatever `fs_homepath` value I provided instead of refusing to start because the default home path (which is, again, irrelevant now since I set the `fs_homepath`) cannot be resolved. Nothing changes if `fs_homepath` isn't set during startup.

Issue #229, opened 2020-03-05
Add support of ETTV (ETpro*)


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2013-03-05
Redmine Issue: 229, https://dev.etlegacy.com/issues/229

---

ETTV allows you to connect up to a special server using your PC, and view a game in progress much like a spectator would. But it's more powerful than any normal spectator function - ETTV allows you all the features of a shoutcaster - being able to see the names of players over their heads, seeing a dynamite timer over the dynamite, etc. You can even watch spectators on the server, such as 'cameramen', people that ensure you get the best overview shots of battles, the best 1st person views during 1 on 1's, and those crucial views just before a panzer dishes out some ownage on an unsuspecting enemy group.

http://wolfwiki.anime.net/index.php/ETTV

ETLegacyTeam:
From: @jackeri (Jacker)
Date: 2013-03-06T09:17:01Z

---

as ETTV is basically just an modified eted (ET server) we should just rewrite the functionality into the server source. Not saying it is a small task but that way we dont need to upkeep another binary and every server maybe configured for tv slave support instead of servers starting up another binary.

http://wolfwiki.anime.net/index.php/ETTV:Server_Configuration

The reason i'm all for the rewrite is simple, if we just simply support the existing ettv implementation that means we basically support the usage of 2.6 servers instead of the etl servers.

As a side note: at some point when 2.60b fix was not yet released many actually used the modified ettv binary as a normal game server instead of the original server binary.

PS. i do like the sound of ETLTV.. :D

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2017-03-05T23:28:47Z

---

For reference: https://github.com/ioquake/ioq3/pull/265

Not sure how the mentioned server-side demo and multiview protocol + server-side multiview demo recording compare to our current implementation.

mittermichal:
<img width="514" alt="Screenshot 2024-10-23 at 15 33 47" src="https://github.com/user-attachments/assets/80b0d553-2218-40b6-b95a-06c4a8f28cce">

https://www.gamestv.org/event/67040-team-poland-2-vs-czech-republic/

```
$ ls legacy/tvdemos | wc -l && du -sh legacy/tvdemos
252
5.5G    legacy/tvdemos
```

;)

PR #2470, opened 2024-02-01
server: add possibility for etlded to act as a client (ettv), refs #229

ryzyk-krzysiek:
- add possibility for etlded to act as a client
- add command `tv connect <ip> <masterpassword>` (sv_etltv_maxslaves and sv_etltv_password must be set on master server)
- add command `tv disconnect`
- add command `record <name> (optional)` for recording tv demos
- add command `stoprecord` for stopping recording
- add command `demo <name>` for demo playback
- add command `ff <seconds>` command - fastforwarding demo (works similiar to timescale) stops on map_restart or can stop it with ff 0
- add `cvar sv_etltv_autorecord <0/1>` automatically record tv demos
- add `cvar sv_etltv_autoplay <0/1>` automatically play demos (only on specifically named demos from autorecording, fe. demo0001->demo0002 etc.)
- add `cvar sv_etltv_clientname` etltv client name
- demos are recorded into tvdemos folder with extension tv_84
- etpro tv demos are playable, best is to use ettv tvgame and start etlded with +fs_game etpro (put etpro tvgame into etpro mod folder). Need 2.6b client in order to watch
- slave server will not download missing files from master server, so need to make sure it has every pk3 it needs
- only available for dedicated server

Missing ettv features:

- more than 64 client servers
- chaining slave server still not tested
- delayed feed (ettv_delay), everything is live, old like ettv solution of delay by playing back demo might work (not tested)

refs #229

### DIFF
--- a/src/game/g_public.h
+++ b/src/game/g_public.h
@@ -43,49 +43,6 @@
 typedef qboolean (*addToSnapshotCallback)(int entityNum, int clientNum);
 
 /**
-  * @struct entityShared_s
-  * @brief entityShared_t
-  *
-  * @warning Don't add or remove fields to keep 2.60 compatibility
-  */
-typedef struct
-{
-	qboolean linked;                ///< qfalse if not in any good cluster
-	int linkcount;
-
-	int svFlags;                    ///<  SVF_NOCLIENT, SVF_BROADCAST, etc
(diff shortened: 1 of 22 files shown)


jackeri (approved):
Looks pretty good to me.

Issue #862, opened 2020-03-05
Fix construction sound


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2015-08-17
Redmine Issue: 862, https://dev.etlegacy.com/issues/862

---

"No construction sound when building CP or barrier on bremen map (to be checked, other people around heard the sounds) (delaying to later)"

---

Related commit(s):
9d15c9d68b576ca5ef16a9cfe2485f9c65d3da19 - mod: optionally fix construction sound (g_misc bit..
 914d1cd5b87c7c808d0766ad5e26448a0e4e8716 - game: multi client construction sounds refs 862
 bb487587cf58a7b59784e67df4edd564a5cd103a - game: set fixed construction sound as default and ..

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2015-08-17T17:49:56Z

---

The construction sounds are actually related to the entity built, not the player that builds stuff. Pliers themselves don't make any sound, but the construction sound is started when the build of entity starts.

If engineers stop to construct, the sound will still finish playing, and then resuming the construction will not play any sound.

ETLegacyTeam:
From: @IR4T4 (IR4T4)
Date: 2016-01-10T10:37:13Z

---

Spyhawk wrote:
> If engineers stop to construct, the sound will still finish playing, and then resuming the construction will not play any sound.
1.) Do we want to change this? Play the same sound on resuming?

2.) This ticket was created because of sound issues at start of constructing reported by player at our last release gaming session. Last night I did run into same. This occures when the connection is just bad (ping > 350 and higher, flickering 'connection interrupted' message is shown) .  I did connect to an australian server from europe. In fact I've had more issues than this missing construction sound: The game feels delayed in total which isn't a real surprise ...

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2016-01-12T16:29:33Z

---

IR4T4 wrote:
> 1.) Do we want to change this? Play the same sound on resuming?

I'd find more logical that the active pliers produce the sound, similarly to a weapon. This would probably require to find a shorter sound effect we can play in loop... is that worth it? Don't know, and far from critical.

> 2.) This ticket was created because of sound issues at start of constructing reported by player at our last release gaming session.

Nothing we can do for that specific issue then.

ETLegacyTeam:
From: @IR4T4 (IR4T4)
[...]

PR #2484, opened 2024-02-20
mod: construction sound no longer loops, fixes #2429

### LINKED ISSUE #2429: Don't loop construction sound
Play it only once, or restart if you stop using the pliers and then start again. Related to #862.

nanocirs:
Solved the issue where construction sound looped every 4s. Construction sound is now only triggered when starting to build.

### DIFF
--- a/src/game/g_weapon.c
+++ b/src/game/g_weapon.c
@@ -1006,8 +1006,12 @@ static qboolean TryConstructing(gentity_t *ent, gentity_t *trigger)
 
 		if (!ent->client->constructSoundTime || level.time > ent->client->constructSoundTime)
 		{
-			// construction sound sent as event (was temp entity)
-			G_AddEvent(ent, EV_GENERAL_SOUND, GAMESOUND_WORLD_BUILD);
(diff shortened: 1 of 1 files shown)


Vorschreibung:
Heya, I'm not a regular contrib here, so take my notes with a grain of salt - tried your patch and can confirm that this works. Plays sound only once the construction starts.

Even if you die and come back, as long as it is still partially constructed, will also not play the sound again until it's either destroyed or construction aborted due to idleness.

Suggest to add a comment above the conditional that mentions that `constructible->s.angles2[0]` here is (AFAIS) the `health` of the constructible (grep for `addhealth`) - and/or to mention something like `// play construction sound only at the start of construction`.

I have checked with `2.60b` and it seems to behave the exact same way now. :metal: :blush:

Vorschreibung:
The looping behavior might be wanted, see https://github.com/etlegacy/etlegacy/issues/862

Vorschreibung:
It was confirmed that looping the sound is wanted behavior to give a continuous sound hint while constructing.
AFAIS this can be closed.

PR #2573, opened 2024-05-22
game: Prevent callvote when warmup/game is about to end

Vorschreibung:
Currently callvote'ing during warmup shortens the vote time (e.g. if callvote happens with 10s of warmup remaining, instead of 30s vote time it will be reduced to 10s as well), this often leads to failed callvotes, as there's not enough time for participation as players are often AFK, especially during warmup, or simply don't have enough time to notice and react.

A common observation on pub servers with 30+ players is that once the initial vote (e.g. shuffleteams) fails during warmup due to missing participation and then the vote is called again at the start of the game right after, enough people switch their voting behavior from 'yes' to 'no', presumably as the erroneously assume that the first vote failed not due to lack of participation but due to lack of approval, that the second vote consequently fails, even tho the first vote looked likely to succeed.

Preventing callvotes when warmup is about to end was already the behavior in dc4828e6ef16bb1249775c34c6f09dfaa9f266e6 and then was changed to reducing the votetime instead in
bacade7892a1a2c1563c6897897e2854e17b6a6e.

### DIFF
--- a/src/renderer2/tr_animation_mdm.c
+++ b/src/renderer2/tr_animation_mdm.c
@@ -1678,7 +1678,6 @@ static void R_CalcBone(const int torsoParent, const refEntity_t *refent, int bon
 				{
 					diff[j] = AngleNormalize180(diff[j]);
 				}
-				//angles[j] += thisBoneInfo->torsoWeight * diff[j];
 			}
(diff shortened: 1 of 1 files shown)


Vorschreibung:
Sorry for the duplicate, current one here: https://github.com/etlegacy/etlegacy/pull/2574

PR #2574, opened 2024-05-22
game: Prevent callvote when warmup/game is about to end

Vorschreibung:
Currently callvote'ing during warmup shortens the vote time (e.g. if callvote happens with 10s of warmup remaining, instead of 30s vote time it will be reduced to 10s as well), this often leads to failed callvotes, as there's not enough time for participation as players are often AFK, especially during warmup, or simply don't have enough time to notice and react.

A common observation on pub servers with 30+ players is that once the initial vote (e.g. shuffleteams) fails during warmup due to missing participation and then the vote is called again at the start of the game right after, enough people switch their voting behavior from 'yes' to 'no', presumably as the erroneously assume that the first vote failed not due to lack of participation but due to lack of approval, that the second vote consequently fails, even tho the first vote looked likely to succeed.

Preventing callvotes when warmup is about to end was already the behavior in dc4828e6ef16bb1249775c34c6f09dfaa9f266e6 and then was changed to reducing the votetime instead in
bacade7892a1a2c1563c6897897e2854e17b6a6e.

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -3191,6 +3191,11 @@ qboolean Cmd_CallVote_f(gentity_t *ent, unsigned int dwCommand, int fRefCommand)
 				CP("cp \"You cannot call a vote as a spectator.\"");
 				return qfalse;
 			}
+			else if (g_gamestate.integer == GS_WARMUP_COUNTDOWN && (level.warmupTime - level.time) < VOTE_TIME)
+			{
+				CP("cp \"You cannot call a vote when warmup is about to end.\"");
+				return qfalse;
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/game/g_cmds.c
```diff
@@ -3191,6 +3191,16 @@ qboolean Cmd_CallVote_f(gentity_t *ent, unsigned int dwCommand, int fRefCommand)
...
+			else if (g_gamestate.integer == GS_WARMUP_COUNTDOWN && (level.warmupTime - level.time) < VOTE_TIME)
+			{
+				CP("cp \"You cannot call a vote when warmup is about to end.\"");
+				return qfalse;
+			}
+			else if (g_gamestate.integer == GS_PLAYING && (level.startTime + (g_timelimit.value * 60000) - level.time < VOTE_TIME))
+			{
+				CP("cp \"You cannot call a vote when game is about to end.\"");
+				return qfalse;
+			}
```
Aciz: I would probably revert this change and keep the old behavior for end of game scenario, at least in stopwatch mode. When playing a map with dynamite objective, players might call surrender/match reset/next map to get over with the game at that point, which would not be possible after this.

Vorschreibung: Thanks for the feedback and the review! Makes a lot of sense!

Vorschreibung:
I guess alternatively:
1. deciding between blocking or shortening could be dependent on the playercount
2. deciding between blocking or shortening could be dependent on the gamemode
3. deciding between blocking or shortening could be a server variable

However it does seem to me that a good default here would go a long way.

Aciz (approved):

PR #2575, opened 2024-05-22
cgame: Increase eventParam size 8 → 16

Vorschreibung:
fixes #2570

---

Can't confirm that it's the right way to fix it, but can confirm that this solves the linked issue :3


Aciz:
This breaks 2.60b compatibility.

Vorschreibung:
Sorry for the noise!

Aciz:
Np, generally anything in playerstate or entitystate cannot be touched, as these structs are handled in both engine and game code, so it's not possible to to add more fields to them, or change the size of things.

PR #2576, opened 2024-05-23
game: snap bullet impact ent number to max uint8

### LINKED ISSUE #2571: eventParm integer wraparound in EV_BULLET causes incorrect impacts for entities
When firing to an entity, the entity number is stored in `ent->s.eventParm`

https://github.com/etlegacy/etlegacy/blob/6935de0c07596db0fcbf9243999fb7ecc4e7a1b4/src/game/g_weapon.c#L3574-L3575

This is problematic because `eventParm` is transmitted as 8-bit integer over the net, so this causes integer wraparound in cgame when `EV_BULLET` is processed if the `entityNum` is > `255`

https://github.com/etlegacy/etlegacy/blob/6935de0c07596db0fcbf9243999fb7ecc4e7a1b4/src/qcommon/msg.c#L1156

This makes `CG_Bullet` potentially process impacts with incorrect entitynumber, if the entity number is > `255`, which can cause things like this to happen (`entityNum` wraps around to value < `MAX_CLIENTS`, so flesh impact is applied)

2024-05-20-221636-etl_base
image
image

Vorschreibung:
`eventParm` is transmitted as 8-bit integer over the wire, so to sidestep an overflow that can cause bullet impacts to wrongly appear as flesh (as the assumption is that bullet impacts on entites with number 1..MAX_CLIENTS hit player bodies) we always snap to 255 (max uint8 value)

fixes https://github.com/etlegacy/etlegacy/issues/2571

### DIFF
--- a/src/game/g_weapon.c
+++ b/src/game/g_weapon.c
@@ -3572,7 +3572,11 @@ void Bullet_Fire_Extended(gentity_t *source, gentity_t *attacker, vec3_t start, 
 
 	// send bullet impact
 	tent                   = G_TempEntity(impactPos, EV_BULLET);
-	tent->s.eventParm      = traceEnt->s.number;
+	if (traceEnt->s.number > 255) {  // snap to max uint8_t (wire size)
+		tent->s.eventParm  = 255;
+	} else {
+		tent->s.eventParm  = traceEnt->s.number;
+	}
 	tent->s.weapon         = GetMODTableData(mod)->weaponIcon;
 	tent->s.otherEntityNum = attacker->s.number;
 	tent->s.modelindex     = hitType;   // send the hit sound info in the flesh hit event


Vorschreibung:
@Aciz 2nd attempt :3

Aciz:
No, this is not correct.

Imagine a scenario where the entity with `entityNum 300` is an entity which consist of surfaces with `surfaceparm wood`. You expect that when you shoot at this entity, you get wood impact sounds. Now because this is > 255, the value gets capped to 255, so the properties of `entityNum 255` get applied to this entity. The real `255` entity could be something completely different, like a window with `surfaceparm glass`, and this will completely break the impact handling of the entity.

The correct solution (which I already mentioned at Discord) is to revert then change that made `eventParm` carry the entity number, because:

1. It's 8 bits, so it has max value of 255 while `MAX_GENTITIES` is `1024`, so you simply cannot make this work.
2. The change(s) completely break impact handling on old demos. Previously a bullet trace for the entity was done on cgame side, which took care of applying correct impact properties. This trace is no longer there, which means that when viewing old demos, impacts are completely borked because old demos do not contain the `eventParm` at all, so since the client only relies on this to determine the entity that a bullet hit, impacts are completely non-functional on old demos.

Vorschreibung:
Thanks for the explanation! Was not aware you mentioned the solution already on Discord, sorry!

Vorschreibung:
wrt. your scenario in the first paragraph, weirdly enough when I tested it, it seemed to work fine - the bullet reactions to `surfaceparm` apparently, at least client side, are not done via `eventParm` - but instead traced via `vec3_t end` (which I assume is passed elsewhere) AFAIS https://github.com/etlegacy/etlegacy/blob/deb750cd9f1cf90b24722bc08e4e229d7009473f/src/cgame/cg_weapons.c#L7083 ?

But anyway I'll walk back from here because I clearly don't get the full picture :3
Thanks for your reviews and the explanations!

Aciz:
The main issue lies in the event handling

https://github.com/etlegacy/etlegacy/blob/deb750cd9f1cf90b24722bc08e4e229d7009473f/src/cgame/cg_event.c#L2460-L2463

`eventParm` gets passed as the `fleshEntityNum` for the `CG_Bullet` call, which means that in case of wraparound, it could wrap to be a valid client number. This sets `flesh` to `qtrue`, which applies flesh impacts on any entity that wraps around to a value between `0-63`.

Vorschreibung:
Right, that's how far I got :3 - it seemed to me that `eventParm` is only used here as the `fleshEntityNum` local - which in turn is only used to determine if it's within `MAX_CLIENTS` :'D

AFAIS with my limited purview is that it's actually not used for any other things like `surfaceparm` calcs :/ - those seem to be retraced via `es->pos.trBase`.

But don't mind me :see_no_evil:

Aciz:
> AFAIS with my limited purview is that it's actually not used for any other things like `surfaceparm` calcs :/ - those seem to be retraced via `es->pos.trBase`.

Yeah, you're actually correct about that, my bad, the entity is correctly traced on client side if the entity is not a client. Still, my points mentioned earlier stand, using the `eventParm` to store the `entityNum` is just not a good idea.

Vorschreibung:
I completely agree :) it's pretty confusing and the way it seems now, `eventParm` for this case might as well be a BOOL I think :D.

If it's ok I'll give this another look tomorrow, maybe I can come up with something better :3.

PR #2581, opened 2024-05-27
cmake: Track etmain dependencies

Vorschreibung:
Static files packed into 'legacy_*.pk3' were not tracked as dependencies
by CMake.

This meant that e.g. changing 'etmain/ui/main.menu' would not cause
cause a rebuild of the 'legacy_*.pk3' if it has been built prior without
rerunning CMake or cleaning the build tree.

This was done by globbing twice, one time with 'GLOB_RECURSE' which is
used for 'DEPENDS', and once with 'GLOB' which is used for the arguments
to 'cmake -E tar'.

This is a workaround due to the fact that shelling out to 'cmake -E tar'
with the 'GLOB_RECURSE' LIST would provide more arguments than Windows
can handle and also due to 'cmake -E tar' not having implemented the
'-T'/'--files-from' option that GNU tar provides.

'file(ARCHIVE_CREATE ...)' would require a CMake version bump and
further does not seem to work with files outputted by targets.


Vorschreibung:
Should work now across all platforms, read to review again :heart_hands:

PR #2584, opened 2024-05-31
ui: fix Rect_ContainsPoint area check being 1px too small

Aciz:
This is especially noticeable in dropdown (combo) menus where there's a "dead pixel" in between every single entry due to the check effectively ignoring the very border of the elements.

### DIFF
--- a/src/ui/ui_shared.c
+++ b/src/ui/ui_shared.c
@@ -236,8 +236,8 @@ const char *String_Alloc(const char *p)
 	if (len + strPoolIndex + 1 < STRING_POOL_SIZE)
 	{
 		int ph = strPoolIndex;
-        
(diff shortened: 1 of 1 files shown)


Vorschreibung:
Nice catch!! Tried and works!

For reference fixes this:
image

I've also noticed tho that the same happens in the limbo menu:
image

Seems to use different code :thinking:

Aciz:
> Nice catch!! Tried and works!
> 
> For reference fixes this: image
> 
> I've also noticed tho that the same happens in the limbo menu: image
> 
> Seems to use different code 🤔

Good catch, it uses a different function indeed, fixed that too.

Vorschreibung:
Awesome! That Limbo menu has been fiddly for me as of recent, this will help for sure!
I've retested, LGTM :shipit:

PR #2586, opened 2024-05-31
Add misc/uncrustify.sh

Vorschreibung:
Bash script to check via 'uncrustify' if currently staged files pass it's checks.

Does not modify files but print diffs to be applied manually if need be.

Can be also passed a git rev to check the changed files of the rev.

### DIFF
--- a/misc/uncrustify.sh
+++ b/misc/uncrustify.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# uncrustify files in the worktree, either staged ones if no rev is passed, or
+# the ones changed in the passed rev
+#
+# REQUIRES: uncrustify
+set -eo pipefail
+
+# cd to parent dir of current script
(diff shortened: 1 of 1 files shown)


jackeri:
Isn't this already handled by easybuild.sh and easybuild.bat? They both have `crust` commands in them.

Vorschreibung:
> Isn't this already handled by easybuild.sh and easybuild.bat? They both have `crust` commands in them.

They crustify the whole worktree, which currently causes 120+ worktree changes, as a lot of uncrustified code is currently in there. - That's a bit of a footgun, especially for newcomers, I've shot my worktree like this once and then had to separate the unstaged changes I myself made with the uncrustify changes. - Crustifying the whole worktree seems more dangerous than useful and it doesn't seem to be regularly run.

I agree tho that this shouldn't exist next to the already existing `crust` commands in `easybuild`, I'll rethink this a bit and make a new suggestion or modify this one.

PR #2587, opened 2024-06-01
UI accelerators

Vorschreibung:
This change adds Accelerator key functionality to ET's Menus, demo: https://www.youtube.com/watch?v=j33ZnEXoO54 by parsing `^_` metacharacters out of UI texts, highlighting them and registering their action to the onKey handler.

If the general idea of Accelerators are desirable and the current implementation suffices, I'd follow up filling in more Accelerators for UI Elements until it's in a release-able state.

### DIFF
--- a/src/ui/ui_main.c
+++ b/src/ui/ui_main.c
@@ -232,6 +232,7 @@ void _UI_DrawTopBottom(float x, float y, float w, float h, float size)
 
 /**
  * @brief UI_DrawRect
+ * Draw an unfilled rectangle.
  * Coordinates are 640*480 virtual values
  * @param[in] x
  * @param[in] y
@@ -575,9 +576,10 @@ void Text_Paint_Ext(float x, float y, float scalex, float scaley, vec4_t color, 
 
 	if (text)
 	{
-		const char *s    = text;
-		int        len   = Q_UTF8_Strlen(text);
-		int        count = 0;
+		const char *s             = text;
+		int        len            = Q_UTF8_Strlen(text);
+		int        count          = 0;
+		qboolean   is_accelerator = qfalse;
 
 		trap_R_SetColor(color);
 		Com_Memcpy(&newColor[0], &color[0], sizeof(vec4_t));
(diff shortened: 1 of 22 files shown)


src/ui/ui_main.c
```diff
@@ -250,6 +251,26 @@ void _UI_DrawRect(float x, float y, float width, float height, float size, const
...
+void _UI_DrawBox(float x, float y, float width, float height, const float *color)
+{
+	trap_R_SetColor(color);
+
+	UI_AdjustFrom640(&x, &y, &width, &height);
+	trap_R_DrawStretchPic(x, y, width, height, 0, 0, 0, 0, uiInfo.uiDC.whiteShader);
+
+	trap_R_SetColor(NULL);
+}
+
```
Aciz: `UI_FillRect` (`DC->fillRect`) is line by line identical to this.

https://github.com/etlegacy/etlegacy/blob/ac4b2deeccd6cdbb9cea4e6ba87400dbf602ab87/src/ui/ui_atoms.c#L309-L317

Vorschreibung: Total oversight, thanks! Fixed

jackeri:
Really like this idea, though it does break the automatic generation of the translation files (which we haven't really used in a long time), but if that is all we can probably live with it.

src/ui/ui_menu.c
```diff
@@ -914,6 +914,49 @@ void  Menus_Activate(menuDef_t *menu)
...
+		const char *s   = item->text;
+
+		if (s == NULL) {
+			continue;
+		}
+
+		int       len                  = Q_UTF8_Strlen(s);
+		int       count                = 0;
+		short int accelerator          = 0;
+		qboolean  accelerator_upcoming = qfalse;
```
Aranud: This doesn't respect C89 standard. We try to stick as much as possible to this declaration style for concistency with code. Variables should be declared at top of the scope then `len` assigned after `s` NULL check.

Aranud (changes requested):

src/ui/ui_menu.c
Vorschreibung: :metal: Done

Aranud (approved):

Vorschreibung:
The scope of this has never been sufficiently clarified while meta-chars to translation files were explicitly rejected. - I think it makes more sense to close this altogether as that would constitute a different approach altogether that someone can take up in the future and where it's unlikely these commits would be of any help.

PR #2590, opened 2024-06-05
game: Skip warmup on devmap

Vorschreibung:
Assists in smoother/faster devmap cycles.

As is `g_warmup` is capped to at least 10 seconds.


ryzyk-krzysiek:
If you would like to start directly on gamestate `GS_PLAYING` you should be able to do so with config.

`g_customConfig test`
File `test.config`
```
configname "^1This is test config"

init
{
	setl gamestate "0"
}
```

PR #2591, opened 2024-06-05
game: Enable instant respawn on devmap

Vorschreibung:
Assists in smoother/faster devmap cycles.

### DIFF
--- a/src/game/g_active.c
+++ b/src/game/g_active.c
@@ -1737,8 +1737,13 @@ void SpectatorClientEndFrame(gentity_t *ent)
 		gclient_t *cl;
 		qboolean  do_respawn = qfalse;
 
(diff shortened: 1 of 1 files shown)


ryzyk-krzysiek:
You could use configs instead for various mod/server or even client settings. For fast respawns:

`g_customConfig test` (can set it on startup with `+set`)
File `test.config`
```
configname "^1This is test config"

init
{
	setl g_useraxisrespawntime "1"
	setl g_useralliedrespawntime "1"
}
```

Hardcoding instant respawns doesn't sound right when it already happens when warmup is on and is customizable outside of warmup.

EDIT: changed the cvars because the limbo cvars might not work on some maps (overwritten by mapscripts)

PR #2593, opened 2024-06-07
cmake: Switch build from C99 to C90 Standard

Vorschreibung:


Vorschreibung:
@jackeri via Discord:
> The engine is allowed to be c99

PR #2594, opened 2024-06-07
cmake: Add some C warnings

Vorschreibung:
Adding:

	-Wdeclaration-after-statement
		this warns about the C89 style where declarations are not
		allowed to come after statements, but must always be at the
		start of a body

	-Wunused-but-set-variable
		warns about unused variables

Adding these CFLAGS as of now produces 82 warnings, all of which are -Wdeclaration-after-statement


Vorschreibung:
If this is of interest/desirable, I'd follow this up with a commit to fix the 82 warnings.

PR #2596, opened 2024-06-09
qcommon: popup update menu only if running legacy mod

Aciz:
This menu does not exist in other mods obviously, so resort to console print only.


Aciz:
Actually I'm wrong here, the menu does exist, but legacy has re-ordered the `uiMenuCommand_t` enum so that should be fixed to point to correct numbers instead.

https://github.com/etlegacy/etlegacy/blob/a49d34885008c99b40d12c382700548626201fff/src/game/bg_public.h#L2804-L2837

PR #2598, opened 2024-06-17
Fix all -Wdeclaration-after-statement warnings

Vorschreibung:
As promised a followup to 247d7fd270a9ba8f1d4f8b4fff8b70c70e5d16c0 to appease the 82 introduced mixing declarations and statements warnings.

Changes validated by `crustify`.

### DIFF
--- a/src/cgame/cg_draw.c
+++ b/src/cgame/cg_draw.c
@@ -4516,6 +4516,7 @@ void CG_Coronas(void)
 		trace_t  tr;
 		float    dist;
 		vec3_t   dir;
+		float    fov;
 		qboolean visible, behind, toofar;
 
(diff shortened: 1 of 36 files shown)


src/cgame/cg_draw.c
```diff
@@ -4536,7 +4537,7 @@ void CG_Coronas(void)
 				toofar = qtrue;
 			}
 			// dot = DotProduct(dir, cg.refdef_current->viewaxis[0]);
-			const float fov = cosf((cg.refdef_current->fov_x / 2) * (float)(M_PI / 180));
```
Vorschreibung, line 4539: Head's up, opted for not introducing another scope here and removed `const` here instead.

PR #2599, opened 2024-06-17
Improve german translation

Vorschreibung:
Various canonicalizations and filling in missing translations.


jackeri:
These need to be done in transifex, not in the repo. https://app.transifex.com/etlegacy/etlegacy/translate/

PR #2613, opened 2024-07-03
cgame: Only render rifle grenade when NOT firing

### LINKED ISSUE #2602: Rnade visual bug. Rnade flying towards player in first frames

Vorschreibung:
Fixes https://github.com/etlegacy/etlegacy/issues/2602


Vorschreibung:
NOTE: `WEAPON_DROPPING` seems like a mistake, but it's actually the correct weaponstate, which also aligns with the `ATTACK2` anim of both `kar98` and `m1_garand`

PR #2616, opened 2024-07-04
renderer: fix SonarCloud array bound issue

balazs-benics-sonarsource:
This SonarCloud report was marked as false-positive, while I believe this is a true-positive issue.
We noticed your feedback and we decided to take a look to improve.

https://sonarcloud.io/project/issues?id=etlegacy_etlegacy2&issues=AY9Y8kEHq42NS6QGFOc-&open=AY9Y8kEHq42NS6QGFOc-&tab=code

I agree that the report steps could be improved on our end though.

---

Issue: If `client` holds the value 32, the control flow will avoid the `Ren_Drop` and lead to accessing out-of-bounds elements of the `tr.scratchImage` array. This is an off-by-one bug.

### DIFF
--- a/src/renderer/tr_backend.c
+++ b/src/renderer/tr_backend.c
@@ -900,7 +900,7 @@ void RE_UploadCinematic(int w, int h, int cols, int rows, const byte *data, int 
 	int     start;
 	image_t *image;
 
-	if (client < 0 || client > (sizeof(tr.scratchImage) / sizeof(tr.scratchImage[0])))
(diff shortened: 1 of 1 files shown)


Aranud:
Thanks you very much for your feedback and pin pointing the exact issue.

In fact, it's a whole human issue (mine to be more precise). I totally miss read the report and was looking for a lower limit out of access bound ... which was already correct. The report is just fine and already explain finely the real issue.

Sorry for this false-positive report !

balazs-benics-sonarsource:
Awesome! Let us know if you have feedback to share. Enjoy Sonar solutions.

PR #2620, opened 2024-07-04
game: Smooth PM_GroundTrace (OTG) Jump Animation

Vorschreibung:
The kickoff/OffTheGround (OTG) move mechanic has been very oversensitive
when it comes to jump animations, insofar that moving across fairly flat
surfaces with only slight angle changes will cause a kickoff event and
consequently a jump animation event, which causes players to "stumble" /
"hobble" across seemingly flat terrain.

In order to remedy the worst cases, we introduce a directional check in
the kickoff event that gates the jump animation playback.

In order for a kickoff to trigger a jump animation, it now has to be
facing _somewhat_ towards the sky or anyhwere downwards.

The values have been tuned to still cause a jump animation when getting
a kickoff via prominent terrain features.

The gameplay mechanical code of the kickoff events remains completely
unchanged, as such it is a purely cosmetic.

Issue #944, opened 2020-03-05
Fix ladder movement


ETLegacyTeam:
Author: @rmarquis (Spyhawk)
Date: 2016-09-20
Redmine Issue: 944, https://dev.etlegacy.com/issues/944

---

The animation at the top of a ladder is buggy. The character climbs in the void until the feet arrive at the top of the ladder, after which the ANIM_ET_CLIMB_DISMOUNT is played.

This is visible in third person view only. 

---

Related issue(s): #406

ETLegacyTeam:
From: @realkemon (keMoN)
Date: 2018-08-11T10:11:30Z

---

Noticed today again and decided to record a video.

PR #2623, opened 2024-07-06
etmain: Fix climbing animations

Vorschreibung:
Climbing animations went through a regression between OG 2.60b ET and
current ETL, this commit addresses these regressions and cleans climbing
animations up along it.

- 'climb_dis' and 'climb_start' were way too slow and lengthy - causing
  the bugged-looking animations when leaving or entering a ladder

- 'climb_start' was effectly ditched, as it's rarely played and causes
  the player model to clip into the terrain on most occassions

- note that 'climb_dis' is actually played always when leaving a ladder,
  this means it's played both when climbing up and leaving the ladder,
  but also just the same when climbing down and leaving the ladder

- sometimes when grabbing a ladder to climb up, a jump animation is
  played, this is due to a kickoff/OTG event which plays the jump
  animation and is dependent on the surface one stands on when grabbing
  the ladder - this is addressed by gating the animation behind an
  additional check for being on a ladder

- when grabbing al adder to climb down, a jump animation is also played
  via 'PM_GroundTraceMissed'/freefall, as when falling onto a ladder
  from above, for a few frames there is no contact yet with the ladder -
  to remedy this, the 'ANIM_ET_CLIMB_MOUNT' animation, which triggers
  shortly after is set to 'force' to overrule the few animation frames
  of the previous jump animation

- 'climb_up' was way too fast, climbing looked comical, made slower

- 'climb_down' was too fast, made slower

### DIFF
--- a/etmain/animations/human/base/body.aninc
+++ b/etmain/animations/human/base/body.aninc
@@ -118,10 +118,10 @@
 
 		//drop				2975	5	0	20	0	0	0
 
-		climb_up			2981	21	21	20	-1	0	0
-		climb_dis			3003	31	0	20	0	0	0
-		climb_start			3035	46	0	20	0	0	0
-		climb_down			3082	28	28	20	-1	0	0
+		climb_up			2981	21	21	10	-1	0	0
+		climb_dis			3004	7	0	40	0	0	0
+		climb_start			2981	1	0	100	-1	1	0  // effectively disable
+		climb_down			3082	28	28	12	-1	0	0
 
 		jump_1step_2h		3111	11	0	20	0	0	0
 		jump_1step_1h		3131	11	0	20	0	0	0
--- a/src/game/bg_pmove.c
+++ b/src/game/bg_pmove.c
@@ -1840,8 +1840,10 @@ static void PM_GroundTraceMissed(void)
  */
 static void PM_GroundTrace(void)
 {
-	vec3_t  point;
-	trace_t trace;
+ 	vec3_t       point;
+ 	trace_t      trace;
+ 	const vec3_t upwards = { 0.0f, 0.0f, 1.0f };
+ 	float        upwardsdirection;
 
 	point[0] = pm->ps->origin[0];
 	point[1] = pm->ps->origin[1];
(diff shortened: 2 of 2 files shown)


Vorschreibung:
OG ET 2.60b: https://www.youtube.com/watch?v=qhLELEreCMU
pre-commit: https://www.youtube.com/watch?v=2h5D83LXG_g
post-commit: https://www.youtube.com/watch?v=UTAz_GEgqNo

Vorschreibung:
A bunch of comments on this:

1. `climb_dis` and `climb_start` were **way** too slow and lengthy - causing the bugged-looking animations when leaving or entering a ladder
2. sometimes when entering/grabbing a ladder, a jump animation is played, this is seemingly due to a kickoff/OTG event which plays the jump animation and dependent on the surface one stands on when grabbing the ladder - and which is not addressed in this commit, but can be addressed separately if need be
3. `climb_up` was too fast on ETL in comparison to 2.60b, made slower
4. `climb_down` was too fast on ETL in comparison to 2.60b, made slower

Vorschreibung:
**IMPORTANT** Should only be merged _AFTER_ merging: https://github.com/etlegacy/etlegacy/pull/2620

Aranud:
Potential fix for https://github.com/etlegacy/etlegacy/issues/944

ryzyk-krzysiek:
I quickly checked because it's something I looked into in the past but never tried `ETMain` only `ETPro`.

It looks like the regression was introduced in `ETPro` which was merged into `Legacy` here https://github.com/etlegacy/etlegacy/commit/8dcc6eb4ea03c4afba94ce03098adcb6f6e76a96 and the difference in climb animations is change from 40fps to 20fps. (I was unaware of this commit untill now and it could be wise to check what else could have changed since there could be more broken stuff) 

ETMain
```
climb_up		2981	21	21	40	-1	0	0
climb_dis		3003	31	0	40	0	0	0
climb_start		3035	46	0	40	0	0	0
climb_down		3082	28	28	40	-1	0	0
```

Legacy and ETPro
```
climb_up		2981	21	21	20	-1	0	0
climb_dis		3003	31	0	20	0	0	0
climb_start		3035	46	0	20	0	0	0
climb_down		3082	28	28	20	-1	0	0
```

Vorschreibung:
I've diffed them and looked through them, LGTM - grenade firing is the only one which had it's fps changed (this time doubled instead of halfed), but looks fine to me ingame, there's a lot of quirks in the released OG animations.

PR #2630, opened 2024-07-10
Allow ETL_HOME env-var to set homedir

Vorschreibung:
Assists in running ETL multiple times locally with different configs, e.g. for debugging purposes.


Vorschreibung:
Can also be done via `+set fs_homepath`

PR #2634, opened 2024-07-13
game: Allow non-negative g_warmup values

Vorschreibung:
This allows disabling warmup altogether via 'g_warmup 0'


Vorschreibung:
I went through git blame history to see if there was any reason why it was capped to at least 10 seconds - found nothing.

Issue #724, opened 2020-03-05
In-game mouse cursor shows when opening certain popup menus


ETLegacyTeam:
Author: @saukko28 (Saukko)
Date: 2014-12-14
Redmine Issue: 724, https://dev.etlegacy.com/issues/724

---

See the screenshot attached, you see a tiny bit of the cursor. When you check by pressing ESC, the cursor starts from the very same place.

-*S

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2019-06-27T09:48:47Z

---

Alright I gathered more info on this one. The only popup menu that actually works as intended is fireteam voicecommand menu. I haven't found yet what could make it different from the others.

-*S

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-06-27T10:05:42Z

---

> I haven’t found yet what could make it different from the others.

The fireteam command menu is hardcoded (in cgame), while the other menus use the script system (.menu files).

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2019-06-27T10:19:51Z

---

Actually now I came to notice that there is this event handler in cg_newDraw. So yeah, you are absolutely correct. So at this point the only question is that to which direction this case should be heading? Should eveyrhting be hardcoded or rather not?

-*S

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2019-06-27T10:36:49Z

---

Hardcoding provides greater flexibility (which the fireteam menu make use of), but doing so is indeed lot of work. Using menu file is much simpler and faster, but we can't exactly do everything we want. Is it worth hardcoding all the menu to fix that little cursor issue? Absolutely not.

ETLegacyTeam:
From: @saukko28 (Saukko)
Date: 2019-06-27T11:36:41Z

---

That is a valid point. This is such a minor issue so it might not be worth it. But still in the end of the day, a fixable bug is always fixable. ;)

But we can let this be as it is... for now.

-*S

Vorschreibung:
Redmine ticket is gone, what popup menus are affected here? - Can't make sense of this issue / reproduce.

PR #2635, opened 2024-07-14
mod: properly hide mouse cursor if menu bypasses mouse input

Aciz:
Instead of doing silly hacks that move the mouse cursor off-screen when opening a menu which sets `cl_bypassmouseinput 1`, just don't draw it in the first place. This also fixes the cursor position resetting when opening a menu which bypasses mouse input, retaining the previous position when a mouse-controlled menu was opened the last time.

The cursor drawing is a somewhat complex system so I'd appreciate if some testing was done on this, to make sure the cursor is visible/hidden in the correct places. I've tested the following menus:

* main/in-game menu
* vsay menu
* ft menu
* class/teams/spawnpoint menu
* chat
* limbo
* speaker editor
* HUD editor
* demo info window (pause/seek/rewind controls)
* intermission
* shoutcaster mode (since it has it's own event handling, cursor isn't drawn)

refs #724

### DIFF
--- a/src/cgame/cg_camera.c
+++ b/src/cgame/cg_camera.c
@@ -293,7 +293,7 @@ void CG_CameraEditorDraw(void)
 
 		// render cursor
 		trap_R_SetColor(NULL);
-		CG_DrawPic(cgDC.cursorx, cgDC.cursory, 32, 32, cgs.media.cursorIcon);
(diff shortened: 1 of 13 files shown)


Aciz:
Intermission handling was broken, but this is fixed now. Also tested shoutcaster mode briefly, seems to work fine.

PR #2636, opened 2024-07-14
cgame: add support for devmap-specific map autoexecs

Aciz:
When cheats are enabled, try to execute `autoexec_devmap_mapname` or `autoexec_devmap_default` first, and if not found, fallback to regular map autoexecs.

This also refactors the config loading a bit, there's now a separate function to check if a file exists, so the return value of `CG_execFile` doesn't need to be ignored.

### DIFF
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -821,25 +821,54 @@ void CG_RestoreProfile(void)
 }
 
 /**
- * @brief Try to exec a cfg file if it is found
+ * @brief Checks whether a cfg file exists or not
  * @param[in] filename
- * @return
+ * @return qtrue if file exists
  */
-qboolean CG_execFile(const char *filename)
+qboolean CG_ConfigFileExists(const char *filename)
 {
 	int handle = trap_PC_LoadSource(va("%s.cfg", filename));
-
 	trap_PC_FreeSource(handle);
+	return handle ? qtrue : qfalse;
+}
 
-	if (!handle)
-	{
-		// file not found
-		return qfalse;
-	}
-
+/**
+ * @brief Execs a .cfg file
+ * @param[in] filename
+ */
+void CG_execFile(const char *filename)
+{
 	trap_SendConsoleCommand(va("exec %s.cfg\n", filename));
+}
 
-	return qtrue;
+/**
+ * @brief Executes map-specific autoexec files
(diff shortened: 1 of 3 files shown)


src/cgame/cg_main.c
```diff
@@ -821,25 +821,54 @@ void CG_RestoreProfile(void)
...
+void CG_execFile(const char *filename)
+{
+	trap_SendConsoleCommand(va("exec %s.cfg\n", filename));
+}
+
+/**
+ * @brief Executes map-specific autoexec files
+ * @param[in] cheats
+ */
+void CG_MapAutoexec(qboolean cheats)
```
Vorschreibung, line 848: nit: consider renaming `cheats` arg to `useDefaults` or similar - 2nd callsite passes `qfalse` despite `cgs.sv_cheats == 1`

Vorschreibung:
Tested, works great! :partying_face: :heart_hands:

ryzyk-krzysiek:
I wonder what is the use case for this. Why would you need specific config for `devmap` when cheats are basically few commands that when not used the game is working the same way as if cheats are off?

Aciz:
> I wonder what is the use case for this. Why would you need specific config for `devmap` when cheats are basically few commands that when not used the game is working the same way as if cheats are off?

I had the same initial thought when this came up in Discord, but it can actually be quite useful. For example, you can have a completely different set of keybindings in devmap for practicing stuff, without messing up your "real" keybindings. Sure this can be achieved by just writing different configs and execing them, but this can just be used to automate things.

Vorschreibung:
Disclaimer: I've brought this up initially in Discord, so probably no surprise that I'm expressing support for this.

My initial use-case was to set defaults that are very useful for `devmap`ing without having to hard code them, while also being able to comfortably tweak/disable them.

Specifically: 

1. Disabling warmup during devmap once https://github.com/etlegacy/etlegacy/pull/2634 is submitted - as warmup serves no purpose during devmap and resets the playerposition

2. Disable respawntimes during devmap, which is something I have locally as a patch and which might be already doable via server vars or if not, could be considered and added later, or not if deemed too intrusive.

Aside from them, the mechanism seems useful to me both as a comfortable way for custom settings (e.g. custom keybindings that Aciz mentions) - and also as a canonical and simple way to communicate devmap settings between map makers in a consistent way.

PR #2641, opened 2024-07-15
Cleanup hweapon comments / unused persEnum_t entries

Vorschreibung:
Specify that some comments on `mg42` are about stationary misc_mg42, not about the wieldable mg42 weapon to increase grep-ability.

Also clean up some unused `persEnum_t` entries.

### DIFF
--- a/src/game/bg_public.h
+++ b/src/game/bg_public.h
@@ -553,7 +553,7 @@ typedef struct pmoveExt_s
 	int silencedSideArm;           ///< Keep track of whether the luger/colt is silenced "in holster", prolly want to do this for the kar98 etc too
 	float sprintTime;
 
-	// MG42 aiming
+	// stationary heavy weapon (e.g. misc_mg42, misc_aagun) aiming
 	float varc, harc;
 	vec3_t centerangles;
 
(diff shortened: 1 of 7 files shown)


src/game/bg_public.h
```diff
@@ -704,13 +704,8 @@ typedef enum
...
 
 	PERS_REVIVE_COUNT,
-	PERS_HEADSHOTS,                ///< Deprecated. Remove?
-	PERS_BLEH_3,
 
-	// mg42                        ///< TODO: I don't understand these here. Can someone explain?
-	PERS_HWEAPON_USE,              ///< enum 12 - don't change
-	// wolfkick
-	PERS_WOLFKICK
+	PERS_HWEAPON_USE = 12,         ///< non-zero when using a stationary hweapon (mg42=1, aagun=2)
```
Aciz: Just FYI this whole enum can hold max 16 values, and it's shared with the engine as it's part of playerstate, it's extremely important to keep track of this and not change the internal numbering as that causes incompatibility with old demos. I would in fact prefer to explicitly define all these 16 values, even the ones that are spare and clearly mark them as unused.

Vorschreibung: @Aciz please check out new revision

Aciz (approved):

Vorschreibung:
Found 1 more unused entry (PERS_HITS), renamed in last revision.

PR #2642, opened 2024-07-15
game: fix sniper scope staying up if revived and holding +attack

### LINKED ISSUE #2637: Cvops scope is up after revive
When Cvops gets killed while scope is up. After revive Cvops has scope up. (Should not be as damage unscopes, therefore getting killed should unscope)

Aciz:
Normally while player spawns, holding `+attack` ceases all weapon functionality until the button is let go. This means that if a player dies while scoped in with a sniper and holds down `+attack` as they get revived, the `PM_Weapon` function does not run and therefore doesn't update the weapon in playerstate, resulting in the player keeping the scope.

fixes #2637


Aciz:
Closing in favor of #2644 which also handles this

PR #2649, opened 2024-07-16
cgame: Add HUD "No Countdown" style to 'followtext'

Vorschreibung:
Add a way to keep the followtext without the reinforce countdown, which can also be inferred from the roundtimer.


Vorschreibung:
Comparison
shot_1721163882
shot_1721163881

PR #2650, opened 2024-07-16
cgame: Add hotkeys for navigating limbo

Vorschreibung:
Add hotkeys to navigate limbo menu.


Vorschreibung:
Demo: https://www.youtube.com/watch?v=YuK2pTWxvR4

Vorschreibung:
Would prefer grid hotkeys for the class selection, i.e. instead of `smefc` just left-to-right on whatever comes below your `1` key, in my case `qwert` - but AFAIS there's currently nothing for this implemented.

PR #2652, opened 2024-07-17
game: Fixup giving ammo

Vorschreibung:
- 'give all'        now only gives up to 'maxammo' for each weapon
- 'give ammo'       now only gives up to 'maxammo'
- 'give ammo all'   now forces 9999 ammo
- 'give ammo -9999' now caps the ammo at 0 instead of becoming negative

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -650,6 +650,7 @@ void Cmd_Give_f(gentity_t *ent, unsigned int dwCommand, int value)
 	int      j;
 	int      amount       = 0;
 	qboolean hasAmount    = qfalse;
+	qboolean isAllAmount  = qfalse;
 	int      i            = 1;
 	qboolean validGiveCmd = qfalse;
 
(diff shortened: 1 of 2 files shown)


src/game/g_cmds.c
```diff
@@ -695,8 +696,16 @@ void Cmd_Give_f(gentity_t *ent, unsigned int dwCommand, int value)
 	trap_Argv(++i, amt, sizeof(amt));
 	if (*amt != '\0')
 	{
-		hasAmount = qtrue;
-		amount    = Q_atoi(amt);
+		if (Q_stricmpn(amt, "all", 3) == 0)
+		{
+			hasAmount   = qtrue;
```
Aranud, line 701: This allow using "all" as value for skill and health amount. Which is not wanted.

Aranud (changes requested):

Vorschreibung:
Superseded by https://github.com/etlegacy/etlegacy/pull/2652

src/game/g_cmds.c
Vorschreibung: Not an issue in new PR

PR #2653, opened 2024-07-17
Add some missing header includes & Fix circular 'q_shared.h' include

Vorschreibung:
```
Add some missing header includes
```

&

```
qcommon: Fix circular 'q_shared.h' include

Fix the following circular/recursive include:

           src/qcommon/q_shared.h
includes → src/qcommon/q_math.h
includes → src/qcommon/q_shared.h

By factoring out the required parts used by 'q_math.h' from 'q_shared.h'
into 'q_primitives.h'.
```


Vorschreibung:
Appeases `clangd` greatly / helps with IDE "intellisense".

PR #2667, opened 2024-07-19
etmain: Reframe 1P flamethrower animations

Vorschreibung:
Comparison: https://www.youtube.com/watch?v=KlyigmoebPQ

---

Add an ATTACK1 anim that slightly pulls back the flamethrower while shooting.

Add a slower ATTACK_LASTSHOT anim that plays when out of ammo.

Adjust RAISE and DROP animations to be more smooth and less abrupt.

Fix "hover"-flame, which occurs on cg_fov values other than 90, by adding a 'fov_y' dependent offset - as a result the idle flame should now always stick to the muzzle.

### DIFF
--- a/src/cgame/cg_weapons.c
+++ b/src/cgame/cg_weapons.c
@@ -3601,6 +3601,33 @@ void CG_AddPlayerWeapon(refEntity_t *parent, playerState_t *ps, centity_t *cent)
 		CG_PositionRotatedEntityOnTag(&flash, &gun, "tag_flash");
 	}
 
+	// @XXX hardcode offset flash positions, to correct wrong official
+	// muzzle locations in some cases
+	{
+		vec3_t  forward;
+		float flash_offset = 0.0f;
+
+		switch (weaponNum) {
+			case WP_FLAMETHROWER:
+				// adjust flamethrower muzzle according to cg_fov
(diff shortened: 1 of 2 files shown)


src/cgame/cg_weapons.c
```diff
@@ -3601,6 +3601,33 @@ void CG_AddPlayerWeapon(refEntity_t *parent, playerState_t *ps, centity_t *cent)
...
+				break;
+		}
+
+      if (flash_offset != 0.0f) {
+         AxisToAngles(flash.axis, angles);
+         AngleVectors(angles, forward, NULL, NULL);
+         VectorMA(flash.origin, flash_offset, forward, flash.origin);
+      }
+	}
+
```
Aranud, line 3630: Don't do a switch. Flamethrower is certainly the only weapon where the tag_flash is miss-placed compared to all other weapons.

Aranud (changes requested):

src/cgame/cg_weapons.c
Vorschreibung: > Don't do a switch. Flamethrower is certainly the only weapon where the tag_flash is miss-placed compared to all other weapons.

I have 4 more upcoming in my stash for engi+covi rifle, but I agree that these should just be modified when I add them.

Aranud: Fine, let keep stuff like this

Aranud (approved):

PR #2670, opened 2024-07-21
cgame: Make Smoke Bomb puff direction deterministic

### LINKED ISSUE #2622: Smoke Bomb smoke is randomly rendered different for each client
Have not tested but this seems the case coming from few players. If this is true it creates inconsistency where you can see your enemy but your enemy can not see you etc. Therefore it would an issue as u would expect the smoke to be seen the same for every client looking at it.

Vorschreibung:
Instead of chosing random directions upon the creation of each Smoke sprite/puff, rely on a hashed value, derived from the server timestamp attached to the smoke bomb entity + it's current smoke puff number.

Shamelessly repurpose the unused 'teleportFlag' field of 'centity_t', to track the smoke puff number.

Fixes https://github.com/etlegacy/etlegacy/issues/2622

### DIFF
--- a/src/cgame/cg_effects.c
+++ b/src/cgame/cg_effects.c
@@ -133,7 +133,7 @@ localEntity_t *CG_SmokePuff(const vec3_t p, const vec3_t vel,
 	le->radius  = radius;
 
 	re             = &le->refEntity;
-	re->rotation   = Q_random(&seed) * 360;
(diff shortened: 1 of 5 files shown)


Vorschreibung:
Pushed new rev, reused the already existing LCG and cleaned up the naming a tad.

PR #2671, opened 2024-07-21
server: Reduce sv_{,dl_}timeout

Vorschreibung:
sv_timeout    from 60 to 10
sv_dl_timeout from 300 to 60

If a client crashes or shoots the process down, these take effect.


Vorschreibung:
10 was said to be too low, on second thought on public servers with big playercounts, being kicked for discs too early could mean losing your slot and if someone else joins in the meantime you would have to wait in queue - so kicking players too quickly is also undesired

`sv_dl_timeout` might also actually turn out to be the total amount of time allowed for a client to download a file and reducing it to 1min would also be too little

let's keep the current ones

PR #2672, opened 2024-07-21
cgame: Announce 'prepare' on warmup start

Vorschreibung:
...instead of when 10s before match start.


Vorschreibung:
New rev, shaved off a local.

Vorschreibung:
New PR, now only applies when `g_doWarmup 1`  otherwise fallback to 10s announce as previously

Issue #1389, opened 2020-03-05
Windows 64bit Support for ETL


ETLegacyTeam:
Author: @rafal1137 (RaFaL)
Date: 2020-01-10
Redmine Issue: 1389, https://dev.etlegacy.com/issues/1389

---

Let add 64bit support for Windows and there is a reason and prove for make it happen.

Whenever i do devmap some maps with our HD textures installed like goldrush, dinkershing by kemon I am starting to get errors due too some limitations.
Not sure if everyone experienced that one.

PS: And of course R2 is used.

---

Related commit(s):
818318be697f2b395d6727bb1d327a9073e586f8 - client: dirty fix for buflen bufferoverflow, refs ..
 5cad58f1d0a8d867ed4ed1afd182950ab3d0f883 - renderer2: fix compiler warning, refs 1389

u1ps:
Windows 64bit Support for ETL is here :D ?

Issue #157, opened 2020-03-05
Provide 64-bit binaries


ETLegacyTeam:
Author: @adanski (Dragonji)
Date: 2013-02-07
Redmine Issue: 157, https://dev.etlegacy.com/issues/157

---

So, what about x64 binaries?

I know there are no mods that support ET x64 but it would be nice to have x64 binaries so we can play and test around with legacy mod.

---

Related issue(s): #364
Child issue(s): #195

ETLegacyTeam:
From: @morsik (morsik)
Date: 2012-12-25T19:38:41Z

---

I hope we'll release x86_64 version with 2.70. It's working without any problem and it's easy to compile (just disable cross building - set variable in CMake)

Linux x86_64 - sure yes
Windows x86_64 - don't know. Windows 64bit is not that problem like linux, where it always says "Can't find library". Mixing bits in Windows is common, while in linux aren't. We can think about it of course, but it's not our priority.
Mac OS X x86_64 - i hope yes, but Radegast must say that while only he has OS X (-;

ETLegacyTeam:
From: @TheDushan (TheDushan)
Date: 2012-12-29T21:30:29Z

---

I will provide Win64 support and binary files for ET: Legacy. Because of that I have assigned this ticket to myself.

ETLegacyTeam:
From: @rmarquis (Spyhawk)
Date: 2013-04-12T13:28:31Z

---

ALthough I can compile in full 64bits using the system Lua librairy (5.2), it seems I'm not able to compile in 64bits with the embedded Lua 5.1 lib:

Linking C shared module legacy/qagame.mp.x86_64.so
/usr/bin/ld: libs/lua/src/liblua.a(lapi.o): relocation R_X86_64_32 against `luaO_nilobject_' can not be used when making a shared object; recompile with -fPIC
libs/lua/src/liblua.a: could not read symbols: Bad value
collect2: error: ld returned 1 exit status
make[2]: *** [legacy/qagame.mp.x86_64.so] Error 1
make[1]: *** [CMakeFiles/qagame.mp.x86_64.dir/all] Error 2
make: *** [all] Error 2

ETLegacyTeam:
From: @adanski (Dragonji)
Date: 2013-11-03T12:39:13Z

---

Just a thought... Are 64-bit bins really needed? ET will never break 4GB RAM usage limitation ;)

[...]

PR #2674, opened 2024-07-22
qcommon: fix 64-bit VM_Call/SystemCall argument expansion

Aciz:
Macro the shit out of `VM_Call` and `SystemCall` to correctly cast all arguments to `intptr_t` so we don't do undefined behavior on 64-bit platforms when passing 4-byte integers etc. to `VM_Call` and `VM_DllSyscall`.

Also fixes 64-bit Windows using incorrect path in `VM_DllSyscall` as `__x86_64__` isn't defined on Windows.

refs #1389, #157


Aciz:
`SystemCall` is now fixed too, along with `VM_DllSyscall` `#if/else` condition, which wasn't correctly setup for 64-bit Windows.

PR #2677, opened 2024-07-22
game: Tweak 'give' command

Vorschreibung:
- added `give skill none` to reset skills
- added `give hp` as an alias for `give health`
- added `give health none` to damage a player`s health to 0
- added `give weapon` as an alias for `give weapons`
- added `give weapon <number>` to give only a specific weapon instead of all
---
- changed `give all` to give up to 'maxammo' for each weapon, previously gave forced 9999 ammo for each weapon
---
- changed `give ammo` now only giving up to 'maxammo' for each weapon, previously gave forced 9999 ammo for each weapon
- changed `give ammo -9999` now caps the ammo at 0 instead of setting player ammo to negative values
- added   `give ammo all` to force 9999 ammo for each weapon
- added   `give ammo clip` to give a clip worth of ammo to each weapon
- added   `give ammo none` to remove all ammo and empty all clips from each weapon
---
- made plain `give` to show most available combinations

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -649,6 +649,7 @@ void Cmd_Give_f(gentity_t *ent, unsigned int dwCommand, int value)
 	int      cnum;
 	int      j;
 	int      amount       = 0;
+	qboolean isNoneAmount = qfalse;
 	int      i            = 1;
 	qboolean validGiveCmd = qfalse;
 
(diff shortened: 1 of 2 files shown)


Vorschreibung:
Supersedes https://github.com/etlegacy/etlegacy/pull/2652

Issue #2679, opened 2024-07-22
Stats update every 5 sec can cause a drop in snaps in bigger server?


u1ps:
When there are close to 50 players online. Every 5 secs there will be spike in lagometer. Depending on activity in server these spikes can cause a delay / microstutter. (can turn into yellow line in lagometer)

![image](https://github.com/user-attachments/assets/ed2a0b9e-a0e3-4d6e-bd04-384138c3eba0)

u1ps:
https://github.com/user-attachments/assets/503254bc-462f-4f70-95fe-b06337d80cea

One more example of this causing yellow spike every 5 sec

PR #2681, opened 2024-07-22
mod: change the way /scores server command works, refs #229

ryzyk-krzysiek:
- add back /scores and auto updates to scores in tvgame
- does not break old demos

Note:

/scores command causes lag client side, because it is really big it fragments the packet (mtu is 1300) if sent all at once (just 10 players can cause the packet to be split even 3 times). This causes the player to get rate limited which in turn means delayed packets which means players has nothing to interpolate to.

Instead of sending everything at once the strings are saved per player basis in qagame and added to snapshot one per clientthink, this still can be too much on some settings (com_maxfps and sv_fps) and with many players on server. Command is displayed on client only after receving all strings. The downside to this is there will be more delay when the scores commands appears on client.

Pretty much the only reason I would like to add it is to be able to do automatic updates of scores for etltv (only legacy mod) without causing server lag every X seconds (etltv server is also a client that gets it's packets fragmented).

refs #229

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2712,6 +2712,8 @@ typedef struct cgs_s
 	int sv_fps;                 // FPS server wants to send
 	sampledStat_t sampledStat;  // fps client sample data
 
+	int scoresCount;
+	char scores[MAX_SCORES_CMDS][MAX_STRING_CHARS];
 } cgs_t;
 
 //==============================================================================
--- a/src/cgame/cg_servercmds.c
+++ b/src/cgame/cg_servercmds.c
@@ -576,9 +576,12 @@ void CG_ParseWolfinfo(void)
 			// only _after_ calling this place here (which would make
 			// 'trap_S_StartLocalSound' work correctly) - we add this edge case
 			// to play the 'FIGHT' announcer voice line on the spectator itself.
-			if (cg.snap->ps.pm_flags & PMF_FOLLOW) {
+			if (cg.snap->ps.pm_flags & PMF_FOLLOW)
+			{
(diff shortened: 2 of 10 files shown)


Aciz:
This probably fixes/at least improves the situation with https://github.com/etlegacy/etlegacy/issues/2679?

Edit: ah no, this is the manual `/scores` command, right? And not the automated one that keeps scoreboard up-to-date?

ryzyk-krzysiek:
> This probably fixes/at least improves the situation with #2679?

Not really, `/scores` is a massive lag on a server like TM, in fact if you spam it enough it can stall your game to the point you get disconnected IIRC, but because you always use it with console open and then spend time reading it and not playing, you pretty much never notice it.

If it is really every 5 seconds then it is most likely scoreboard update, which is updated every 5s through here:
https://github.com/etlegacy/etlegacy/blob/1de8d85411ae930c31b2a50421276fa13e30e7d0/src/game/g_active.c#L1703-L1704

But it is at most 2 strings so the only way would be to disable that update I guess (well you can increase it but lag will stay anyway just less frequent).

u1ps:
Is it even possible to disable these updates right now? 

Speaking of this table if i am correct?

image

ryzyk-krzysiek:
No it's not possible to disable it and it's not really an issue because you need to request it on your own and when you do it you are not playing.

EDIT: You could disable it in lua afaik but I see no reason.

u1ps:
/scores and scoreboard update different things then right?

Manual /scores request causes much bigger spike but the one in https://github.com/etlegacy/etlegacy/issues/2679 happens every 5 sec automatically (and is smaller in spike but still a spike). If there was way to disable it will lose some critical data?

Issue #1886, opened 2022-01-15
stuck midair


zenixje:
https://clips.twitch.tv/ObliqueSpookyPandaBigBrother-W13atm2B0rXgFCWS

This was on snapshot v2.79.0-7-gb5ad814. I have seen the same thing happen when we had the server on sv_fps 100.  But that was not the case here.

ryzyk-krzysiek:
Likely caused by antiwarp, when increasing `sv_fps`, `g_maxWarp` should be increased accordingly as explained here:
https://github.com/etlegacy/etlegacy/pull/1880

Aranud:
Old issue. To be reopen in case it is reproducible.

x0rnn:
This is definitely still happening. I thought it was something wrong with func_fakebrushes, but no.

[2023-11-26-022407-fueldump.zip](https://github.com/user-attachments/files/16347523/2023-11-26-022407-fueldump.zip)

Btw, we have sv_fps 20 and g_maxWarp 6

x0rnn:
Also regarding the demo above, you always get stuck at the same location, it's not random...

Aciz:
Any custom mapscripts? LUAs? This doesn't happen with default mapscripts and no LUAs loaded.

x0rnn:
Yes, but I checked and there are no func_fakebrushes there to begin with. Lua I don't know why or how it could cause that. Also the twitch video has no mapscript with func_fakebrushes I would assume.

mittermichal:
same thing happened to me on TM server few days ago on venice, will post demo and video later

x0rnn:
Stairs in venice, right?

mittermichal:
yes, those that tank goes through right after it passes AT gun

Aciz:
All of the stuck spots in these clips correspond to the spectator spawn point in each map, so very likely something to do with that.

PR #2685, opened 2024-07-24
game: init s.solid to 0, refs #1886

ryzyk-krzysiek:
- s.solid like many other fields are retained for client entities. If you connect to a slot where previous player had that field set (was in a team) then you have that field set because it is only set by server in trap_LinkEntity and spectators are never linked.
- This is VET bug and can be reproduced in other mods (ETMain, ETPro, ETJump), it's not as obvious there because legacy has different WolfReviveBbox function (which causes the floating), but you can reproduce the stuck print with developer 1. 
- It only happens on spectator spawn point because r.currentOrigin is not updated for spectators. 
- I wouldn't be surprised if there were more hidden issues like this because the gentity struct is not fully cleared for player entities.

refs #1886

### DIFF
--- a/src/game/g_utils.c
+++ b/src/game/g_utils.c
@@ -784,6 +784,7 @@ void G_InitGentity(gentity_t *e)
 	e->r.ownerNum = ENTITYNUM_NONE;
 	e->nextthink  = 0;
 	e->free       = NULL;
+	e->s.solid    = 0;
 
 	// init scripting
 	e->scriptStatus.scriptEventIndex = -1;


Aciz:
Is there any reason to just not completely wipe the `gentity_t` struct in `ClientBegin`? I don't see anything immediately harmful with that, though you might need to do some additional setup to init the player entity maybe?

ryzyk-krzysiek:
That's what I wanted to do at first but quickly run into a problem with bots where it totally broke them because it also clreared `r.svFlags` which is set before `ClientBegin` in `ClientConnect`. Imo it would be best if it would be cleared fully (somewhere) fe. non-player entities are cleared in `G_FreeEntity` so they wouldn't need to be memset again on init. 

But I decided it will be out of scope for this for me (I don't want to spend time right now on thinking how and where to do this so it makes sense and be 100% sure it doesn't break anything).

PR #2687, opened 2024-07-25
Chargebar needle

Vorschreibung:
Adds an option to add a needle on the chargebar indicating when the next charge is ready to be used.


Vorschreibung:
Quick Demo: https://www.youtube.com/watch?v=XtO80OLRD7o

PR #2693, opened 2024-07-29
cgame: Use ':' instead of '|' to separate akimbo ammo

Vorschreibung:
Vbar '|' causes vertical layouting boo boo's when using akimbo pistols compared to normal ammo display.

':' arguably also just fits better as '/' is already used to separate reserve ammo count.


Vorschreibung:
See 
Untitled

PR #2694, opened 2024-07-30
game: Fix Akimbo post-reload firing animation

Vorschreibung:
Due to a timing issue, for the first shot right after reloading the pistol is determined not by the freshly reloaded full clip, but by the clip state prior to the reload.

Fixed by postponing the akimboFire check.


Vorschreibung:
Big thanks to @Aciz's debugging help 🫶

PR #2695, opened 2024-07-31
game: Make medic's passive heal deterministic

Vorschreibung:
Medic's currently heal every second via 'ClientTimerActions', which loops continuously in the background.

This means that in case a Medic takes damage, the timer could be anywhere between 0.1 and 0.9 seconds off the next passive heal, effectively making it random in a way that neither the player nor the opponent can control.

This commit alleviates the issue by resetting and pausing the timer once the player has reached MAX_HEALTH, making the passive heal deterministic.

### DIFF
--- a/src/game/g_active.c
+++ b/src/game/g_active.c
@@ -838,6 +838,17 @@ void ClientTimerActions(gentity_t *ent, int msec)
 {
 	gclient_t *client = ent->client;
 
+   // reset and pause client timer when we've reached maxhealth, such that once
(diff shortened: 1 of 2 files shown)


Vorschreibung:
Could also pin the paused `residualTime` to `500` instead of `0` to pin it to the average of the status quo.

However, IMHO `0` would also be justifiable, which would constitute a teensy nerf on average to Medic's quite powerful passive HP regen compared to now.

PR #2698, opened 2024-08-01
game: Flamer reduce indirect dmg, boost direct dmg

Vorschreibung:
Reduce 2/3 sources of flamethrower damage by reducing damage from 5 to 4
	- indirect flame chunk damage
	- indirect burn dot damage

And then boost direct flamethrower damage (requires tracking aim, only against a single player) to 12
	- direct player hit flame chunk damage


x0rnn:
Controversial PR, needs more (or no at all) discussion before (if) flamethrower gets more powerful. It's easy to track same way it's easy to track with SMG, so 1 HP less indirect damage does not compensate for the fact direct damage gives a huge boost to it with this (12 HP vs 5 HP).

Vorschreibung:
Converted this to a draft, will clarify some things tho, because I think looking into this after the Flamer has been cleaned up is still worthwhile considering and the negative reaction is at least partially based on misunderstandings, which I can't follow:

1. Boosting direct damage but reducing indirect damage with the flamer can be made in a way that does not make the Flamer more powerful as a whole - but shift the skill ceiling instead. - The aim is to give the flamer a genuine skill ceiling, instead of a trivial weapon usable by everyone that is hard to balance without having it either too weak or too strong.

2. Tracking aim with SMG is very different than tracking aim with Flamer as of now (and can be further differentiated if desired):

- Flame Chunks don't give distinct visual feedback when being generated at all, while of course with an SMG it is very clear when each shot occurs

- Flame Chunks (currently) have a very short range compared to SMG

- Flame Chunks are projectiles that travel and have to collide with a player (i.e. they are not `hitscan` like the SMG)

3. The intention wasn't to put this in while there's flamer exploits that are currently in the process of being fixed. Of course first the flamer should be fixed for `sv_fps 40` and cleaned up of exploits and issues. - Only after that does this PR become relevant.

Vorschreibung:
While I'm not against discussions on this and of course many things can be discussed productively, I think it's impossible to substitute real playtesting with a discussion - assuming that you agree with the following premises:

1. The flamer is a low-skill-ceiling weapon (apart from knowing & using exploits that we want to fix) - which constitutes a balancing issue with the constant danger of having it either too weak or too strong

2. Boosting single target damage via tracking aim seems like the most productive way to introduce a higher skill ceiling.

----

I'd love discussion/proposals how a higher skill ceiling could be introduced without boosting single target damage for example.

Or possibly I'm wrong that the flamer doesn't already have a genuine non-exploitative skill ceiling and I'm not aware of it as of yet.

I think these are productive avenues for discussion - the details of this PR however, given that one agrees with the above premises, I think should be playtested and tweaked and discussion can't substitute that.

x0rnn:
>The aim is to give the flamer a genuine skill ceiling, instead of a trivial weapon usable by everyone that is hard to balance without having it either too weak or too strong.

This is simply not possible with flamethrower, the same way you can't do it with mortar. They are both trivial weapons where you just spam fire and it damages everything around it. The flamethrower is mostly already employed by noobs who can't aim otherwise, so they compensate with weapons that do damage in a radius. I'm afraid this would actually nerf it for those players whereas it would make it even deadlier (and more annoying) for those who can aim. I don't see the need to try and make the flamethrower an actual viable weapon instead of a trivial toy that it is and ought to be.

Vorschreibung:
> I don't see the need to try and make the flamethrower an actual viable weapon instead of a trivial toy that it is and ought to be.

I appreciate the honesty - however indeed I think this verdict is too pessimistic and that it's worth a try to make it more interesting for both players and opponents.

Given the current state of it, where the majority of it seemed to have "happened" more or less rather than looking like a deliberate, critically examined and playtested choice when the game was made, I would be surprised if the current state just happened to be the best possible game-design it could offer.

Issue #2655, opened 2024-07-17
Implement in-game Benchmark feature


u1ps:
Upgrade to /timedemo

---

Use 1 recorded demo as base for all the players running the benchmark for consistent results between different systems (clients). 

You would look for "X" amount of fps (avg) or drops below "X" fps. "X" being considered norm/recommended amount of fps by ETL client. 

1 click accessible feature from game options "Benchmark". Will give result after run and recommend what to do if fps is below recommended. Give option to change Quality preset 1 stage lower and rerun for quick and easy use. 

---

Good feature to have for new profile/player to quickly setup graphical options. 

Will be easier for players with less tech knowledge/graphical settings. (Amount of those payers is quite big actually)

Should be slightly more convenient to use for everyone who seek to troubleshoot fps

PR #2705, opened 2024-08-08
client: improve timedemo output and add benchmark command

Aciz:
* Add min, max, 99th percentile and 99.9th percentile minimum framerate to `timedemo` output
* Make it look prettier and improve the readability
* Add `benchmark` command - wrapper command that executes `timedemo 1; demo <demoname>`

The frametime storage should be enough for ~3h of demo playback from my calculations. It uses a circular buffer, so if it ever gets filled, it will simply start overwriting from the start, meaning that the start of the demo will be excluded from 99th/99.9th percentile framerates.

image

refs #2655

### DIFF
--- a/src/client/cl_cgame.c
+++ b/src/client/cl_cgame.c
@@ -1510,9 +1510,9 @@ void CL_FirstSnapshot(void)
 	cls.state = CA_ACTIVE;
 
 	// set the timedelta so we are exactly on this first frame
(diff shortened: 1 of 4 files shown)


Aciz:
Added another quite useful metric: Stability, represents the % of frames that were rendered at or above `com_maxfps` value, so basically how stable your desired FPS cap would be.

image

Issue #2269, opened 2023-02-20
Medpack/ammopack throw distance not long enough for diagonal throw


oheikk:
You will catch your own med pack / ammo pack if you aim diagonally in relation to the grid. This is also an issue in Vanilla ET.

I'm not sure if this should be considered a bug or a feature.

![image](https://user-images.githubusercontent.com/107760898/220198207-a9d7cd24-ad8f-4617-b6d8-f4538405b7a2.png)

ryzyk-krzysiek:
In my opinion this is vanilla ET feature. For collision detection with items bounding box is used instead of actual player capsule, this is why you can also pick up items through thin walls/doors.

As per VET comment:
> Items can be picked up without actually touching their physical bounds to make grabbing them easier 

https://github.com/etlegacy/etlegacy/blob/af544048b5c9ca82b29271015cfa818673e738f0/src/game/g_active.c#L474-L491

https://github.com/etlegacy/etlegacy/blob/af544048b5c9ca82b29271015cfa818673e738f0/src/game/bg_misc.c#L2521-L2547

oheikk:
I wouldn't suggest changing the item pick-up code, but rather the medpack/ammopack throwing code to use the $L^\infty$ norm:
![image](https://user-images.githubusercontent.com/107760898/220207449-ec43a975-40aa-4df2-8292-ab10fb3e4776.png)

oheikk:
Video for clarity because I realised the pictures might not be all that clear. _(edit: better video)_

https://user-images.githubusercontent.com/107760898/220313360-cf3ee150-38c7-4977-808c-e1742344d769.mp4

PR #2723, opened 2024-08-20
game: use actual player bbox values for item pickups

Aciz:
Magic number bad

refs #2269

### DIFF
--- a/src/game/bg_misc.c
+++ b/src/game/bg_misc.c
@@ -2532,7 +2532,7 @@ qboolean BG_PlayerTouchesItem(playerState_t *ps, entityState_t *item, int atTime
 
 	BG_EvaluateTrajectory(&item->pos, atTime, origin, qfalse, item->effect2Time);
 
-	return (vec3_distance(ps->origin, origin) < 36);
+	return (vec3_distance(ps->origin, origin) < (ps->maxs[0] - ps->mins[0]));
 }
 
 /**


src/game/bg_misc.c
```diff
@@ -2532,7 +2532,7 @@ qboolean BG_PlayerTouchesItem(playerState_t *ps, entityState_t *item, int atTime
 
 	BG_EvaluateTrajectory(&item->pos, atTime, origin, qfalse, item->effect2Time);
 
-	return (vec3_distance(ps->origin, origin) < 36);
+	return (vec3_distance(ps->origin, origin) < (ps->maxs[0] - ps->mins[0]));
```
mittermichal, line 2535: wouldn't using X value only cause some pickups items that have larger bbox X to be picked up easier than others?
Wouldn't it be more accurate to rather trace the bounding box?

Aciz: `ps->mins/maxs` are player boundaries, not item boundaries. The old code just simply hardcoded 36 as the value (`18 - -18`, which is the same as `ps->maxs[0] - ps->mins[0]`).

mittermichal: oh so its explained in the in the referenced issue, but still this doesn's seem like proper solution, nor that the solution was explained or tested

mittermichal: oh I see missed that its from ps, thanks

PR #2730, opened 2024-08-25
cgame: fix some particle effects with MissileCam

Vorschreibung:
Both 'CG_DrawMissileCamera' and 'CG_DrawActiveFrame' (which renders the "main" camera) were adding particle effects like the smoke of a WP_SMOKE_MARKER (the field ops smoke grenade) to their respective scene graphs via 'CG_AddLocalEntities'.

However, some of the consequent 'CG_Add*' calls entail checks to not-render and actually remove entities if they are close to the camera. (Note, remove the entity itself, not merely removing it from the respective scene graph).

Because both the MissileCam and the "Main"Cam would call these functions, a particle entity close to the MissileCam would actually be removed altogether and subsequently disappear also for the "Main"Cam.

This is circumvented by removing the delete calls (`CG_FreeLocalEntity`) and returning early before adding the entity to the scene graph.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -3372,7 +3372,7 @@ void CG_UpdateFlamethrowerSounds(void);
 void CG_InitLocalEntities(void);
 localEntity_t *CG_AllocLocalEntity(void);
 localEntity_t *CG_FindLocalEntity(int index, int sideNum);
-void CG_AddLocalEntities(void);
+void CG_AddLocalEntities(qboolean master);
 void CG_DemoRewindFixLocalEntities(void);
 
 // cg_locations.c
--- a/src/cgame/cg_draw.c
+++ b/src/cgame/cg_draw.c
@@ -4697,7 +4697,7 @@ void CG_DrawMissileCamera(hudComponent_t *comp)
 		CG_AddPacketEntities();
 		CG_AddMarks();
 		CG_AddParticles();
-		CG_AddLocalEntities();
+		CG_AddLocalEntities(qfalse);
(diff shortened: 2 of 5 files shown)


Vorschreibung:
Big thanks to @Aciz for helping debugging this, without which this patch wouldn't exist. 🫶

Vorschreibung:
Note that this does also change the smoke behavior, passing through the SM clouds now doesn't remove them altogether - so passing through a SM will have them reappear once you're out of the leash range.

Vorschreibung:
New PR, now retains the original soaking smoke cloud behavior

src/cgame/cg_local.h
```diff
@@ -3372,7 +3372,7 @@ void CG_UpdateFlamethrowerSounds(void);
 void CG_InitLocalEntities(void);
 localEntity_t *CG_AllocLocalEntity(void);
 localEntity_t *CG_FindLocalEntity(int index, int sideNum);
-void CG_AddLocalEntities(void);
+void CG_AddLocalEntities(qboolean master);
```
Aciz, line 3375: I'd name this something more specific than `master`, or at least add a comment what it signifies. Somebody's going to look at this in few months and wonder what the hell is this parameter.

Aciz (approved):

src/cgame/cg_local.h
Aranud: ^

Vorschreibung: Added comment description to latest commit

Vorschreibung:
@Aranud Documented the new parameter as instructed in latest commit.

PR #2732, opened 2024-08-25
etmain: tweak 1P rifles

Vorschreibung:
1. Reframe covi/engi rifles for smoother animation.

2. Fix Flash offsets for both rifles.

3. Fix covi rifles disappearing for a brief time right after de-scoping.

4. Offset the normal kar98 axis engi rifle brass ejection, as it has a
   high tendency to eject brass into the playerview in an obscuring way
   via the default 'tag_brass' tag.

5. Offset the scoped kar98 silencer, as it's misaligned by default

### DIFF
--- a/etmain/models/multiplayer/kar98/weapon.cfg
+++ b/etmain/models/multiplayer/kar98/weapon.cfg
@@ -17,19 +17,19 @@ newfmt
 ///   /   /   /   /   /   ______________________________/
 //   /   /   /   /   /   /
 
-0	1	20	1	0	0	0	// IDLE1 (for kar98)
-125	1	20	1	0	0	0	// IDLE2 (for gpg40)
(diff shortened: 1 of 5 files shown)


Aranud:
I'm not able to reproduce the point 4. with the brass ejection into the player view from current master. Any tips to reproduce it ?

Vorschreibung:
> I'm not able to reproduce the point 4. with the brass ejection into the player view from current master. Any tips to reproduce it ?

Should've added, run forward while shooting with the kar, you don't need to sprint.

Vorschreibung:
(at least with cg_fov 100)
image
image

Vorschreibung:
btw. 4. also seems to only happen once I go above 60 fps (my current display goes up to 144)

PR #2738, opened 2024-08-29
etmain: Skip 3P reload animations for PF/Flamer

Vorschreibung:
These are triggered sometimes (e.g. directly upon respawning), but they
make no sense as these weapons are instantly "reloaded" and all have no
1st person reload animation as well.

### DIFF
--- a/etmain/animations/scripts/human_base.script
+++ b/etmain/animations/scripts/human_base.script
@@ -2729,11 +2729,15 @@ reload
 	}
 	weapons bazooka
 	{
-		torso reload_rifle
+		NOOP
 	}
 	weapons panzerfaust
 	{
-		torso reload_panzer
+		NOOP
+	}
+	weapons flamethrower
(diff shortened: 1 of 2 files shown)


Vorschreibung:
Comparison: https://www.youtube.com/watch?v=NCDDwuCbAFs

Aranud (commented):
Why not remove the weapons declaration from reload section instead of adding a NOOP flag ?

Vorschreibung:
> Why not remove the weapons declaration from reload section instead of adding a NOOP flag ?

I tried that. Removing the pf/bazooka/flamethrower rules altogether would not work, there seems to be a default reload rule that then takes effect, although it's not explicitly stated in the file AFAIS.

Vorschreibung:
Leaving the rule body empty btw. does not parse. - I figured a NOOP command would be useful down the line anyhow.

PR #2740, opened 2024-08-29
Prioritize 3P torso reload anim when jumping

Vorschreibung:
Some weapons (like SMGs or Pistols) had their torso reload animation overriden by jumping.
    
This commit ensures that reloading torso animations are always shown when jumping during reload.

Also cleans up unneeded 'jump' and 'jumpbk' rules.


Vorschreibung:
Comparison: https://www.youtube.com/watch?v=v2vUsANjMoc

PR #2761, opened 2024-09-02
etmain: Canonicalize 3P 'aimless' weapon animations

Vorschreibung:
This commit introduces a new weapons category 'aimless' that subsumes
all weapons that should not be held in front of the player as if they
could aim with them (except pliers).

Along with this refactor, knives are now also held in this way, instead
of holding them like pistols.


Vorschreibung:
Comparison: https://www.youtube.com/watch?v=YzH3ex4f1z0

Vorschreibung:
Ready to review - now refactored and only knife is now not held like pistol - pliers remain untouched

Issue #2535, opened 2024-04-13
Add kill assist feature


u1ps:
Reward player with + 1 XP Battle Sense for kill assist.

Battle Sense seems most appropriate for it.

rmarquis:
This is already implemented in Battle Sense computation, via the cold/warm/hot bonuses.

u1ps:
With the new XP gain popup feature u can see the reward faster and u can see the assist in real time this way. And it makes sense to reward it as direct result of event. Many modern shooters have this feature too either as XP or just popup. 

Just a quick example i found from youtube

https://www.youtube.com/watch?v=FH_Ux_XaJuI

u1ps:
Teamkill should not count as assist for enemy player.

Example: Allied player shoots Axis player and that Axis player gets killed by teammate. In that situation Allied player would get assist. But that should not be assist since it was teamkill instead of kill.

Also Assist feature is part of ETL now so better to close this and open above as new issue?

PR #2763, opened 2024-09-02
mod: add kill assist + add team bleeding complaint, refs #2535

Aranud:
Kill assist a rewarded with Battle sense XP. 2 pts when more than half the max health has been removed, otherwise 1pts is granted. Only up to 4 players can be rewarded, selected by the highest amount of damage infliceted to the target.

g_teambleedComplaint has been added to control the minimum of damage required to max health to consider a team bleed has wanted. Negative value disable the feature.

Also add a new stats for highest assists in debriefing panels.


src/cgame/cg_draw.c
```diff
@@ -2732,8 +2734,8 @@ void CG_DrawVote(hudComponent_t *comp)
 		case -1: str = CG_TranslateString("Your complaint has been filed"); break;
 		case -2: str = CG_TranslateString("Complaint dismissed"); break;
 		case -3: str = CG_TranslateString("Server Host cannot be complained against"); break;
-		case -4: str = CG_TranslateString("You were team-killed by the Server Host"); break;
-		case -5: str = CG_TranslateString("You were team-killed by a bot."); break;
+		case -4: str = CG_TranslateString(va("You were %s by the Server Host", (cgs.complaintClient == CG_LastAttacker()) ? "team-killed" : "team-bleeded")); break;
+		case -5: str = CG_TranslateString(va("You were %s by a bot.", (cgs.complaintClient == CG_LastAttacker()) ? "team-killed" : "team-bleeded")); break;
```
Aciz: `team-bleeded` isn't a word, it's `team-bled`

src/game/g_stats.c
```diff
@@ -574,6 +574,98 @@ void G_AddKillSkillPointsForDestruction(gentity_t *attacker, meansOfDeath_t mod,
...
+						memmove(&bestTotalDamagePoints[j + 1], &bestTotalDamagePoints[j], sizeof(*bestTotalDamagePoints) * (ARRAY_LEN(bestTotalDamagePoints) - j - 1));
+					}
+
+					bestPlayersAssistReward[j] = i;
+					bestTotalDamagePoints[j]   = target->client->damageReceivedFromPlayers[i];
+					break;
+				}
+			}
+		}
+	}
```
Aciz: I wonder if it would be simpler to have the `damageReceivedFromPlayers` array as a 2D array with `[damage][clientnum]`, if that was the case I believe this whole thing could just be a qsort?

src/game/g_local.h
```diff
@@ -658,6 +658,7 @@ typedef struct
 	int spec_invite;
 	int spec_team;
 	int kills;
+	int kills_assists;
```
Aciz: `kills_assists` to me implies this included both kills and assists, I'd just use either `kill_assists` or `assists` for this.
```suggestion
	int kill_assists;
```
```suggestion
	int assists;
```

src/game/g_stats.c
```diff
@@ -574,6 +574,98 @@ void G_AddKillSkillPointsForDestruction(gentity_t *attacker, meansOfDeath_t mod,
...
+		ent = &g_entities[dmgReceivedSts[i] - target->client->dmgReceivedSts];
+
+		// ignore damage received from self and attacker (killer)
+		if (ent == target || ent == attacker)
+		{
+			continue;
+		}
+
+		// reward only opposite team
+		if (ent->client->ps.teamNum != target->client->ps.teamNum)
```
Aciz: This should be `sess.sessionTeam` for safety
```suggestion
		if (ent->client->sess.sessionTeam != target->client->sess.sessionTeam)
```

src/cgame/cg_debriefing.c
```diff
@@ -3948,6 +3970,7 @@ const char *awardNames[NUM_ENDGAME_AWARDS] =
 	"Highest Experience Points",
 	"Most Highly Decorated",
 	"Highest Fragger",
+	"Highest Assist",
```
Aciz: `Most Assists` is more fitting here imo, just like you have `Most Deaths`.

```suggestion
	"Most Assists",
```

There's also perhaps a conversation to be had here whether this should just be the number of assists, or most damage via assists, as the assists reward XP based on the amount of damage dealt, but that would require quite a bit of rework, so perhaps something to revisit in the future after some playtesting and feedback.

Aciz (changes requested):
Seems to work fine, I left some suggestions that aren't really critical (except perhaps the team check in `G_AddKillAssisPoints`), more of opinionated comments.

One thing that needs to be fixed though is `G_LoseKillSkillPoints`, it will always say the reason for losing XP is team kill, even if the complaint was filed for team bleeding.

https://github.com/etlegacy/etlegacy/blob/455ffb8b190dbfeff2764fcf4cddae998d00eeab/src/game/g_stats.c#L485-L499

Aciz (approved):

PR #2766, opened 2024-09-04
client: Add priority system for 3P animations

Vorschreibung:
This PR adds a simple plain value-based priority system for 3P animations:

- Add a new 'priority' int to 'animation_t' and parse it
- Add 0 'priority' values to all animations
- Reformat all animations to align their values vertically
- Rename 'prone_panzor_*' to 'prone_panzer_*'

And then assigns priorities to a basic set of animations:

- Touching ground now outprioritizes jumping animations (instead of
  jump-hovering while already on the ground).

- Run-jumping into a ladder causes the ladder animation now to outprioritize the
  still playing jumping animation.

- Priming grenades now outprioritizes raising grenades.

- Attacking with the knife now outprioritizes raising the knife.


Vorschreibung:
Ready for review

Vorschreibung:
Some things to point out for reviewing.

1. The whole thing is just adding a priority value to 3P animations and then forcing animation transitions if the queued animation priority is > current animation priority. - By itself this is nothing new, as the function already had a `qboolean force` that was passed in code at some spots. - Having priority values assigned to animations however moves this out of the code and makes it easier to maintain.

2. The 1st commit by itself ought not to change any animation behaviors.

3. The 2nd commit assigns priorities along the lines of:
```
Priorities:
    20 - weapon reload
    15 - weapon firing
    13 - weapon holding (grenades) / detach/attach
    10 - weapon raising
     8 - movement climbing
     5 - movement ground
     3 - movement air/jumping
     0 - idling / swimming / default
```
This is maintained as a comment in `./etmain/animations/human/base/body.aninc`.

This fixes some minor transition glitches as described in the commit and is not a substantial change in animation behavior.

This can and will be used more in the future to discern specific edge cases.

Vorschreibung:
Comparison: https://www.youtube.com/watch?v=Bf6zCfyn5Es (briefly goes a bit out of sync around the crates, but you get the gist).

Look closely for animation transitions, where previously some animations would have to be played in full for the next animation to play and now are interrupted by the new animation.

PR #2779, opened 2024-09-15
game: Don't play 'nopower' animation when proned

Vorschreibung:
When trying to throw a medpack without the required power while proned,
the torso would still play the 'e_no_juice' animation, which would make
the character deform into a snake of sorts.


Vorschreibung:
Example
image

PR #2781, opened 2024-09-15
client: Rely soley on 3P anim priority system

Vorschreibung:
Priority values for 3P animations currently co-exist with passing the `force` parameter in-code, which is manually set in various code-paths for various reasons.

This commit completely disables forcing 3P animations via code and instead soley relies on the priority values that are set in the '.aninc' files.

This greatly simplifies anim prioritizations, by making the priority value the de facto single source of truth.

Further changes are introduced due-to this:

- Switching weapons during a RAISE animation will now reset the RAISE animation properly (previously switching quickly between weapons would keep the currently running RAISE animation).

- Knife attacks now loop more smoothly.

- Attaching rifle grenades is now not canceled via ladder movement.

- Keep 'nopower' animation when jumping, instead of being broken by it.

And further changes are introduced alongside:

- Animations with the same priority value now override each other.

- Raising a knife while proned now does not play the 'prone_knife_attack' animation.

- Pliers now only play the firing animation as long as the attack button is held.


Vorschreibung:
Ready to review - I disagree with the Sonarcloud complaint.

Vorschreibung:
For reviewing the switching weapons point:

`cg_thirdPerson 1`, then switch quickly between knife and pistol by alternating pressing `1` and `2` - RAISE animation will now properly reset.

Vorschreibung:
Now also alongside a commit to remove all `force` references to 3P animations in the code.

Vorschreibung:
Now includes 3P revive reframe.

PR #2784, opened 2024-09-17
etmain: Reframe 1&3P Syringe

Vorschreibung:
Fit 1P 'ATTACK_LASTSHOT' / 3P 'firing_syringe' animation to gameplay hit check.

NOOP 3P 'raiseweapon', as drawing and firing syringe is too quick for both to play and it lerps unconsistently otherwise.

3P 'self_inject' already fits.


Vorschreibung:
Now also covers the prone animations that I've missed at first. Ready to review.

PR #2786, opened 2024-09-17
etmain: Cleanup 3P 'human_base'

Vorschreibung:


Vorschreibung:
Ready to review - now cleans up both movement-type and event descriptions as well.

Issue #1637, opened 2021-04-11
Raising sv_fps breaks certain gameplay elements


Aciz:
Some gameplay elements are hardcoded to only function properly with `sv_fps 20` (50ms update rate on the server). Raising `sv_fps` to higher values, for example 40, breaks these things. So far the commonly documented issues are following:

* Covert ops steals disguises faster 
  * https://streamable.com/izgdtw

* Flamethrower hit detection range is increased (no visual change, only hit detection is affected)
  * https://streamable.com/nsxilx (`sv_fps 40`) & https://streamable.com/n218ao (`sv_fps 125`)

* `script_mover` has slight hitching/movement speed increase between spline points/path_corners. Might not be instantly visible here, but in the past I have experienced issues in ETJump with raising `sv_fps` for recording smooth demos in localhost, some jumps with `scirpt_movers` in them become impossible.
  * https://streamable.com/20w7k5 (slight mover hitching with railgun tugs)
  * https://streamable.com/j0y8es (impossible jump with high `sv_fps` values)

* Pausing a match with `/ref pause` causes a visual bug where match timer will keep counting down. This is a visual bug only - the timer corrects itself after the match is resumed. Happens only on certain `sv_fps` values.
  * https://streamable.com/ksjieu

If higher `sv_fps` values are desired to possibly improve hit detection/general snappiness of the game, these issues should be resolved and the corresponding actions made independent of the 50ms update rate from the server.

rmarquis:
See related discussion on old ticket #285.

Especially, [this section](https://www.ra.is/unlagged/faq.html#ISFPSHT20SR) of the FAQ of unlagged:


>  Is setting sv_fps higher than 20 still recommended? (top)

> It's a somewhat common practice to raise the server framerate to 30 or 40 to make the gameplay “smoother.” What “smoother” means in this case is that Quake 3's built-in 50ms lag becomes 33ms or 25ms, and everyone sees things a bit more like how they happen on the server.
>
> Those are really the only two reasons to do it. Now, when Unlagged is running, full compensation or not, the built-in lag is taken care of completely. There goes one reason. It also does every hit test based on what players see rather than what's actually happening, so the second reason isn't any longer valid either.
>
> It does have some detrimental effects:
> 
> More incoming bandwidth is required of players
cl_timenudge can't extrapolate as much, meaning it's less effective (this becomes an issue when sv_fps is over 33)
Higher pings (it seems the client engine holds onto snapshots a little longer)
[...]

PR #2787, opened 2024-09-17
client: add cvar cl_interpolation, refs #1637

ryzyk-krzysiek:
- increase MAX_CLIENT_MARKERS from 17 to 40 (antilag buffer, supports 400ms ping (from 170ms) on sv_fps 100)

Note about cl_interpolation:
Buffers server packets to smooth over packetloss/ping instability. Values 0-4 depending on 'sv_fps' and 'snaps' (scales off the smaller value of the two). Set to 0 for most responsive gameplay.

Similiar to other engines the cvar allows to change the intepolation amount to smoother out the gameplay at the cost of responsivnes. Interpolation in ET engine by default is tied to sv_fps/snaps, meaning at sv_fps 20 interpolation is locked to 50ms, at sv_fps 100 interpolation is locked to 10ms.

I believe this turned out to be a problem few years ago when I tried to push for sv_fps 100 as some players reported that everyone is laggy. While I didn't have any information past that my assumption was and is that interpolation is just too small. sv_fps 100 means that a client expects new snapshot every 10ms. If a player uses com_maxfps 125 (8 ms frametime) that means that it basically expects a new snapshot every frame which could be very unreliable when not in a LAN setting but over internet. 

cl_interpolation is meant to help with that, by buffering additional snapshot(s) before rendering them and extending the window where a delayed/missing snapshot becomes an issue. Because as far as I know noone ever played ET on sv_fps 10, the default max interpolation is still 50ms, meaning on sv_fps 20 the cvar has no effect. 

On sv_fps 40 only allowed values are 0 or 1, which means that at most one additional snapshot can be buffered, the result of that is it will feel as if playing on sv_fps 20 (in terms of noticing a respones for example a hitsound). 

sv_fps 100 supports values from 0 to 4, 4 would mean as if playing on sv_fps 20, value 1 would mean intepolation of 20ms so faster than sv_fps 40 by 5ms, while not much it could be good enough for people that internet connection would not allow for enjoyable experience on sv_fps 100 with default 10ms interpolation.

cl_interpolation 0 sv_fps 40 sv_packetloss 50: https://youtu.be/XfKe-RfBras
cl_interpolation 1 sv_fps 40 sv_packetloss 50: https://youtu.be/SAoQUwAHsiQ
cl_interpolation 0 sv_fps 100 sv_packetloss 50: https://youtu.be/Fc9CWcaHZpU
cl_interpolation 1 sv_fps 100 sv_packetloss 50: https://youtu.be/ndITp2PHjVs
cl_interpolation 4 sv_fps 100 sv_packetloss 50: https://youtu.be/qGXXIo6VgGA

refs #1637

### DIFF
--- a/src/client/cl_cgame.c
+++ b/src/client/cl_cgame.c
@@ -208,6 +208,57 @@ qboolean CL_GetSnapshot(int snapshotNumber, snapshot_t *snapshot)
 	return qtrue;
 }
 
+extern cvar_t *sv_fps;
+
+/**
+ * @brief CL_InterpolationCheckRange
+ */
+static void CL_InterpolationCheckRange(void)
+{
+	int snaps, updateRate, buffer;
+
+	if (!cl_interpolation->integer)
+	{
+		return;
+	}
+
+	snaps      = Cvar_VariableValue("snaps");
+	updateRate = snaps < sv_fps->integer ? 1000 / snaps : 1000 / sv_fps->integer;
+	buffer     = (FRAMETIME / 2) / updateRate - 1;
+
+	if (cl_interpolation->integer > buffer)
+	{
+		Com_Printf(S_COLOR_YELLOW "WARNING: cvar '%s' is over allowed buffer (%d > %d), setting to %d\n", cl_interpolation->name, cl_interpolation->integer, buffer, buffer);
+		Cvar_SetValue(cl_interpolation->name, buffer);
+	}
+}
+
+/**
+ * @brief CL_SetSnapshotLerp
+ */
+void CL_SetSnapshotLerp(void)
+{
+	int i = 0, snapshotNumber = cl.snap.messageNum;
+
+	CL_InterpolationCheckRange();
+
+	while (i < cl_interpolation->integer)
+	{
+		if (cl.snapshots[--snapshotNumber & PACKET_MASK].valid)
+		{
+			i++;
+		}
+
+		// wrapped around
+		if (cl.snapshots[snapshotNumber & PACKET_MASK].messageNum == cl.snap.messageNum)
+		{
+			break;
+		}
+	}
+
+	cl.snapLerp = cl.snapshots[snapshotNumber & PACKET_MASK];
+}
+
 /**
  * @brief CL_SetUserCmdValue
  * @param[in] userCmdValue
@@ -1354,8 +1405,6 @@ int CL_FindIncrementThreshold(void)
 #define RESET_TIME  500
 #define HALVE_TIME  100
 
-extern cvar_t *sv_fps;
-
 /**
  * @brief Adjust the clients view of server time.
  *
@@ -1386,14 +1435,14 @@ void CL_AdjustTimeDelta(void)
 		return;
 	}
 
-	newDelta   = cl.snap.serverTime - cls.realtime;
+	newDelta   = cl.snapLerp.serverTime - cls.realtime;
 	deltaDelta = abs(newDelta - cl.serverTimeDelta);
 
 	if (deltaDelta > RESET_TIME)
 	{
 		cl.baselineDelta = cl.serverTimeDelta = newDelta;
-		cl.oldServerTime = cl.snap.serverTime;   // FIXME: is this a problem for cgame?
-		cl.serverTime    = cl.snap.serverTime;
+		cl.oldServerTime = cl.snapLerp.serverTime;   // FIXME: is this a problem for cgame?
+		cl.serverTime    = cl.snapLerp.serverTime;
 
 		if (cl_showTimeDelta->integer)
 		{
@@ -1443,8 +1492,8 @@ void CL_AdjustTimeDelta(void)
 				else
 				{
 					// must be done this way to avoid incorrect svFrameTime when on a slow client
-					svFrameTime = (cl.snapshots[(cl.snap.messageNum - 0) & PACKET_MASK].serverTime)
-					              - (cl.snapshots[(cl.snap.messageNum - 1) & PACKET_MASK].serverTime);
+					svFrameTime = (cl.snapshots[(cl.snapLerp.messageNum - 0) & PACKET_MASK].serverTime)
+					              - (cl.snapshots[(cl.snapLerp.messageNum - 1) & PACKET_MASK].serverTime);
 				}
 
 				// find the new threshold if not set or client/server frametime has changed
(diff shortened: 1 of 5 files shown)


HeDo88TH:
If this works It could be the turning point to allow high pingers to play on `sv_fps` and `snaps` 100

ryzyk-krzysiek:
Other than for antilag buffer high ping shouldn't really matter, cl_interpolation is pretty much all about ping stability (packets from server arriving at consistent interval) or server packet loss. People that reported issues with sv_fps 100 were around 50 ping.

ryzyk-krzysiek:
> cl_interpolation is pretty much all about ping stability (packets from server arriving at consistent interval) or server packet loss. 

Also client time drift, although it is possible in ETLegacy to adjust it with `cl_extrapolationMargin`

Aranud:
Neat

Oksii:
Possibly small bug? 

image

Seems to be checking for the client's sv_fps value rather than snaps. 
Setting the client's sv_fps to 40/50/100/125 enables the various steps

Aciz:
Regarding the comment left by @Oksii I think this is indeed incorrect?

https://github.com/etlegacy/etlegacy/blob/b2e8a578130b2ed33f9cc579215df4c090dd6f89/src/client/cl_cgame.c#L227

This would be the `sv_fps` value that is set on your client, and not the server value (which is what you'd want here I believe). Probably just an oversight from localhost testing.

The correct way to get the server fps would be I guess what `CL_AdjustTimeDelta` does:

https://github.com/etlegacy/etlegacy/blob/b2e8a578130b2ed33f9cc579215df4c090dd6f89/src/client/cl_cgame.c#L1496-L1501

ryzyk-krzysiek:
Yes it's an oversight by me, didn't actually realize the value is not properly set on client and I remember testing it too, so double fail. **For temporary solution** I would recommend forcing the value server side with `command "sv_cvar sv_fps EQ 40"`. 

I've quickly tried some first ideas but none worked. Currently the cvar is marked as `CVAR_SYSTEMINFO` but client can set it freely, same if changed to `CVAR_SERVERINFO`. I will try to think of some solution to this that makes the most sense (at least to me), because correct reporting of `sv_fps` from server to clients would be good to have in my opinion. I would like it to be fixed it in engine so any mod can utilize it, because as long as their snapshot reading in `cgame` is the same as in ETMain/Legacy then `cl_interpolation` cvar would work for them too.

> The correct way to get the server fps would be I guess what `CL_AdjustTimeDelta` does:
> 
> https://github.com/etlegacy/etlegacy/blob/b2e8a578130b2ed33f9cc579215df4c090dd6f89/src/client/cl_cgame.c#L1496-L1501

I believe this would be incorrect, `cl.snapLerp.messageNum - 0` and `cl.snapLerp.messageNum - 1` is not guaranteed to give `sv_fps` value when there is packetloss. That is why I check for snap validity here: https://github.com/etlegacy/etlegacy/blob/b2e8a578130b2ed33f9cc579215df4c090dd6f89/src/client/cl_cgame.c#L248-L251

The `CL_AdjustTimeDelta` during packetloss can and will return `svFrameTime` for invalid (old) snapshot and that time could even go as high as few thousands miliseconds. I still never debugged this code properly, but this could be a bug.

Aciz:
> I've quickly tried some first ideas but none worked. Currently the cvar is marked as `CVAR_SYSTEMINFO` but client can set it freely, same if changed to `CVAR_SERVERINFO`. I will try to think of some solution to this that makes the most sense (at least to me), because correct reporting of `sv_fps` from server to clients would be good to have in my opinion. I would like it to be fixed it in engine so any mod can utilize it, because as long as their snapshot reading in `cgame` is the same as in ETMain/Legacy then `cl_interpolation` cvar would work for them too.

There's `CL_SystemInfoChanged` in `cl_cgame.c` which is basically parsing the systeminfo configstrings every time it changes, so it could probably be cached there to some global variable or something, so it's always available in engine? This would work for both legacy and etjump (which also has it in systeminfo), for other mods I don't think it's necessarily that relevant as they'll be running on default `sv_fps` anyway, so you can just make an assumption that it's 20 (and from what I've understood, `cl_interpolation` cannot be changed on default `sv_fps 20` anyway).

Edit: the relevant bit for getting the systeminfo cs:

```cpp
char *systemInfo = cl.gameState.stringData + cl.gameState.stringOffsets[CS_SYSTEMINFO];
```

ryzyk-krzysiek:
As a note `CL_AdjustTimeDelta` `svFrameTime` could be computed for both known snapshots like this:

```
svFrameTime = (cl.snapshots[(cl.snapLerp.messageNum) & PACKET_MASK].serverTime)
	 - (cl.snapshots[(cl.snapLerp.deltaNum) & PACKET_MASK].serverTime);
```

It's still not guaranteed actual `sv_fps`, but at least it should never compute frameTime with snapshot that is outdated.

ryzyk-krzysiek:
> > I've quickly tried some first ideas but none worked. Currently the cvar is marked as `CVAR_SYSTEMINFO` but client can set it freely, same if changed to `CVAR_SERVERINFO`. I will try to think of some solution to this that makes the most sense (at least to me), because correct reporting of `sv_fps` from server to clients would be good to have in my opinion. I would like it to be fixed it in engine so any mod can utilize it, because as long as their snapshot reading in `cgame` is the same as in ETMain/Legacy then `cl_interpolation` cvar would work for them too.
> 
> There's `CL_SystemInfoChanged` in `cl_cgame.c` which is basically parsing the systeminfo configstrings every time it changes, so it could probably be cached there to some global variable or something, so it's always available in engine? This would work for both legacy and etjump (which also has it in systeminfo), for other mods I don't think it's necessarily that relevant as they'll be running on default `sv_fps` anyway, so you can just make an assumption that it's 20 (and from what I've understood, `cl_interpolation` cannot be changed on default `sv_fps 20` anyway).
> 
> Edit: the relevant bit for getting the systeminfo cs:
> 
> ```c++
> char *systemInfo = cl.gameState.stringData + cl.gameState.stringOffsets[CS_SYSTEMINFO];
> ```

Yep, in theory it is set in `CL_SystemInfoChanged`, but `CVAR_NONEXISTENT` check fails, `sv_fps` is loaded from etconfig afaik in my setup, so it exists and client doesn't set its value from server in `Cvar_Get`.

Then it is not set in the following block either, this could be skipped like pmove cvars, but I think the cvar itself could still be changed by user later again, so it's not exactly perfect solution in my mind right now.

https://github.com/etlegacy/etlegacy/blob/b2e8a578130b2ed33f9cc579215df4c090dd6f89/src/client/cl_parse.c#L678-L695

Like you have said (if I understood you correctly) we could just create some other mean to hold that value like serverId, pure server or cheats enabled is, but im not sure about it either, because with cvar it gives `cgame` the ability to look it up easily too and maybe some other utility I'm missing right now.

PR #2789, opened 2024-09-18
cgame: Speed up Limbo Team/Class Counter

Vorschreibung:
On servers with about 48 players, the team counter can take quite a long time to tick up to the current player count.

Speed it up, so we can see the current state quicker.


Vorschreibung:
Now also speeds up 'XP' counter.

PR #2792, opened 2024-09-18
cgame: Reset weapon smoke upon respawning

Vorschreibung:
Currently 'muzzleFlashTime', which determines the time that smoke generated from the muzzle after firing any weapon will not reset via respawning.

Fixed by resetting it on respawn.


Vorschreibung:
This is a tiny regression that seems to have happened between VET and ETL.

e.g.: As a Soldier with PF:

In VET self-killing via holding grenade does NOT make you spawn with PF generating smoke.
In ETL self-killing via holding grenade does make you spawn with PF generating smoke.

This commit alleviates the latter.

PR #2794, opened 2024-09-18
cgame: Fix MG42 bipod

Vorschreibung:
The MG42 bipod is currently broken and always appears expanded, when it ought to appear only when MG42 is deployed in prone.

This is fixed, which also addresses the following error line that would appear when wielding an MG42 in 3P:

R_AddMD3Surfaces: no such frame 0 to 4096 for 'models/multiplayer/mg42/mg42_3rd_bipod.md3' (2)


Vorschreibung:
Example of what this is about
image

PR #2799, opened 2024-09-19
etmain: Tweak M1 Garand 1P RELOAD anim

Vorschreibung:
Turns out that 'fps' is actually parsed via 'strtof', so we can finally float this to not have to make an always suboptimal choice between either trimming too much or leaving too much space at the end.


Vorschreibung:
Superseded by https://github.com/etlegacy/etlegacy/pull/2800

PR #2810, opened 2024-09-24
Add stopsound & stop weapon reload sound on obituary

Vorschreibung:


mittermichal:
is this stopping reload sound of some other player because he died?

Vorschreibung:
> is this stopping reload sound of some other player because he died?

Currently if you e.g. reload a weapon and then get killed while the reload sound plays, the reload sound is not interrupted but continues playing despite you having died.

You can simply test by reloading and selfkilling during warmup/instaspawn.

To fix this, this WIP PR attempts to add a way to stop channel sounds for a given player go interrupt the reload sound playing when Obituary happens.

Vorschreibung:
TODO:
- [ ] Make sure setting `.thesfx = NULL` actually "frees up" the channel down the line
- [ ] Add `S_AL_StopSounds` for `OpenAL` backend
- [ ] Consider some other combinations of narrowing the channels to be freed (e.g. by sound, by flags ...)

src/client/snd_dma.c
```diff
@@ -773,6 +773,55 @@ void S_Base_StartSoundEx(vec3_t origin, int entnum, int entchannel, sfxHandle_t
...
+
+	// shut off other sounds on this channel if necessary
+	for (i = 0; i < MAX_CHANNELS ; i++)
+	{
+		if (entchannel == CHAN_AUTO)
+		{
+			if (s_channels[i].entnum == entnum)
+			{
+				s_channels[i].thesfx = NULL;
+				// S_ChannelFree(&s_channels[i]);
```
Vorschreibung, line 811: I also seem to have misunderstood what `S_ChannelFree` does, because using it made it crash when I: `reload mp40 - selfkill - spawn with mg42 - shoot with mg42`.

Setting `.thesfx = NULL` seems to work however.

```
Thread 1 "etl.x86_64" hit Catchpoint 2 (signal SIGSEGV), S_ChannelMalloc (time=1252352) at /home/vor/workspace/etlegacy/src/client/snd_dma.c:245
245		freelist     = *(channel_t **)freelist;
[GDB] Caught signal SIGSEGV, Segmentation fault!
#0  S_ChannelMalloc (time=1252352) at /home/vor/workspace/etlegacy/src/client/snd_dma.c:245
#1  0x00005555555e7acd in S_Base_StartSoundEx (origin=0x7fffffffd99c, entnum=-1, entchannel=0, sfxHandle=687, flags=0, volume=64)
    at /home/vor/workspace/etlegacy/src/client/snd_dma.c:701
#2  0x00005555555e7ec2 in S_Base_StartSound (origin=0x7fffffffd99c, entnum=-1, entchannel=0, sfxHandle=687, volume=64)
    at /home/vor/workspace/etlegacy/src/client/snd_dma.c:831
...
```

Vorschreibung:
Actually turns this can already be done via `trap_S_StartSoundEx(NULL, es->otherEntityNum, CHAN_WEAPON, 0, SND_CUTOFF_ALL);` - also `EV_STOPSTREAMINGSOUND` at least sometimes clears the weapon sound already - this is another side-effect caused by instant spawning, that this ev. seems toget dropped

it also at least sometimes doesn't get emitted during WARMUP for reasons unbeknownst

Followup: https://github.com/etlegacy/etlegacy/pull/2811

Vorschreibung:
This btw. might be the same underlying issue that causes satchel explosions to not gib players when instaspawning and/or sometimes during warmup.

i.e. instantly spawning / spawning too fast, seems to mess with events currently in weird ways

PR #2811, opened 2024-09-24
cgame: Make sure reload sounds stop on death

Vorschreibung:
In at least some scenarios 'EV_STOPSTREAMINGSOUND' does not get emitted correctly, causing player reload sounds to continue despite death.

This is fixed by also stopping 'CHAN_WEAPON' sounds on 'EV_OBITUARY' client side.

### DIFF
--- a/src/cgame/cg_event.c
+++ b/src/cgame/cg_event.c
@@ -199,6 +199,9 @@ static void CG_Obituary(entityState_t *ent)
 		CG_Error("CG_Obituary: target out of range\n");
 	}
 
+	// this happens alongside 'EV_STOPSTREAMINGSOUND', as it sometimes doesn't seem to get emitted
+	trap_S_StartSoundEx(NULL, target, CHAN_WEAPON, 0, SND_CUTOFF_ALL);  // kill weapon sound (could be reloading)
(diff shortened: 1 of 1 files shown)


ryzyk-krzysiek:
> In at least some scenarios 'EV_STOPSTREAMINGSOUND' does not get emitted correctly

As far as I remember it's because:

1. For yourself if you die and respawn on the same server frame then the `ps.events` are deleted before thay are sent.
2. For other players even if they don't die and respawn on the same server frame `BG_PlayerStateToEntityState` will never happen if they go straight into limbo.

Vorschreibung:
> > In at least some scenarios 'EV_STOPSTREAMINGSOUND' does not get emitted correctly
> 
> As far as I remember it's because:
> 
> 1. For yourself if you die and respawn on the same server frame then the `ps.events` are deleted before thay are sent.
> 2. For other players even if they don't die and respawn on the same server frame `BG_PlayerStateToEntityState` will never happen if they go straight into limbo.

Thanks for the info! As this is tied to other events not being emitted, I'll give this a look later. - It would be really cool to "fix" in some shape or form.

PR #2814, opened 2024-09-25
cgame: Comment out recently unreferenced static vars

Vorschreibung:
Oversight/followup to 4a1d209a15733f6149a76f3ec2e652c381a401e2


Vorschreibung:
Otherwise causes warnings at compile time.

PR #2815, opened 2024-09-25
cgame: Hide 3P throwables briefly after throw

Vorschreibung:
Helps selling the idea that a player actually throws what they hold in their hands, rather than gesturing a throw while keeping the item in their hands at all times.


Vorschreibung:
Ready to review.

Commits/Changes ↓
```
cgame: Rename 'centity_t.muzzleFlashTime' to '.firedTime'
cgame: Always set 'centity_t.firedTime'
cgame: Hide 3P rifle grenade immediately on firing
cgame: Hide 3P throwables briefly after throw
cgame: Render 3P GPG40 Rifle Grenade
```

PR #2820, opened 2024-09-27
cgame: allow missile camera in demo playback

mittermichal:
useful for reviewing demos after game or making videos

### DIFF
--- a/src/cgame/cg_ents.c
+++ b/src/cgame/cg_ents.c
@@ -1432,7 +1432,7 @@ static void CG_Missile(centity_t *cent)
 		}
 	}
 
-	if ((cgs.sv_cheats || !cg.demoPlayback) && cent->currentState.clientNum == cg.predictedPlayerState.clientNum)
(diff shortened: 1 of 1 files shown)


src/cgame/cg_ents.c
```diff
@@ -1432,7 +1432,7 @@ static void CG_Missile(centity_t *cent)
 		}
 	}
 
-	if (cgs.sv_cheats && cent->currentState.clientNum == cg.predictedPlayerState.clientNum)
+	if ((cgs.sv_cheats || !cg.demoPlayback) && cent->currentState.clientNum == cg.predictedPlayerState.clientNum)
```
Aciz: This does the exact opposite :D it should be like this

```suggestion
	if ((cgs.sv_cheats || cg.demoPlayback) && cent->currentState.clientNum == cg.predictedPlayerState.clientNum)
```

mittermichal: damn, I tested it though 😄

mittermichal: it seemed working because I had sv_cheats 1 😄

mittermichal: fixed

PR #2821, opened 2024-09-27
Animation demo backwards compat

Vorschreibung:


mittermichal:
it would be nice to also check the "additional commits" number because at least one community is playing on relatively new built, but it would require additional parsing and because you did several animation commits, it might get little bit crazy to support everything, so maybe its not worth additional effort

Vorschreibung:
> it would be nice to also check the "additional commits" number because at least one community is playing on relatively new built, but it would require additional parsing and because you did several animation commits, it might get little bit crazy to support everything, so maybe its not worth additional effort

You mean in order to do demo backwards compat for different snapshots?
I'm not quite sure what the commit count otherwise would change.

mittermichal:
I tested this and there seem to be peculiar issue that when I replay demo with sv_pure 1 and I happen to have referenced mod pk3 (e.g legacy_v2.82.0.pk3) I get:
```
ERROR: CG_RegisterPlayerClasses: failed to load character file 'characters/temperate/axis/soldier.char' for the Axis Soldier
```
because its trying to find it only in legacy_v2.82.0.pk3 pak?

maybe easy solution is to disable sv_pure by default in replay ui

Vorschreibung:
> I tested this and there seem to be peculiar issue that when I replay demo with sv_pure 1 and I happen to have referenced mod pk3 (e.g legacy_v2.82.0.pk3) I get:
> 
> ```
> ERROR: CG_RegisterPlayerClasses: failed to load character file 'characters/temperate/axis/soldier.char' for the Axis Soldier
> ```
> 
> because its trying to find it only in legacy_v2.82.0.pk3 pak?
> 
> maybe easy solution is to disable sv_pure by default in replay ui

no, please check the commit - it only looks for specific animation files: https://github.com/etlegacy/etlegacy/commit/240890465666de47ca6a8139d2f70c6fbb82cab8#diff-cdd1eadafa271b18af4301b9faa839f2ad5df6bf6df8e496f1c96bd0b5f348c6R28

Vorschreibung:
~~@mittermichal please pass me the demo - either here or via Discord~~
nevermind, I can reproduce locally

Vorschreibung:
it seems this has always been the case and is unrelated to the animation demo backwards compat introduced in this PR

mittermichal:
> it seems this has always been the case and is unrelated to the animation demo backwards compat introduced in this PR

alright, it can be tackled in separate issue then

mittermichal:
> > it would be nice to also check the "additional commits" number because at least one community is playing on relatively new built, but it would require additional parsing and because you did several animation commits, it might get little bit crazy to support everything, so maybe its not worth additional effort
> 
> You mean in order to do demo backwards compat for different snapshots? I'm not quite sure what the commit count otherwise would change.

yes, we can also ask the server owner what snapshot he used 😄 

@Oksii do you have some good idea what snapshot you used in your servers mostly so we could make demo playback from them compatitible with newer mod?

Vorschreibung:
> yes, we can also ask the server owner what snapshot he used 😄
> 
> @Oksii do you have some good idea what snapshot you used in your servers mostly so we could make demo playback from them compatitible with newer mod?

That won't be possible tho - it makes absolutely no sense to keep demo playback compatibility between snapshot versions, or to add support for snapshot demo playback in a later release version, because the files have to be duplicated and kept in-tree for each backwards version.

We're not going to keep temporary states of snapshots in the code tree for the future.

Snapshots are temporary testing versions of current `master` that are going to likely change.

If you have snapshot demos, there is no guarantee that they will work in a future release.
If you want to watch them, keep the snapshot.

Aciz:
> I tested this and there seem to be peculiar issue that when I replay demo with sv_pure 1 and I happen to have referenced mod pk3 (e.g legacy_v2.82.0.pk3) I get:
> 
> ```
> ERROR: CG_RegisterPlayerClasses: failed to load character file 'characters/temperate/axis/soldier.char' for the Axis Soldier
> ```
> 
> because its trying to find it only in legacy_v2.82.0.pk3 pak?
> 
> maybe easy solution is to disable sv_pure by default in replay ui

Running `sv_pure 1` defeats the entire point of this, If you run with pure enabled, you're not running this code anyway, but rather the code in the module that's in the pk3.

And yes, `sv_pure 0` should be default, it was in 2.60b and I have no idea why it was changed in ETL.

mittermichal:
it does make sense t

> > I tested this and there seem to be peculiar issue that when I replay demo with sv_pure 1 and I happen to have referenced mod pk3 (e.g legacy_v2.82.0.pk3) I get:
> > ```
> > ERROR: CG_RegisterPlayerClasses: failed to load character file 'characters/temperate/axis/soldier.char' for the Axis Soldier
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > because its trying to find it only in legacy_v2.82.0.pk3 pak?
> > maybe easy solution is to disable sv_pure by default in replay ui
> 
> Running `sv_pure 1` defeats the entire point of this, If you run with pure enabled, you're not running this code anyway, but rather the code in the module that's in the pk3.
> 
> And yes, `sv_pure 0` should be default, it was in 2.60b and I have no idea why it was changed in ETL.

if I load some older version demo with sv_pure 1 it says "Initializing legacy cgame v2.82.1-300-g2408904" - head commit of this PR. I assume it doesn't unpack cgame from pk3 and just loads the unpacked one in fs_homepath

src/cgame/cg_demobackwardscompat.c
```diff
@@ -0,0 +1,33 @@
...
+	{
+		return qtrue;
+	}
+
+	return qfalse;
+}
+
+void CG_DemoBackwardsCompatInit()
+{
+	if (CG_DemoVersionBelow(2, 83, 0))
```
mittermichal: isn't it more correct to do version<=2.82.1 ? then it would not need to assume what next git tag is

Vorschreibung: doesn't matter in this case because we already know the next release number

Vorschreibung:
> it does make sense t

Care to give even the slightest hint of an actual argument?

mittermichal:
> > it does make sense t
> 
> Care to give even the slightest hint of an actual argument?

sorry i cut the sentence for some reason and I don't know what I wanted to say 😄

mittermichal:
my concussion is to not bother with "commit numbers". I had conversation with oksii and he explained that the players realize that they are playing on unstable mod version and it just used for specific gathers not official matches and its not a lot of games. It is not worth the effort.

Also agree with @Aciz that sv_pure should 0 by default, but that should be topic of anohter issue/PR

for me this PR looks good ❤️

Aciz:
> it does make sense t
> 
> > > I tested this and there seem to be peculiar issue that when I replay demo with sv_pure 1 and I happen to have referenced mod pk3 (e.g legacy_v2.82.0.pk3) I get:
> > > ```
> > > ERROR: CG_RegisterPlayerClasses: failed to load character file 'characters/temperate/axis/soldier.char' for the Axis Soldier
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > because its trying to find it only in legacy_v2.82.0.pk3 pak?
> > > maybe easy solution is to disable sv_pure by default in replay ui
> > 
> > 
> > Running `sv_pure 1` defeats the entire point of this, If you run with pure enabled, you're not running this code anyway, but rather the code in the module that's in the pk3.
> > And yes, `sv_pure 0` should be default, it was in 2.60b and I have no idea why it was changed in ETL.
> 
> if I load some older version demo with sv_pure 1 it says "Initializing legacy cgame v2.82.1-300-g2408904" - head commit of this PR. I assume it doesn't unpack cgame from pk3 and just loads the unpacked one in fs_homepath

This sounds like a bug to me, it makes no sense, the only reason you'd want pure in demo playback is to be able to load an older version, without needing to make a separate `fs_game` folder with older version installed.

src/cgame/cg_demobackwardscompat.c
```diff
@@ -0,0 +1,33 @@
...
+	{
+		strcpy(cg.demoBackwardsCompat.characterAnimationGroup, "animations/human_base_2_82_1.anim");
+		strcpy(cg.demoBackwardsCompat.characterAnimationScript, "animations/scripts/human_base_2_82_1.script");
+	}
+}
```
Aciz: `cg_main.c` already contains a function that checks for compatibility, should probably move that into this file

https://github.com/etlegacy/etlegacy/blob/af6404d1b335cf8160b7fdb93a6b5dce6a5d708a/src/cgame/cg_main.c#L3181-L3194

src/cgame/cg_character.c
```diff
@@ -450,6 +450,17 @@ qboolean CG_RegisterCharacter(const char *characterFile, bg_character_t *charact
...
 
+	// override for demo backwards compat
+	if (cg.demoBackwardsCompat.characterAnimationGroup[0] != '\0') {
+		CG_Printf("DEMOBACKWARDSCOMPAT: Loading animationGroup '%s'\n", cg.demoBackwardsCompat.characterAnimationGroup);
+		strcpy(characterDef.animationGroup, cg.demoBackwardsCompat.characterAnimationGroup);
+	}
+
+	if (cg.demoBackwardsCompat.characterAnimationScript[0] != '\0') {
+		CG_Printf("DEMOBACKWARDSCOMPAT: Loading animationScript '%s'\n", cg.demoBackwardsCompat.characterAnimationScript);
+		strcpy(characterDef.animationScript, cg.demoBackwardsCompat.characterAnimationScript);
```
Aciz: There's a safe string copy function in `q_shared.c`, should use that here rather than raw `strcpy`

https://github.com/etlegacy/etlegacy/blob/af6404d1b335cf8160b7fdb93a6b5dce6a5d708a/src/qcommon/q_shared.c#L1576-L1594

```suggestion
		Q_strncpyz(characterDef.animationGroup, cg.demoBackwardsCompat.characterAnimationGroup, sizeof(characterDef.animationGroup));
	}

	if (cg.demoBackwardsCompat.characterAnimationScript[0] != '\0') {
		CG_Printf("DEMOBACKWARDSCOMPAT: Loading animationScript '%s'\n", cg.demoBackwardsCompat.characterAnimationScript);
		Q_strncpyz(characterDef.animationScript, cg.demoBackwardsCompat.characterAnimationScript, sizeof(characterDef.animationScript));
```

Aciz (changes requested):

src/cgame/cg_demobackwardscompat.c
```diff
@@ -0,0 +1,33 @@
...
+
+	return qfalse;
+}
+
+void CG_DemoBackwardsCompatInit()
+{
+	if (CG_DemoVersionBelow(2, 83, 0))
+	{
+		strcpy(cg.demoBackwardsCompat.characterAnimationGroup, "animations/human_base_2_82_1.anim");
+		strcpy(cg.demoBackwardsCompat.characterAnimationScript, "animations/scripts/human_base_2_82_1.script");
```
Aciz: Oh and again here, use `Q_strncpyz`

```suggestion
		Q_strncpyz(cg.demoBackwardsCompat.characterAnimationGroup, "animations/human_base_2_82_1.anim", sizeof(cg.demoBackwardsCompat.characterAnimationGroup));
		Q_strncpyz(cg.demoBackwardsCompat.characterAnimationScript, "animations/scripts/human_base_2_82_1.script", sizeof(cg.demoBackwardsCompat.characterAnimationScript));

**Comment by Aciz** [comment]
A small suggestion: rather than adding these files prefixed with `_2_82_1`, they could perhaps just be placed in a folder called `compat/version`, so the file structure would look like this:
```
etmain
└── animations
    ├── human
    │   └── base
    │       └── compat
    │           └── 2_82_1
    │               └── xxx.anic
    └── scripts
        └── compat
            └── 2_82_1
                └── xxx.script
```

**Comment by Vorschreibung** [comment]
Ready for review again.

**Comment by Vorschreibung** [comment]
There's now a generalized shim for file loading in `cgame` that prefers a compatibility path, if set.

This PR includes backwards compatibility for `2.82.1` with the following file tree additions:
```
etmain/compat
└── 2_82_1
    └── animations
        ├── human_base.anim
        └── scripts
            └── human_base.script
```

File: src/cgame/cg_main.c [hunk 5]
```diff
@@ -2680,6 +2680,52 @@ void CG_LoadHudMenu(void)
...
+int CG_LoadCompatSource(const char *filename)
+{
+	if (cg.demoBackwardsCompat.compatPath[0] != '\0')
+	{
+		int res = trap_PC_LoadSource(va("%s%c%s", cg.demoBackwardsCompat.compatPath, PATH_SEP, filename));
+		if (res != 0)
+		{
+			if (developer.integer)
+			{
+				Com_Printf("CG_LoadCompatSource: %s\n", va("%s%c%s", cg.demoBackwardsCompat.compatPath, PATH_SEP, filename));
```
Aciz: In cgame, you'd usually use `CG_Printf` for printing to console (same in `CG_FOpenCompatFile`) as that is a cgame-specific print.

Aciz (approved):
Fundamentally seems okay, I don't necessarily agree with some of the naming stuff (e.g. generically named `demoBackwardsCompat_t` struct that only actually holds a path to an animation file, with a generic `compatPath` variable), but those are just semantics and can be addressed in the future with more backwards compatibility stuff being done.

Vorschreibung:
> Fundamentally seems okay, I don't necessarily agree with some of the naming stuff (e.g. generically named `demoBackwardsCompat_t` struct that only actually holds a path to an animation file, with a generic `compatPath` variable), but those are just semantics and can be addressed in the future with more backwards compatibility stuff being done.

I've kept the struct despite having only 1 field for future backwards compat areas that can't be fixed merely by overriding files via `compatPath`.
(e.g. adding backwards compat for older events etc.)

src/cgame/cg_main.c
Vorschreibung: Good point, done

Vorschreibung:
Addressed the Review Changes and rebased with `master`.

Issue #760, opened 2020-03-05
Pause issues


ETLegacyTeam:
Author: @saukko28 (Saukko)
Date: 2015-01-07
Redmine Issue: 760, https://dev.etlegacy.com/issues/760

---

There are two issues the pause command causes:
- If all the team's players are dead and you join a team while game is paused, you will see the result on the screenshot.
- The second issue comes when you follow these steps:
1. Die as the last player in the team
2. Pause the game
3. You will see a new player model showing the dead animation on top of your body

-*S

PR #2822, opened 2024-09-27
client: add possibility to pause sounds, refs #760

ryzyk-krzysiek:
- add engine extension `CG_MATCHPAUSED` to let engine know when match is paused
- add `PauseSounds` function to `soundInterface_t` which is called on `CG_MATCHPAUSED`
- add reload sound as pausable sound (maybe more in the future)
- add cvar `s_debugPause` (`CVAR_CHEAT`) for easier debugging

Note:
Adding pausable sounds is possible during the time when sounds are paused, which makes them play only after they are unpaused. This could be changed in cgame to not allow adding such sounds (by for example removing the flag before starting the sound), right now the developer needs to be aware of it and not add them. Looping sounds and some other sounds not supported because some engine calls don't support setting sound flags.

https://www.youtube.com/watch?v=h8TwN6dwN6A

refs #760


Vorschreibung:
Nice!

ryzyk-krzysiek:
Added sound pausing to OpenAL

Aranud:
I can't pause reload sound with OpenAL sound backend...

ryzyk-krzysiek:
I don't know, works just fine for me.

https://streamable.com/4u9yid

Aranud:
A full rebuild does the trick. Work fine, sorry.

PR #2825, opened 2024-09-29
cgame: Add 'cg_muzzleFlash 2' for hiding 1P only

Vorschreibung:
Setting 'cg_muzzleFlash' to '2' now hides muzzle flashes only for
yourself (1P) and not for others (3P).


Vorschreibung:
I disagree btw. with SonarCloud :)

Aranud:
Conflict ! I let you do the necessary to hold this situation :)

Vorschreibung:
> Conflict ! I let you do the necessary to hold this situation :)

Done 🫶

PR #2831, opened 2024-09-30
game: Retain old 'BG_PlayerTouchesItem' for CTF flags

### LINKED ISSUE #2827: New Entity pickup/touching regression
Reported on Discord by `adr` [link](https://discord.com/channels/260750790203932672/346956915814957067/1290058964494581912) - on latest snapshot with the new pickup/touch, picking up objectives on standard maps doesn't work as expected anymore.

https://github.com/user-attachments/assets/1538b44b-867b-4f0e-bff2-082aebb43193

Vorschreibung:
The newly introduced 'BG_PlayerTouchesItem' is working fine and dandy for medi/ammo packs, however it also affected mission objectives (CTF flags) in existing maps and ways that messed with player expectations.

This commit readds the old 'BG_PlayerTouchesItem' logic in 'BG_PlayerTouchesItemOld' and uses it whenever we are to intersect with CTF flag entities.

Fixes https://github.com/etlegacy/etlegacy/issues/2827

### DIFF
--- a/src/game/bg_public.h
+++ b/src/game/bg_public.h
@@ -2128,6 +2128,7 @@ void BG_AddPredictableEventToPlayerstate(int newEvent, int eventParm, playerStat
 
 void BG_PlayerStateToEntityState(playerState_t *ps, entityState_t *s, int time, qboolean snap);
 
+qboolean BG_PlayerTouchesObjective(playerState_t *ps, entityState_t *item, int atTime);
 qboolean BG_PlayerTouchesItem(playerState_t *ps, entityState_t *item, int atTime);
 qboolean BG_PlayerSeesItem(playerState_t *ps, entityState_t *item, int atTime);
 qboolean BG_AddMagicAmmo(playerState_t *ps, int *skill, team_t teamNum, int numOfClips);
--- a/src/game/bg_misc.c
+++ b/src/game/bg_misc.c
@@ -2519,8 +2519,32 @@ gitem_t *BG_FindItemForClassName(const char *className)
 }
 
 /**
+ * @brief Old Version of 'BG_PlayerTouchesItem' retained for intersect checking Objectives.
+ * @param[in] ps
+ * @param[in] item
+ * @param[in] atTime
+ * @return
+ */
+qboolean BG_PlayerTouchesObjective(playerState_t *ps, entityState_t *item, int atTime)
+{
+	vec3_t origin;
+
+	BG_EvaluateTrajectory(&item->pos, atTime, origin, qfalse, item->effect2Time);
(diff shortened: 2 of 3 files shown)


src/game/bg_public.h
```diff
@@ -2128,7 +2128,8 @@ void BG_AddPredictableEventToPlayerstate(int newEvent, int eventParm, playerStat
 
 void BG_PlayerStateToEntityState(playerState_t *ps, entityState_t *s, int time, qboolean snap);
 
-qboolean BG_PlayerTouchesItem(playerState_t *ps, entityState_t *item, int atTime);
+qboolean BG_PlayerTouchesItemNew(playerState_t *ps, entityState_t *item, int atTime);
+qboolean BG_PlayerTouchesItemOld(playerState_t *ps, entityState_t *item, int atTime);
```
Aranud: Perhaps a more generic / specific name would be welcome. 
Such as BG_PlayerTouchesItemSquare/CylinderBBox

src/game/g_local.h
```diff
@@ -307,6 +307,7 @@ struct gentity_s
 	vec3_t instantVelocity;         ///< per entity instantaneous velocity, set per frame
 
 	const char *classname;          ///< set in QuakeEd
+	long classnameHash;             ///< derived from classname
```
Aranud: Could it be separate in another PR ?

Aranud (changes requested):

src/game/bg_public.h
Aciz: I'd propose `BG_PlayerTouchesItem` & `BG_PlayerTouchesObjective`, since the other is used exclusively for objectives, and the other is for all other items

Vorschreibung:
Resolved all review input. Ready to re-review/merge.

ensiform:
Instead of classname matching you can compare item giType with IT_TEAM which is only true for PW_REDFLAG / PW_BLUEFLAG (objectives) aka the 2 team'd flags

Vorschreibung:
> Instead of classname matching you can compare item giType with IT_TEAM which is only true for PW_REDFLAG / PW_BLUEFLAG (objectives) aka the 2 team'd flags

~~that's on `gitem_t` tho - `hit` is `gentity_t`~~ nvm. didn't see `hit->item`

ensiform:
See 

```c
gentity_s::item
```

Which is valid pointer if its an item.

Vorschreibung:
> See
> 
> ```c
> gentity_s::item
> ```
> 
> Which is valid pointer if its an item.

Thanks! Exactly what we needed.

PR #2833, opened 2024-09-30
game: fix oob write to dmgReceivedSts when attacker wasn't a client

### LINKED ISSUE #2797: Spontaneous Demo playback mode during gameplay in v2.82.1-269-gcf348cb
When pressing `SHIFT + W` AND/OR `SHIFT + S` repeatedly and with speed, you are magically removed from the team and put into what appears to be some type of demo playback mode.  You cannot swap teams or go spectator, the only fix is to disconnect and reconnect.

Here is a full demo file and a clip for quick review:

https://www.twitch.tv/mayaaaan/clip/InspiringSmokyToadSeemsGood-p9pSmCMdArz17Btw

[2024-09-19-demobug-etl_frostbite.zip](https://github.com/user-attachments/files/17066351/2024-09-19-demobug-etl_frostbite.zip)

### LINKED ISSUE #2828: crushing player gives 30 max lives
[outlAw twitch clip](https://www.twitch.tv/outlawryy/clip/RepletePhilanthropicHerdYouDontSay-46pSwjM3yvtXRAb_)
[2024-09-29-213554-supply.zip](https://github.com/user-attachments/files/17181425/2024-09-29-213554-supply.zip) in this demo it happens at 8:23

dunno if it happened to other players at that moment, need to ask

Aciz:
`attacker->s.number` can be any entity in the map, which meant that this could easily cause an out of bounds write. This would manifest into this write getting written into the next `gclient_t` struct in `gentity_t` struct, which wrote garbage data to the next players playerstate most likely (as that's the first field in `gclient_t` struct).

fixes #2797
fixes #2828
refs #2535

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -1730,10 +1730,14 @@ void G_DamageExt(gentity_t *targ, gentity_t *inflictor, gentity_t *attacker, vec
 	{
 		if (attacker)
 		{
-			targ->client->ps.persistant[PERS_ATTACKER]                       = attacker->s.number;
-			targ->client->dmgReceivedSts[attacker->s.number].damageReceived += take;
-			targ->client->dmgReceivedSts[attacker->s.number].mods            = mod;
-			targ->client->dmgReceivedSts[attacker->s.number].lastHitTime     = level.time;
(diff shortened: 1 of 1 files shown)


mittermichal:
fixes: #2828 too

PR #2841, opened 2024-10-01
game: Support 'g_forcerespawn' == -1 to respawn instantly

Vorschreibung:
Also fix 'warmup' respawning to occur on the next server frame instead of the same as the player died to circumvent a set of issues that would otherwise occur.

### DIFF
--- a/src/game/g_active.c
+++ b/src/game/g_active.c
@@ -838,8 +838,8 @@ void ClientTimerActions(gentity_t *ent, int msec)
 {
 	gclient_t *client = ent->client;
 
-   // reset and pause client timer when we've reached maxhealth, such that once
-   // we take damage again, it will take a full second every time to re-heal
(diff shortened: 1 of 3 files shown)


src/game/g_active.c
```diff
@@ -1750,11 +1750,36 @@ void SpectatorClientEndFrame(gentity_t *ent)
...
+			// server frame
+			if (ent->client->respawnTime == 0)
+			{
+				ent->client->respawnTime = level.time;
+			}
+			else
+			{
+				do_respawn = qtrue;
+			}
 		}
```
Aciz: Join these 2 if statements? The bodies are identical.

Aciz (approved):
LGTM

Vorschreibung:
@Aciz please re-review, couldn't reuse existing `client->respawnTime` for all cases - added `client->instantRespawnDelayTime` instead.

Issue #2414, opened 2023-11-18
Show "This weapon is out of stock" message if not available


x0rnn:
On ETPro if you did "/team r 0 35" or had a bind for it it would select Mortar for your next spawn, and if it was unavailable it would print a message "This weapon is out of stock". On ETL there is no such message (using 51 as axis mortar index). You have to manually open the Limbo menu to check for availability.

x0rnn:
Similarly, in ETPro if you played as Medic for example and then changed class to Engineer with Rifle, if you checked in Limbo menu it would highlight your class as Engineer with Rifle. In Legacy, it still shows your class as Medic, and you won't know if you'll spawn as Engineer with Rifle or with MP40, because there is no message about the weapon not being available, thus you'd have to make an additional check in the Limbo menu and see if Rifle is available.

mittermichal:
this should be culprit: https://github.com/etlegacy/etlegacy/pull/2032/commits/9d92da6311f7a1045ec967dd0e8b344d806b6916

I was talking here https://github.com/etlegacy/etlegacy/pull/2844#issuecomment-2387087672 that I use a [script](9d92da6311f7a1045ec967dd0e8b344d806b6916) but only on TeamMuppet server which is still on 2.82.1, so it gives me the unavailability message with `cpm`.

x0rnn:
But that commit is 2 years old, that's before 2.81.1

mittermichal:
oh right I missed the year 😄

x0rnn:
But are you sure you get the unavailability message? What exactly does it say?

Aciz:
> But are you sure you get the unavailability message? What exactly does it say?

This is what @mittermichal means I guess

![image](https://github.com/user-attachments/assets/36c5f4e9-4a14-48c3-a066-b7b3110d0e44)

This only happens with `class` command, not `team` command. 

Edit: bind
`bind KP_PGUP "class e 2; play sound/menu/select; echo ^z[ ^7Engineer ^z: ^7Rifle ^z]"`

x0rnn:
Ah, so it should be a simple addition to make it work with /team too?

mittermichal:
just using class s 3 on TM 2.82.1:
![2024-10-04-135550-sw_oasis_b3](https://github.com/user-attachments/assets/1fe3e905-62f5-4323-b904-05e5ebc34773)

Is using team command worth using? Class command seems to select primary and secondary correctly now.

x0rnn:
[...]

PR #2844, opened 2024-10-01
cgame: Stop Class Limbo Selection take effect immediately by default

Vorschreibung:
It used to be the case that by default selecting a class automatically
makes it take effect - this however means that you can't look up via
limbo menu e.g. if a Soldier weapon is currently available without also
switching to the Soldier class.

This is by default now taken out, but can be re-enabled via
'cg_limboClassClickConfirm'.

---

Also add 'z' Accelerator as another way to confirm Limbo selection.

### DIFF
--- a/src/cgame/cg_limbopanel.c
+++ b/src/cgame/cg_limbopanel.c
@@ -1687,8 +1687,18 @@ qboolean CG_LimboPanel_ClassButton_KeyDown(panel_button_t *button, int key)
 
 			CG_LimboPanel_RequestWeaponStats();
 
-			cgs.ccManuallySetSecondaryWeapon = qfalse;
-			CG_LimboPanel_SendSetupMsg(qfalse);
+			// NOTE : It used to be the case that by default selecting a class
+			// automatically makes it take effect - this however means that you
+			// can't look up via limbo menu e.g. if a Soldier weapon is
+			// currently available without also switching to the Soldier class.
+			//
+			// This is by default now taken out, but can be re-enabled via
+			// 'cg_limboClassClickConfirm'.
+			if (cg_limboClassClickConfirm.integer)
+			{
+				cgs.ccManuallySetSecondaryWeapon = qfalse;
+				CG_LimboPanel_SendSetupMsg(qfalse);
+			}
 		}
 
 		return qtrue;
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2839,6 +2839,8 @@ extern vmCvar_t cg_skybox;
 extern vmCvar_t cg_redlimbotime;
 extern vmCvar_t cg_bluelimbotime;
 
+extern vmCvar_t cg_limboClassClickConfirm;
+
 extern vmCvar_t cg_movespeed;
 
 extern vmCvar_t cg_drawNotifyText;
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -222,6 +222,8 @@ vmCvar_t cg_quickMessageAlt;
 vmCvar_t cg_redlimbotime;
 vmCvar_t cg_bluelimbotime;
 
+vmCvar_t cg_limboClassClickConfirm;
+
 vmCvar_t cg_antilag;
 
 vmCvar_t developer;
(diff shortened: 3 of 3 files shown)


mittermichal:
I don't think this is good idea because many people are use to not clicking the OK button to switch class. This would be unpleasant surprise for them. 

I know exactly what you are trying to solve, but I think its better to not change it because lot of people will just be annoyed and won't really appreciate the positive, maybe they won't even notice.

I use class + weapon selection script to check availability, so it doesn't bother me much and for limbo menu I think it would be better to somehow display weapon availability right away somehow. Maybe as small red weapon icon around class icon.

Vorschreibung:
> I don't think this is good idea because many people are use to not clicking the OK button to switch class. This would be unpleasant surprise for them.
> 
> I know exactly what you are trying to solve, but I think its better to not change it because lot of people will just be annoyed and won't really appreciate the positive, maybe they won't even notice.

I don't think many people are used to it at this point, I suspect most people don't even know about this.
If they are, there's now 'y', 'z' and 'Return' as Accelerators to confirm a selection, so that should be trivial.

But I can see that some might prefer the old behavior, so I'll add a CVAR for this so if they complain users have a way to get the old behavior again.

Problem solved.

> I use class + weapon selection script to check availability, so it doesn't bother me much ...

I'm aware about class + weapon selection scripts, they don't matter, the limbo menu is an essential part of the game and the only thing newcomers will experience out of the box and it's usability matters.

> ... for limbo menu I think it would be better to somehow display weapon availability right away somehow. Maybe as small red weapon icon around class icon.

Please provide a mockup of how you're going to show e.g. that soldier does not have PF, Flamer and MG42 available without selecting the Soldier class and entering weapon selection box.

x0rnn:
> I use class + weapon selection script to check availability

What does the script look like? Because unlike in ETPro, legacy doesn't tell you a weapon is out of stock: https://github.com/etlegacy/etlegacy/issues/2414

Vorschreibung:
> > I use class + weapon selection script to check availability
> 
> What does the script look like? Because unlike in ETPro, legacy doesn't tell you a weapon is out of stock: #2414

just fyi, that class and weapon selection scripts are offtopic - this PR is about the limbo menu and the limbo menu only

Vorschreibung:
Ready to review. New commit with `cg_limboClassClickConfirm` to re-enable old behavior.

mittermichal:
> I don't think many people are used to it at this point

Anyone who doesn't use script or weaponmenu, uses limbo menu to switch class. So basically most of casual players ( my assumption based on experience with helping people setting up or just explaining them how to load cofig ) and also I have seen even competitive players use limbo menu.

You have to click to class to check availability because its not visible, if you solve this issue by viewing it immediately then people won't need to click on classes anymore.

Quick sketch of disabled pf and mg42:
image

Vorschreibung:
> > I don't think many people are used to it at this point
> 
> Anyone who doesn't use script or weaponmenu, uses limbo menu to switch class. So basically most of casual players ( my assumption based on experience with helping people setting up or just explaining them how to load cofig ) and also I have seen even competitive players use limbo menu.

I was never arguing that the limbo menu wasn't widely used, quite the opposite, which is why I've put this PR up.

What I am contending however is that users assume, rely or are even aware of class selection already taking effect when clicking on a class icon without confirming via `OK`.

My assumption is actually that most users are not even aware of this and always confirm via `OK`.

> You have to click to class to check availability because its not visible, if you solve this issue by viewing it immediately then people won't need to click on classes anymore.
> 
> Quick sketch of disabled pf and mg42: image

I'm well aware that in theory this would solve the issue - but I can't take your sketch seriously, you're not going to fit weapon icons around those class icons in any shape or fashion that is going to be legible.

So unless you are showing a POC or putting a PR, I have to assume what I have already assumed so far, namely that your suggestion is impractical and won't be done.

I certainly will not attempt it, because I've fiddled around enough with the limbo menu to know that this won't work and that it's a waste of time.

mittermichal:
> What I am contending however is that users assume, rely or are even aware of class selection already taking effect when clicking on a class icon without confirming via OK.
> My assumption is actually that most users are not even aware of this and always confirm via OK.

I see, I have no idea if other players use OK button for class selection. I don't remember ever pressing that button myself.

> I'm well aware that in theory this would solve the issue - but I can't take your sketch seriously, you're not going to fit weapon icons around those class icons in any shape or fashion that is going to be legible.

> So unless you are showing a POC or putting a PR, I have to assume what I have already assumed so far, namely that your suggestion is impractical and won't be done.

> I certainly will not attempt it, because I've fiddled around enough with the limbo menu to know that this won't work and that it's a waste of time.

Sure, was just trying to sketch my idea and I won't do any PR.

New players could also be pointed to using classmenu during game instead and the classmenu could be improved to show weapon availability. 

I'm not completely against this, it does have benefits. Just not sure if it won't make too many people annoyed. And I wanted to bring have discussion.

mittermichal:
Also this doesn't affect spawn selection in limbomenu, so that might be confusing.

Vorschreibung:
> Also this doesn't affect spawn selection in limbomenu, so that might be confusing.

Can be done in a future PR if need be, I don't see the need for it, exactly because of the very specific issue that I've outlined in the description and that this PR solves.

Vorschreibung:
> I'm not completely against this, it does have benefits. Just not sure if it won't make too many people annoyed. And I wanted to bring have discussion.

Please don't take this the wrong way, but I have no interest in entertaining ideas and having discussions on things that clearly won't happen - I have an interest to actually get something done that will have a positive net effect.

Towards this, I think your point that some people could dislike this is a valid concern and I think it's sufficiently addressed with the CVAR added in this commit.

I think we have enough info on this PR by now that devs can make a decision on it.

Vorschreibung:
Ready for review.

PR #2847, opened 2024-10-03
sys: Add simple cmdline arg parsing

Vorschreibung:
Also add '-h/--help' and '-t/--title' cmdline args, to show help text and set the window title respectively.

### DIFF
--- a/src/sys/sys_main.c
+++ b/src/sys/sys_main.c
@@ -76,6 +76,8 @@
 jmp_buf exit_game;
 #endif
 
+char *GlobalGameTitle = CLIENT_WINDOW_TITLE;
+
 static char binaryPath[MAX_OSPATH]  = { 0 };
 static char installPath[MAX_OSPATH] = { 0 };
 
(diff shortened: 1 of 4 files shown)


Vorschreibung:
@Aciz Please Review 🫶

src/sys/sys_main.c
```diff
@@ -935,6 +1003,26 @@ void Sys_ParseArgs(int argc, char **argv)
...
+		         !strcmp(argv[i], "-t"))
+		{
+			i++;
+
+			if (i >= argc)
+			{
+				Sys_ParseArgsError(va("Option '%s' expected an argument: <title>\n", argv[i - 1]));
+			}
+
+			GlobalGameTitle = argv[i];
```
Aciz: This should have an empty check and fallback to default value, right now you can pass in an empty title, which causes the game to not boot.

image

Also if you just do `-t -- +devmap fueldump`, it causes the game to spawn with title `--`, which is probably not desired.

src/sys/sys_main.c
```diff
@@ -913,17 +915,83 @@ void *Sys_LoadGameDll(const char *name, qboolean extract,
...
- * @brief Sys_ParseArgs
+ * @brief Parses passed process arguments
  * @param[in] argc
- * @param[in] argv
+ * @param[in,out] argv
  */
 void Sys_ParseArgs(int argc, char **argv)
 {
-	if (argc == 2)
+	for (int i = 0; i < argc; ++i)
```
Aciz, line 1026: ~~Declaring variables inside loop initializer isn't C89 compliant.~~

Engine is C99 so this is fine.

src/sys/sys_main.c
```diff
@@ -913,17 +915,83 @@ void *Sys_LoadGameDll(const char *name, qboolean extract,
...
+	fprintf(stream,
+	        "Usage:\n"
+	        "  <etl-bin> [option]* -- [server options]*\n"
+	        "\n"
+	        "Options:\n"
+	        "  -h, --help                 Show help and exit\n"
+	        "  -v, --version              Show version and exit\n"
+	        "  -t, --title <title>        Set the game window title\n"
+	        "\n"
+	        "Server Options:\n"
```
Aciz: This isn't strictly `server options` as these can be used on the client too, more appropriate name would be `launch options` or `startup options`.

Aciz (changes requested):

src/sys/sys_main.c
Vorschreibung: Thanks! Great points!

Vorschreibung: Resolved - with just one caveat ↓
```
<etl-bin> -t +devmap
```
can't be protected against, unless we also disallow starting a game title with `+` prefix - which I'd rather not do to allow passing any title a user wants to

I've added checks for preventing `<etl-bin> -t -- +devmap` and `<etl-bin> -t "" +devmap` tho, great catches! :heart_hands:

Vorschreibung:
Ready for re-review @Aciz :heart_hands:

Vorschreibung:
I disagree with SonarCloud btw. - ready for review again.

src/sys/sys_win32_con.c
```diff
@@ -700,12 +700,14 @@ void Sys_CreateConsole(void)
...
+	char windowname[MAX_STRING_CHARS] = { 0 };
 #if defined (UPDATE_SERVER)
-	const char *WINDOWNAME = "ET: Legacy Update Server";
+	Q_strncpyz(windowname, "ET: Legacy Update Server", sizeof(windowname));
 #elif defined(DEDICATED)
-	const char *WINDOWNAME = "ET: Legacy Dedicated Server";
+	Q_strncpyz(windowname, "ET: Legacy Dedicated Server", sizeof(windowname));
 #else
-	const char *WINDOWNAME = "ET: Legacy Console";
+	Com_sprintf(windowname, sizeof(windowname), "%s Console", GlobalGameTitle);
```
Aciz, line 718: This gets called before command line args are parsed, so it doesn't work, but that's really not a critical thing to fix, can be looked at later as it could be rather complex to do.

Aciz (approved):

PR #2848, opened 2024-10-03
etmain: New death animations

Vorschreibung:
Reframed deaths that skip the flailing with arms around parts of existing death animations.

These got approved by competitive players on Discord.

### DIFF
--- a/etmain/animations/human/base/body.aninc
+++ b/etmain/animations/human/base/body.aninc
@@ -22,22 +22,33 @@
 		//	|   				|   	|   |   |   	|   |   |   <priority>
 		//	|   				|   	|   |   |   	|   |   |   |
 
-		// DEATHS {{{ - do not change the order!
-		death_machinegun_1		0		32	0	20		0	0	0	99
-		dead_machinegun_1		31		1	1	20		0	0	0	99
-		//death_rifle_head_2	33		52	0	20		0	0	0	99
-		death_rifle_head_2		43		42	0	25		0	0	0	99
-		death_crouch			56		26	0	25		0	0	0	99
-		dead_rifle_head_2		84		1	1	20		0	0	0	99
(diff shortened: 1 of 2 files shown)


etmain/animations/human/base/body.aninc
```diff
@@ -22,22 +22,14 @@
 		//	|   				|   	|   |   |   	|   |   |   <priority>
 		//	|   				|   	|   |   |   	|   |   |   |
 
-		// DEATHS {{{ - do not change the order!
```
Vorschreibung, line 25: btw. I've retained the `do not change the order!` warning so far, but after checking a couple of times and thinking about it - I see no reason anymore why the order would matter, especially with the priority system now.

Any check that checks for specific death animations is brittle to begin with (no such thing exist in the code) and if it has to, should rely on either the existing `name` or `nameHash` and error if they couldn't be found, instead of an animation index number that can easily change.

---

The warning came from 2012 with the first commit, there's no further history on who wrote it or why it's there.

I propose to nuke it, which is already what the commit does.

Vorschreibung: I've finally figured out why this warning was there and put it in with a proper explanation of what it is about.

The first frames of the the first animation defined of the `body.aninc` file is used for a single or two frames when a player spawns and then lerps to the animation we actually want and more cases.

I've added copious amounts of documentation on it and retained the current animation for now.

Can be possibly improved later on.

Vorschreibung:
here's a comparison for the initial respawn frames

BEFORE
https://github.com/user-attachments/assets/b943150a-7d8c-4661-9985-ae3b3b56a444

AFTER
https://github.com/user-attachments/assets/09177ad8-44d9-48f7-82c9-61b97b48e773

Vorschreibung:
Ready to review.

Vorschreibung:
Ready to review

PR #2849, opened 2024-10-03
cgame: Change 'cg_centertime' default from 5 to 3

Vorschreibung:
5 was way too high, note that this value * 2 = the amount of seconds any centerprint message is shown.

i.e. 3 amounts to 6 seconds


Vorschreibung:
Misjudged what this does.

PR #2855, opened 2024-10-08
etmain: Fix 1P pliers arm clipping

Vorschreibung:
The arm was clipped too early for any cg_fov above 90 for 16:9.

This commit lengthens the arm and also fixes an obviously bad UV wrap that always only used a small portion of the actual arm skin.


Vorschreibung:
before / after
image

Vorschreibung:
Ready for review

PR #2857, opened 2024-10-08
etmain: Fix Sten barrel pits

Vorschreibung:
Add an updated Sten skin that fixes an obvious UV Wrap issue on the pits of the barrel, whereby one of the pits was positioned on one of the last pixel lines then got stretched over roughly 1/4th of the surface.

Also add some specular detail to the mag.


Vorschreibung:
before / after
image
image

PR #2859, opened 2024-10-08
tvgame: logic improvements and new tvg_autoAction cvar, refs #229

ryzyk-krzysiek:
- rename game system call
- move configstring initialization to `SV_InitGameVM`, this means that now it is possible to look up configstrings during `GAME_INIT`. (not possible in ettv, so anything that relies on it won't work there)
- parse `CS_WARMUP` and `CS_WOLFINFO` on `GAME_INIT` and `GAME_CLIENT_COMMAND`
- add `tvg_autoAction` cvar, bit values `AA_DEMORECORD 1 AA_STATS 2`. Autorecord start demo recording the same way `cg_autoaction` does with client demos (with the same name schema too), only difference as of now is that demo will not be stopped during intermission (client demos are round seperated because of this). Autostats controls if tv server will periodically send commands requesting stats.
- improve handling of spawn points, especially if game is delayed the spawnpoint can end up being very wrong (improvement over etpro tvgame)
- remove unused code
- fix animation files breaking mod loading on different mods than legacy
- extract pmove code from `bg_pmove` to `tvg_pmove`
- fix various pmove prediction issues on mods like ETJump or ETPro
- remove no longer needed `bg_*` files
- overwrite master server `CS_SVCVAR` configstring
- properly shutdown server on `tv disconnect`

Note: `sv_etltv_autorecord 1` is equal to `cl_autorecord 1` and `tvg_autoAction & 1` is equal to `cg_autoaction & 1`. So `sv_etltv_autorecord` records everything, while `tvg_autoAction` starts at the warmup countdown or `GS_PLAYING` (but doesn't end on itermission so whatever is after will keep being recorded. It's not impossible to change in the future if wanted.

refs #229, #2888


src/tvgame/tvg_client.c
```diff
@@ -108,10 +108,23 @@ void SP_info_player_start(gentity_t *ent)
...
- *
- * @param ent - unused
  */
 void SP_info_player_intermission(gentity_t *ent)
+{
+	if (ent->spawnflags < 0 || ent->spawnflags > 2)
+	{
+		Com_Printf("SP_info_player_intermission: incorrect spawnflag %d\n", ent->spawnflags);
+		return;
+	}
```
Aciz: I haven't looked at how the spawn implementation works here so this might not be relevant, but I think this can potentially be wrong. Spawnflags are bitflags, so it's possible to have an `info_player_intermission` entity with `spawnflags 3`, meaning the intermission point is the same regardless of which team wins.

ryzyk-krzysiek: Ah, you are most likely right because `FindIntermissionPoint` does bitwise operation on it but my brain didn't record it untill you pointed it out, thanks I will look into how handle this.

mittermichal:
> move configstring initialization to SV_InitGameVM, this means that now it is possible to look up configstrings during GAME_INIT. (not possible in ettv, so anything that relies on it won't work there)

what advantage does this bring? is it for checking warmup state?

ryzyk-krzysiek:
> > move configstring initialization to SV_InitGameVM, this means that now it is possible to look up configstrings during GAME_INIT. (not possible in ettv, so anything that relies on it won't work there)
> 
> what advantage does this bring? is it for checking warmup state?

The advantage is that you can do it if needed (which didn't make sense that it didn't work in the first place, you could only look up configstrings on 8th+ game frame), right now it only is really used for gamestate storing in `level.gamestate`, which isn't actually necessary because `g_gamestate` cvar works just fine, but maybe in the future it will find more uses for other things.

ryzyk-krzysiek:
But also, gamestate cvar is unreliable in ettv too, I fixed that in etl few patches ago.

mittermichal:
`... +sv_etltv_delay 0 +tv connect uk.ethub.pro:27960 ... ...`

```
148250 Dropped server frame: systimeDelta [95] sv.time [6343300] svcl.serverTime [6343300] frameDelta [6343350]
]  148400 Dropped server frame: systimeDelta [101] sv.time [6343350] svcl.serverTime [6343350] frameDelta [6343450]
]  148550 Dropped server frame: systimeDelta [101] sv.time [6343450] svcl.serverTime [6343450] frameDelta [6343550]
]  148700 Dropped server frame: systimeDelta [100] sv.time [6343550] svcl.serverTime [6343550] frameDelta [6343650]
]  148850 Dropped server frame: systimeDelta [98] sv.time [6343650] svcl.serverTime [6343650] frameDelta [6343700]
  148900 broadcast: print "/// entered the game\n"
  148900 broadcast: print "TiM TASTE entered the game\n"
  148900 broadcast: print "tzX.. entered the game\n"
  148900 broadcast: print "$yndykatorr. entered the game\n"
  148900 broadcast: print "zeifO entered the game\n"
  148900 broadcast: print "Rgg entered the game\n"
  148900 Connected to a pure server.
  148900 ==== TVShutdownGame (1 - te_escape2) ====
  148900 gametime: Wed Oct 09 13:26:21 2024
  148900 Sys_LoadDll -> Sys_TryLibraryLoad(/Users/xmitter/dev/etlegacy/build, legacy, tvgame_mac)... 
  148900 -- Trying zipped .bundle... failed (not a valid zip)
  148900 -- Trying extension-less dylib... succeeded
  148900 Sys_LoadDll(legacy/tvgame) found vmMain function at 0x10670e52c
  148900 Initializing legacyTV game v2.82.1-321-g21246e2
  148900 ------- TVGame Initialization -------
  148900 gamenametv: legacyTV
  148900 gamedate: Oct  9 2024
  148900 51 cvars in use
  148900 Not logging to disk
  148900 map: te_escape2
  148900 gametype: Stopwatch
  148900 gametime: Wed Oct 09 13:26:21 2024
  148900 Lua 5.4 API: no Lua files set
  148900 Enable spawning!
  148900 Disable spawning!
  148900 WARNING TVG_PickTarget: target t1072 not found or isn't in use - this might be a bug (returning NULL)
  148900 WARNING TVG_PickTarget: target t1073 not found or isn't in use - this might be a bug (returning NULL)
  148900 Game Initialization completed in 0.01 seconds
```

It seems that something is not initialized because the frame dropping stops after warmup countdown finished (game init I susspose)

I have asked hedo if I can give you tv password if you want to try, but maybe this is reproducible on localhost easily. I haven't tried yet.

mittermichal:
I think I had same issue also on master branch so it should be not related to this PR. I will keep testing and make separate issue I guess

ryzyk-krzysiek:
Frame dropping means that there was missing/delayed snapshot from master server, `systimeDelta` indicates how much real time has passed since last frame with previous snapshot (at sv_fps 20 you should expect one every 50ms, the error indicates you had a few every ~100ms). Unless this happens all the time it's not that big of a deal, it happens in ettv too. IIRC ideally it should only ever happen when master server is restarting/during `GAME_INIT` because it might take a while so the time between snapshot increases. But network issues like packetloss will trigger it too.

There maybe is some room for improvement for time handling but it's a bit tricky. For example I think there could be snapshot interpolation added to server so it would be less prone to dropped frames, but then viewers watching the game would see further interpolated playerstates instead of the "real" ones sent out from master.

ryzyk-krzysiek:
- tvgame: fix animation files breaking mod loading on different mods than legacy

BG animation parser was failing to load different mods animation files because of various changes that have been made in legacy mod.
Animation parsing is not required, but I've had to put it back in some time ago because of some weird random crashes in pmove.
This time relevant pmove code was extracted from bg_pmove and most bg_ files are no longer needed.
That also made it possible to fix various pmove prediction issues on mods like ETJump or ETPro, which are incompatible with Legacy by default.

mittermichal:
I have been using etlded and tvgame from this PR for few weeks in gather games and competitive matches (with addition of watchdog fix https://github.com/mittermichal/etlegacy/commit/f03b9ba16203f2a269a5776a4919741d3ee98022 discussed here https://github.com/etlegacy/etlegacy/issues/2887#issuecomment-2413944335 ) and it works fine so far.
There are some minor issues and possible improvements but they could be solved separately, I can add them as issues.

mittermichal:
https://github.com/etlegacy/etlegacy/pull/2859/commits/54a34bf21ba6b8cbf6b29e566cce218e5cb91520 works fine

PR #2866, opened 2024-10-10
game: Track missing vars for 'ClientHitboxMaxZ' in antilag

Vorschreibung:
These were missing in antilag markers, but are now required for calculating hitboxes correctly.

### DIFF
--- a/src/game/g_antilag.c
+++ b/src/game/g_antilag.c
@@ -152,6 +152,8 @@ void G_StoreClientPosition(gentity_t *ent)
 	ent->client->clientMarkers[top].pm_flags   = ent->client->ps.pm_flags;
 	ent->client->clientMarkers[top].viewheight = ent->client->ps.viewheight;
 
+	ent->client->clientMarkers[top].groundEntityNum = ent->client->ps.groundEntityNum;
+
 	// Torso Markers
 	ent->client->clientMarkers[top].torsoOldFrameModel = ent->torsoFrame.oldFrameModel;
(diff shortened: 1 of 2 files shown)


Vorschreibung:
@ryzyk-krzysiek Please review :pray:

src/game/g_antilag.c
```diff
@@ -314,6 +336,10 @@ static qboolean G_AdjustSingleClientPosition(gentity_t *ent, int time)
 			ent->torsoFrame.pitchAngle    = ent->client->clientMarkers[j].torsoPitchAngle;
 			ent->torsoFrame.yawing        = ent->client->clientMarkers[j].torsoYawing;
 			ent->torsoFrame.pitching      = ent->client->clientMarkers[j].torsoPitching;
+			if (ent->client->clientMarkers[j].torsoAnimationMovetype)
```
ryzyk-krzysiek, line 339: I guess these checks could be incorrect if `movetype 0` is a valid move type (I don't know if it is). Maybe checking if `ent->torsoFrame.animation` is valid pointer is better, or assuming that these pointers are always valid (`ClientHitboxMaxZ` does not check if it is valid so if it is valid there it should be valid here too I think).

Vorschreibung: 0 is `ANIM_MT_UNUSED` - should never be the case
I'll add the pointer check tho, thanks!

ryzyk-krzysiek: Then I have no further remarks, everything looks correct.

Issue #180, opened 2020-03-05
Map supply doesn't start (supply.pk3)


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2013-01-28
Redmine Issue: 180, https://dev.etlegacy.com/issues/180

---

"
ERROR: G_Script_ScriptParse(), Error (line 15): unknown action: create.
"

The map script uses missing ETpro script actions/cmds not implemented in legacy. I think we should add these before the new RC.

---

Related commit(s):
d71ff9447c99f6a5fe5f1ccf88cadc1a280f926f - game: script actions create & delete refs 180
 56a4182a167ecec627166eb618073775e570a141 - game: script actions create & delete II fixes 180
 61f5486d8f6195348bf43674fd9b8a8ce2194c82 - game: ETPro mapscripting fixes refs 258 refs 180

ETLegacyTeam:
From: @IR4T4 (IR4T4)
Date: 2013-01-28T22:20:05Z

---

Applied in changeset commit:56a4182a167ecec627166eb618073775e570a141.

PR #2868, opened 2024-10-10
mod: fixes to "create", fix CONTENTS_PLAYERCLIP fakebrushes

Aciz:
* Removed implicit linking of entities created via `create` script action, or equivalent LUA hook. Entities should handle linking in their spawn function, if they require linking.
* `func_fakebrush` now handles prediction correctly if contents is set to `CONTENTS_PLAYERCLIP`, this allows creating fakebrushes that only block player movement but don't interfere with gameplay by blocking bullets and projectiles.
* `func_fakebrush` now requires `origin`, `mins`, `maxs` & `contents` to be set explicitly, to avoid unexpected default values to be used implicitly.

refs #180

### DIFF
--- a/src/game/g_lua.c
+++ b/src/game/g_lua.c
@@ -1303,10 +1303,9 @@ static void _et_gentity_getweaponstat(lua_State *L, weapon_stat_t *ws)
 
 gentity_t *G_Lua_CreateEntity(char *params)
 {
-	gentity_t *create;
-	char      *token;
-	char      *p = params;
-	char      key[MAX_TOKEN_CHARS], value[MAX_TOKEN_CHARS];
+	char *token;
+	char *p = params;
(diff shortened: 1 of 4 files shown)


src/game/g_lua.c
```diff
@@ -1349,17 +1348,8 @@ gentity_t *G_Lua_CreateEntity(char *params)
...
-	//create->classname = "lua_spawn"; // make additional param?
-
-	if (!create)  // don't link NULL ents
-	{
-		return NULL;
-	}
-
-	trap_LinkEntity(create);
-	return create;
+	return G_SpawnGEntityFromSpawnVars();
```
Aciz, line 1352: For clarity: `G_SpawnGentityFromSpawnVars` returns `NULL` if the creation fails, so we don't need to explicitly check for that.

PR #2872, opened 2024-10-11
cgame: Add some popumessages defaults

Vorschreibung:


mittermichal:
why filtering out pickups and team joins by default?

Aciz:
I really don't agree with doing any of these by default.

Vorschreibung:
The nays have it.

mittermichal:
we still love you

Vorschreibung:
I appreciate that, but you could send a bit of that love as well to players who have to suffer the current default config who have excessively long text messages fill their screen literally up to where their crosshair is :')

In case people have forgotten how the game actually looks like by default:
image

mittermichal:
popupmessage component width could be lowered instead

Vorschreibung:
> popupmessage component width could be lowered instead

7 messages is already ridiculously low when simply standing in front of an ammo dispenser will fill them up via `Picked up ammo` messages within mere seconds

reducing the width would just cause linebreaks

please don't take this the wrong way, but I've already tried quite a number of things - there are no trivial solutions to this and the status quo is pretty much completely nonsense and the default experience suffers greatly under it

PR #2875, opened 2024-10-12
mod: fix uiMenuCommand_t enum order back to normal

Aciz:
Years ago there were new menu commands added at the middle of the `uiMenuCommand_t` enum. This caused issues with other mods running on ETL, as the enum is also compiled into the engine, which sends auto-update popups via `VM_Call` to UI. Because `UIMENU_WM_AUTOUPDATE` in legacys `bg_public.h` was different from what the menu order is in `etmain` and derivative mods, this would break the auto-update popups on other mods, and potentially open completely wrong menus in the UI when the game is started and updates are checked (for example, ETJump opens a completely unrelated popup menu when the game is started due to `UIMENU_WM_AUTOUPDATE` having value `16` instead of `14` in engine).

This restores the original enum order for entries 0-15, and moves the legacy-exclusive enum entries to the end of the enum, like they should have been inserted to begin with.


src/ui/ui_main.c
```diff
@@ -95,6 +95,18 @@ void UI_MouseEvent(int dx, int dy);
...
+	// UIMENU_WM_AUTOUPDATE is the only menu we need to adjust,
+	// as it's used in engine. Older versions have it set to 16.
+	if (uiInfo.etLegacyClient > 282010335 || menutype != 16)
+	{
+		return menutype;
+	}
+
+	return UIMENU_WM_AUTOUPDATE;
+}
+
```
Aciz, line 109: To further elaborate on this: the **only** menu command that is ever called directly from the engine, that was affected by the re-ordering was `UIMENU_WM_AUTOUPDATE`. There are other calls which set some of the enum values, but they are all below value `12`, which has always had the correct enum order. All other menu commands are sent from the mod code, so they will be correct even though the order has now been changed.

None of these menus are also ever visible during demo playback, therefore we don't need to worry about backwards compatibility with old demos.

PR #2878, opened 2024-10-12
cgame: Add cg_muzzleFlashDlight CVAR

Vorschreibung:
```
0            - force muzzle flash dynamic light off
1            - force muzzle flash dynamic light on
2            - force muzzle flash dynamic light on only for others not yourself
3  (default) - depends on cg_muzzleFlash

Also refactor 'CG_MG42EFX'.
```

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2766,6 +2766,7 @@ extern vmCvar_t cg_tracerSpeed;
 extern vmCvar_t cg_autoswitch;
 extern vmCvar_t cg_fov;
 extern vmCvar_t cg_muzzleFlash;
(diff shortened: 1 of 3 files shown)


Vorschreibung:
**IMPORTANT** 
Only merge after
https://github.com/etlegacy/etlegacy/pull/2874
has been merged!
**IMPORTANT**

Aranud:
It work.

I would recommand to use 0 as default value and limit the usage of negative value for cvar, as player are more use to custom with value between 0 and infinite.

Vorschreibung:
@Aranud fixed, 0 is the new default (previously -1).

PR #2879, opened 2024-10-13
sys: Add --strict-loading cmd line option

Vorschreibung:
By default loading errors (GameDLL etc.) do not cause the engine to
abort, but instead the engine prints a warning and fallbacks to loading
from SEARCHPATH2.

This however means when we do local development, it is hard to catch DLL
loading failures of our modified game DLL, as the engine likely falls
back to loading the game DLL from our `fs_homepath` instead (which
likely already exists) - which can cause a lot of confusion and is hard
to notice in practice.

This commit adds a '--strict-loading' cmd line option to disable this
implicit fallback behavior and make the engine abort printing the
specific loading error whenever a DLL loading failure happens.

### DIFF
--- a/src/qcommon/q_shared.h
+++ b/src/qcommon/q_shared.h
@@ -466,6 +466,8 @@ typedef int clipHandle_t;
 
 #define DEFAULT_NAME "ETLegacyPlayer"
 
+extern char* GlobalGameTitle;
+
 /**
  * @enum messageStatus_t
  * @brief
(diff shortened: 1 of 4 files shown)


Vorschreibung:
**IMPORTANT**
Only merge this **after** https://github.com/etlegacy/etlegacy/pull/2847 has been merged!
**IMPORTANT**

Vorschreibung:
I disagree with SonarCloud

PR #2882, opened 2024-10-14
etmain: Tweak 'etl_plane/bomb.md3'

Vorschreibung:
Fixed UV Wrap, added pseudo-shadows to texture, some random details,
de-saturated to fit the game colors more.


Vorschreibung:
image

Aranud:
The angle is managed in code https://github.com/etlegacy/etlegacy/blob/be98b3adf916524573effe0ec5e50ef05ac4a99d/src/game/g_weapon.c#L2746-L2750.

I would prefer keep the current model angle and use a custom angle in code, so we can reuse the model for other purpose without dealing with the angle, as we don't know the exact value.

jackeri:
You should be able to scale the model as well in code..?

Aranud:
> You should be able to scale the model as well in code..?

I would agree the base model is big ... so I'm not against changing it. 
Still, it's indeed preferable to change it from code if the model is already reused somewhere else ... which is not the case.

Vorschreibung:
Removed the rotation again. Needs both a Z offset and rotation in code (as the bombs currently fall through the ground and detonate below the ground).

Fixed some of the UV Wrap and tweaked the texture (pseudo lightning, desaturated, some arbitrary details).

Vorschreibung:
before
image

after
image

Aranud:
The bomb look nicer, good job.

But you haven't adjust the angle it seem ?

Vorschreibung:
> The bomb look nicer, good job.
> 
> But you haven't adjust the angle it seem ?

I'll put this into Draft and will add commit to translate and angle the model in code.

Vorschreibung:
@Aranud Ready for review again, slightly changed the existing rotation in code.

Attempts to translate the model or modify it's bounding box via `mins` / `maxs` didn't work, which is fine, they just briefly dip into the ground before exploding atm. but no one will notice unless `timescale 0.1` anyhow probably. Something for later.

AFAIS they don't have any bounding box atm. and there's just a trace if the origin dips into the ground (ofc. the bomb model is larger than it's origin).

---

I've now also shifted the Hue such that it matches with the Airplanes better that drop them.

PR #2883, opened 2024-10-14
cgame: Add 'cg_autoCmd' CVAR

Vorschreibung:
Allows setting a command that automatically gets executed when first snapshot was received.

Also executes for each `map_restart`.


Vorschreibung:
Allows team/class selection via command line args, e.g.:
```
$ ./etl.x86_64 +set cg_autoCmd 'team Axis 4 33 3;cg_thirdPerson 1;cg_thirdPersonAngle 180' +devmap battery
```

mittermichal:
I wonder and will test if this can be used also in demo playback

mittermichal:
I used these `wait` hacks as workaround with etpro demos: https://github.com/mittermichal/greatshot-render-worker/blob/main/tasks.py#L127-L137

PR #2884, opened 2024-10-14
etmain: Apply Editorconfig to text files

Vorschreibung:


Vorschreibung:
btw. to verify this, you can tell GH to hide whitespace changes and then there should be no remaining diff with this commit
image

PR #2892, opened 2024-10-15
etmain: Tweak airplane visuals

Vorschreibung:
Modified textures with added pseudo-shadows and some grit.

Also removed the transparency effects, as the models caused transparency overlap between their own faces and there doesn't seem to be a need for it atm. - it however should be fixed and re-enabled at a later point, but this is not possible without recasting the model geom and doing the UV wrap again.

Until then, this is considered an improvement.


Aranud:
Removing the transparency effects made the plane not affected by the fade in / out.

Vorschreibung:
> Removing the transparency effects made the plane not affected by the fade in / out.

After trying more, we could not have both solid airplanes and a fade out effect - so we opt to forego the fade out effect for now.

I've documented the issues and they can be attempted to be solved in the future, but this should be an improvement for now.

PR #2895, opened 2024-10-15
cgame: Limbo defaults to SMG if Soldier can

Vorschreibung:
Currently the SMG will never be autoselected when switching to the Soldier class through Limbo menu.

The reason for that is that 'SetDefaultWeapon(SECONDARY_SLOT)' is called once the Class is changed and never thereafter - however, when the Class is changed, the default primary weapon is set to the SMG and the consequent 'SetDefaultWeapon(SECONDARY_SLOT)' call will reject the SMG, as it always rejects weapons that are already set in the primary slot.

To circumvent this we now track if a user ever switched the weapon panel to the secondary panel (which will reset if they change either class or team) and if they never have opened it, we call
'SetDefaultWeapon(SECONDARY_SLOT) **again** right before we actually place the `team` command in 'CG_LimboPanel_SendSetupMsg'.

The end effect is that just switching to Soldier and selecting a heavyweapon will now always prioritize the SMG as long as the SMG perk is available.

### DIFF
--- a/src/cgame/cg_limbopanel.c
+++ b/src/cgame/cg_limbopanel.c
@@ -1395,6 +1395,21 @@ void CG_LimboPanel_SendSetupMsg(qboolean forceteam)
 		team = cgs.clientinfo[cg.clientNum].team;
 	}
 
+	// if the secondary weapon wasn't manually set, we set the default weapon
+	// for it once again - this is specifically for the edge case that a soldier
+	// with the smg soldier skill switches to a heavywep - for that case, once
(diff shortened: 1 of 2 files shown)


Vorschreibung:
Merely switching to the secondary weapon slot and not changing it will still be considered it being manually set.

This is by design, in my view if you had a chance to see what the secondary weapon has been set to, you consent to that decision and thus "manually" approve.

Otherwise it could add confusion when you see double pistol, don't change it, confirm and end up with an SMG.

Issue #2555, opened 2024-04-29
Investigate using `sv_fps 40` as a default


Calinou:
A lot of effort has been put into fixing `sv_fps`-related issues, so much that most competitive gameplay is now done with `sv_fps 40`. Unfortunately, most casual players won't get to experience this increased tickrate, despite it being beneficial at all skill levels.

There are many advantages to running a higher tickrate on public servers or local matches:

- Other players' movement follow more closely the player's actual position. Even though there's lag compensation and animation-induced weapon firing delay has been resolved recently (thanks again!), there's still a noticeable delay between clicking the left mouse button and hitting a player due to the tick rate. This also reduces peeker's advantage by design.
- Crosshair spread is updated more smoothly. Right now, it's not interpolated, so the 20 Hz update rate is rather obvious.
  - The crosshair spread display could be interpolated on the client side, but this would also increase the perceived delay between the spread update (which isn't predicted on the client-side) and the displayed value. This might be more viable with 40 Hz tickrate, as the delay will be less problematic.
- Being pushed around by other players or moving vehicles is smoother (or at least less frustrating :stuck_out_tongue:).
- Local bot matches feel smoother, with less delay when firing weapons.
- Spectating a player looks more accurate to their true actions.

In terms of CPU/bandwidth requirements, I don't know if it's been tinkered with recently, but I'd expect most public servers to be able to run `sv_fps 40` without trouble.

It might be challenging for the largest active servers like TeamMuppet though (with their 48-player + 16 spectator limit), so perhaps the tickrate should be decided depending on player count (with `sv_fps` being adjusted every few minutes to account for the current player count). Adjusting `sv_fps` during gameplay will cause a small stutter, so we should ensure this isn't done during the heat of the action (e.g. in the last 2 minutes of a match). It should preferably done when many players are wounded or dead, so that fewer players will notice the tickrate change kicking in.

For instance, this could be controlled with the following cvars:

```c
// Default FPS (used on server startup and restored after a long enough duration below the threshold).
sv_fps 40

// Maximum amount of players where the default FPS value is used.
// This value would support up to 12v12 matches (ignoring spectators).
sv_fpsMaxPlayers 24

// `sv_fps` fallback value when above player threshold.
sv_fpsAboveMaxPlayers 20
```

[...]

PR #2897, opened 2024-10-16
game: Default sv_fps to 40

Vorschreibung:
See https://github.com/etlegacy/etlegacy/issues/2555


mittermichal:
I think there are more places in repo where sv_fps is set to 20

Aciz:
> I think there are more places in repo where sv_fps is set to 20

You want to do it this way so that only legacy mod sets it to 40 by default. If you change the cvar default value in engine, you will break all other mods out of box which have no `sv_fps` fixes.

Edit: if this is adjusted, then @Vorschreibung you'd likely want to set `snaps 40` by default too in cgame.

Vorschreibung:
As the upcoming Release will be the first Release with the Flamethrower `sv_fps` fix - I guess it makes sense to wait another Release and to see that everything continues to work on 20+ `sv_fps` as intended before making it default.

x0rnn:
Shouldn't the server configs also get changed? Different mods use different server configs in their respective mod folder.

Also, g_maxWarp should be changed to 8 with sv_fps 40.

PR #2898, opened 2024-10-16
cgame: Improve noAmmoSound on failed weap switch

Vorschreibung:
Covers some previously missing cases (grenades, syringe) and adds a CVAR to disable it.

### DIFF
--- a/src/cgame/cg_event.c
+++ b/src/cgame/cg_event.c
@@ -2290,7 +2290,7 @@ void CG_EntityEvent(centity_t *cent, vec3_t position)
 
 		if (es->number == cg.snap->ps.clientNum)
 		{
-			if ((cg_noAmmoAutoSwitch.integer > 0 && !CG_WeaponSelectable(cg.weaponSelect))
+			if ((cg_noAmmoAutoSwitch.integer > 0 && !CG_WeaponSelectable(cg.weaponSelect, qfalse))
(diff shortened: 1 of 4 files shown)


Aranud:
I think it conflict with this https://github.com/etlegacy/etlegacy/blob/c29a150d0f5e1d4ecd09660dcbfffc9b34816617/src/cgame/cg_event.c#L2286-L2289.

No Ammo sound are played on event so enemy player can also hear it.

If you do the check only for the player it self, a sound will be missing.

Vorschreibung:
> I think it conflict with this
> 
> https://github.com/etlegacy/etlegacy/blob/c29a150d0f5e1d4ecd09660dcbfffc9b34816617/src/cgame/cg_event.c#L2286-L2289
> 
> .
> No Ammo sound are played on event so enemy player can also hear it.
> 
> If you do the check only for the player it self, a sound will be missing.

Good catch!

Vorschreibung:
@Aranud Now doesn't mess with the `noAmmoSound` of individual weapons and removes playing it in some cases where it would unnecessarily play it in the previous commit of this PR.

e.g. the moment you fire your last grenade, it played in the previous commit of this PR right before switching off to another weapon - it now switches off to another weapon without playing the sound and will only play once your try to switch to your empty grenades yourself

Ready for re-review

Vorschreibung:
Resolved conflict, ready for review again.

PR #2900, opened 2024-10-18
cgame: MG42/Browning follows viewangle pitch

Vorschreibung:
MG42 had notoriously bad angles when aiming downwards such that the gun model would obscure the crosshair's target completely.

As such, it was very common to disable gun drawing altogether when playing with a heavy mg via `cg_drawGun 0`.

Instead of forcing players to go through this in order to fully enjoy heavy mgs, the `*_SET` weapon types now all change their pitch depending on viewangle's pitch, as other weapons like mounted mgs or mortars.


Vorschreibung:
before/after when deploying on a slope
image

Vorschreibung:
Ready to review.

PR #2901, opened 2024-10-18
client: Increase 'con_openspeed' default from 3 → 5

Vorschreibung:
A very good default compromise to get both a fast console open and still a smooth transition.


Vorschreibung:
Ready for Review

PR #2902, opened 2024-10-18
cgame: Move drawing airplane into new 'cg_drawAirplanes'

Vorschreibung:
Drawing airplanes used to be part of `cg_visualEffects`, which however has been forced off on many servers due to how visuall obtrusive and fps dependent/uneven especially smoke effects are.

Airplanes were added behind `cg_visualEffects` and got thus practically disabled by default.

However, as they are unobtrusive, practically free from a performance stnadpoint and even provide some historic info on where airstrikes previously happened, they are now split into a separate CVAR so that servers can decide to keep them while still getting rid of the particle effects of `cg_visualEffects`.


Vorschreibung:
Ready for Review

PR #2903, opened 2024-10-19
cgame: Debug draw Artillery on sv_cheats

Vorschreibung:
When using binocs as fops playing on an `sv_cheats 1` server, draw a sprite for where playerstate looks showing if artillery can be called in at the spot or not.


Vorschreibung:
image
image

Vorschreibung:
Ready for Review

PR #2904, opened 2024-10-19
cgame: Add `cg_gunFov` and ETL dynamic gun fov

Vorschreibung:
Adds `cg_gunFov` to customize the "gun fov" and introduce `cg_gunFov 0` featuring a new per-gun, per-cg_fov dynamic mode that targets visibility, consistency and glitch-freeness for at least any `cg_fov` between 90-120, which is being used by default.

Consequently, some examples:

```
/cg_gunFov 0   # dynamic mode (default)
/cg_gunFov 75  # keeps the guns as they were in 2.82.1
/cg_gunFov 90  # keeps the guns as they were in Vanilla ET (previously not possible)
```


Vorschreibung:
Ready for review :+1: 

---

To test, apply this PR and check out guns with `cg_fov 90` and `cg_fov 120` respectively.
They should be familiar to the current look of the guns, with good visibility, have consistent perspective and do not show any visual glitches.

Individual weapons can also be compared between `cg_gunFov 75` (2.82.1, which has the current regression) and `cg_gunFov 0` - like the Panzerfaust, the Pistols or the Flamethrower or the Allied Grenade, which were all accidentally cut off significantly previously.

Vorschreibung:
Some before/after for easy and quick comparisons (all `cg_fov 100`):
image
image

src/cgame/cg_weapons.c
```diff
@@ -2351,6 +2379,257 @@ void CG_AddPlayerWeapon(refEntity_t *parent, playerState_t *ps, centity_t *cent)
...
+		CG_ApplyViewWeaponShift(hand,
+		                        (vec3_t) { -3.5, 2.0, 0.0 },
+		                        (vec3_t) { -7.8, 0.0, 6.0 });
+		break;
+	default:
+		// default to `cg_gunFov 75` - which was the universal default in 8.82.1
+		VectorMA(hand->origin, -3.0, up, hand->origin);
+		break;
+	}
+}
```
Aranud: This should be move in .weapon file as new field.

Aranud (changes requested):

src/cgame/cg_weapons.c
Vorschreibung: > This should be move in .weapon file as new field.

Good point!

Vorschreibung:
Ready to re-review.

Vorschreibung:
Ready for Review again

Aranud (approved):

PR #2909, opened 2024-10-22
mod: Tweak Char Animations

Vorschreibung:
- Fit animations much more closely into action duration by reframing.
    
      All weapons used to share the same reload animation with the same
      animation duration, despite that different guns have different reload
      animations and despite 'Fast Reload' greatly reducing the reload
      speed.
    
      Separate reload animations were custom fitted for the different guns
      to actually match their reload duration.
    
    - Random jerking of player arms was reduced while firing (especially
      crouch firing, but also jumping). - This was previously accidentally
      caused by multiple firing animations being accidentally overlayed onto
      each other.
    
    - Allow animation changes while jumping.
    
      Using binocs binocs mid-air, changing weapons mid-air or scoping with
      a sniper rifle mid-air were all examples where no animation change
      would happen at all. This is now fixed.
    
    Along with some sugar:
    
    - Add an animation for deploying/retract MG42 tripod.
    
    - Crashlanding now plays a short but sweet leg break leg animation.
    
    - Using doors or buttons plays a short but sweet low priority 'activate'
      animation.


Aranud:
Conflict !

Vorschreibung:
> Conflict !

Conflict resolved :metal:

PR #2917, opened 2024-10-24
game: fix knockback falloff computation

Aranud:


ryzyk-krzysiek:
Moving back knockback code introduces other issues:

1. `ps.velocity` is modified by it and it is used for head hitbox creation in `IsHeadShot` which happens later, meaning this bug is back https://github.com/etlegacy/etlegacy/pull/1644 (it's harder to reproduce than before because of the other fix, but it's possible and means that you register a headshot on first trace, and then do trace again and it is no longer headshot).
2. `pm_flags` is modified before `G_ReAdjustSingleClientPosition` meaning it will be overwritten by it.

Aranud:
Thanks you. 
Fixed in e3399d773b5c10903c542c32131bdea07a67df16

PR #2921, opened 2024-10-27
cgame: Default HUD `demotext` to 50% transparency

Vorschreibung:
Slightly reduce the attention being drawn to the component, as it's already fairly largely sized, with a bright red signal.


Vorschreibung:
image

PR #2922, opened 2024-10-27
cgame: Slightly reduce HUD 'popmessages' default size

Vorschreibung:
From 100.0f to 89.7f adds 1 additional line of history (from 7 to 8) and slightly cuts down the horizontal space used, while still feeling wholly familiar to oldtimers.


Vorschreibung:
image

PR #2924, opened 2024-10-27
game: Tweak Spectator aim-following

Vorschreibung:
Currently, as a spectator, issueing an `ATTACK` cmd while hovering over a player will move the view to that particular player.

However, if the trace (which determines if a spectator is currently hovering over another player) misses, the "next player" will be swapped to instead - as would issueing an `ATTACK` cmd without hovering any player by default.

This constitutes an issue, whereby a spectator who intends to follow a particular player, misses his aim and instead to switching to the intended player, switches to a seemingly random player.

To give an explicit solution for this intention, holding down the `SPRINT` key while issueing an `ATTACK` cmd will now never move the view to the "next player" - so missing a player while doing so will simply do nothing in this instance, giving the spectator the opportunity for another attempt.

Previously the `SPRINT` & `ATTACK` combination was used to differentiate cycling between non-bot and bot players, this combination is hereby moved to `WALKING` & `ATTACK` instead.

Furthermore, the way by which we determine which player is being aimed is improved in the following way:

Instead of making a single "fat" trace (with a +/- 64 unit beam), we make 4 separate traces for each `ATTACK` cmd while spectating.

The first one is a simple unsized hitscan, which checks if the spectator has directly hit the player - if not, increasing beam-widths are tried in increasing size to narrow down the potential target.

The maximum beam-width tried however has remained the same (+/- 64 units).

This commit therefore attempts to greatly increase the intuitiveness of using this particular feature.

### DIFF
--- a/src/game/g_active.c
+++ b/src/game/g_active.c
@@ -516,32 +516,53 @@ void G_TouchTriggers(gentity_t *ent)
  */
 qboolean G_SpectatorAttackFollow(gentity_t *ent)
 {
-	trace_t       tr;
-	vec3_t        forward, right, up;
-	vec3_t        start, end;
-	vec3_t        mins, maxs;
-	static vec3_t enlargeMins = { -64.0f, -64.0f, -48.0f };
-	static vec3_t enlargeMaxs = { 64.0f, 64.0f, 0.0f };
+	int     i;
+	trace_t tr;
+	vec3_t  forward;
+	vec3_t  mins, maxs;
+
+	static float enlargeMins[3] = { -4.0f, -4.0f, -3.0f };
+	static float enlargeMaxs[3] = { 4.0f, 4.0f, 0.0f };
+
+	vec3_t start, end;
 
 	if (!ent->client)
 	{
 		return qfalse;
 	}
 
-	AngleVectors(ent->client->ps.viewangles, forward, right, up);
 	VectorCopy(ent->client->ps.origin, start);
 	VectorMA(start, MAX_TRACE, forward, end);
 
-	// enlarge the hitboxes, so spectators can easily click on them..
-	VectorCopy(ent->r.mins, mins);
-	VectorCopy(ent->r.maxs, maxs);
-	VectorAdd(mins, enlargeMins, mins);
-	VectorAdd(maxs, enlargeMaxs, maxs);
 	// also put the start-point a bit forward, so we don't start the trace in solid..
 	VectorMA(start, 75.0f, forward, start);
 
-	// trap_Trace(&tr, start, mins, maxs, end, ent->client->ps.clientNum, CONTENTS_BODY | CONTENTS_CORPSE);
-	G_HistoricalTrace(ent, &tr, start, mins, maxs, end, ent->s.number, CONTENTS_BODY | CONTENTS_CORPSE);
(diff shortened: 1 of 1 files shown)


Vorschreibung:
@keithmss here you go

keithmss:
Thank you so much, I'm a big fan of this one!

src/game/g_active.c
```diff
@@ -517,38 +535,71 @@ void G_TouchTriggers(gentity_t *ent)
...
+		VectorCopy(ent->r.mins, mins);
+		VectorCopy(ent->r.maxs, maxs);
+		VectorAdd(mins, enlargeLargeMins, mins);
+		VectorAdd(maxs, enlargeLargeMaxs, maxs);
+
+		if (G_SpectatorAttackFollowAttempt(ent, &tr, forward, mins, maxs))
+		{
+			goto done;
+		}
+	}
```
Aranud: Use a loop and multiply by 4 the box bounds at each iteration.
The first iteration could start with BBox of 1, so it is icncluded in the loop
or
the first trace could start outside the loop, so it keep the non usage of BBox (less computation for the first trace)

Aranud (changes requested):

src/game/g_active.c
Vorschreibung: Done! Ready for re-review :metal:

src/game/g_active.c
```diff
@@ -516,32 +516,51 @@ void G_TouchTriggers(gentity_t *ent)
...
+		G_HistoricalTrace(ent, &tr, start,
+		                  (i == 0) ? NULL : mins,
+		                  (i == 0) ? NULL : maxs,
+		                  end, ent->s.number, CONTENTS_BODY | CONTENTS_CORPSE);
+
+		if ((&g_entities[tr.entityNum])->client != NULL)
+		{
+			break;
+		}
+	}
```
Aranud, line 566:
```float enlargeMins[3] = { -4.0f, -4.0f, -3.0f };
float enlargeMaxs[3] = { 4.0f, 4.0f, 0.0f };

vec3_t start, end;

VectorCopy(ent->client->ps.origin, start);
VectorMA(start, MAX_TRACE, forward, end);

// also put the start-point a bit forward, so we don't start the trace in solid..
VectorMA(start, 75.0f, forward, start);

// let's do 4 traces with increasing "beam width" to prioritize the player
// we are directly looking at, but also allow some more fuzzy angles with
// decreasing priority
for (i = 0; i < 4; ++i)
{
    if (i == 0)
    {
        G_HistoricalTrace(ent, &tr, start,
                          NULL,
                          NULL,
                          end, ent->s.number, CONTENTS_BODY | CONTENTS_CORPSE);
    }
    else
    {
        G_HistoricalTrace(ent, &tr, start,
                          mins,
                          maxs,
                          end, ent->s.number, CONTENTS_BODY | CONTENTS_CORPSE);

        VectorMA(ent->r.mins, 2, enlargeMins, mins);
        VectorMA(ent->r.maxs, 2, enlargeMaxs, maxs);
    }


    if ((&g_entities[tr.entityNum])->client != NULL)
    {
        break;
    }
}```

Aranud (changes requested):

PR #2927, opened 2024-10-28
cgame: Tweak HUD Editor noise generator

Vorschreibung:
- generate `popupmessages` items more consistently (used to be modulo to cg.time, which was unpredictable and fps dependent)

- for fireteam overlay, draw a box with an info that mentions to join a real fireteam instead


Vorschreibung:
Fireteam overlay warning when noise generator is active and player is not in a real fireteam.
image

PR #2928, opened 2024-10-28
cgame: Unique names when cloning via HUD Editor

Vorschreibung:
Previously cloning a HUD would just use the parent HUD name and suffix
with `%s_copy`, no matter if it existed prior or not, creating multiple
HUDs with the same name.

Instead of doing this, we now check if `%s_copy` is free, if it's not,
we try `%s_copy%d` >= 2 && < 100 until we find a free spot and use that
as a new name.

If no free spot can be found, we print an error and stop.

### DIFF
--- a/src/cgame/cg_hud_editor.c
+++ b/src/cgame/cg_hud_editor.c
@@ -747,6 +747,7 @@ qboolean CG_HudSave(int HUDToDuplicate, int HUDToDelete)
 
 	if (HUDToDuplicate >= 0)
 	{
+		int attemptedSpot;
 		int hudNumber;
 
 		if (hudData.count == MAXHUDS)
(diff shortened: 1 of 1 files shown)


src/cgame/cg_hud_editor.c
```diff
@@ -761,7 +762,39 @@ qboolean CG_HudSave(int HUDToDuplicate, int HUDToDelete)
 
 		CG_CloneHud(hud2, hud);
 
-		Q_strncpyz(hud2->name, va("%s_copy", hud->name), sizeof(hud2->name));
+		// determine a new name
+		for (attemptedSpot = 1; attemptedSpot < 100; attemptedSpot++)
```
Aciz: `MAXHUDS` is 32 so this doesn't really need to go this high (and better to use a constant here too rather than a magic number)

https://github.com/etlegacy/etlegacy/blob/1827e3a5618ee2b65b96aa84f32b841d2dac438e/src/cgame/cg_local.h#L4384

Vorschreibung: No magic number! :heart:

Vorschreibung:
Example usage

https://github.com/user-attachments/assets/691a2562-dc26-45a8-b892-92e412c53794

src/cgame/cg_hud_editor.c
Vorschreibung: Done

PR #2931, opened 2024-10-28
cgame: Fix HUD Editor and Limbo for demo playback

Vorschreibung:
Block demo/replay playback to open Limbo menu and allow HUD Editor to be closed via the default 'F7' hotkey.


Vorschreibung:
@mittermichal Here you go

PR #2934, opened 2024-10-29
cgame: Show damage skin on player heads

Vorschreibung:


src/cgame/cg_players.c
```diff
@@ -2999,6 +3005,85 @@ void CG_Player(centity_t *cent)
...
+				else if (healthPct <= 0.50)
+				{
+					damagedState = 1;
+				}
+				else if (healthPct <= 0.80)
+				{
+					damagedState = 0;
+				}
+			}
+		}
```
Aranud, line 3033: It is inconcistent with the value set for client https://github.com/etlegacy/etlegacy/blob/17ea8b747152254852b09127ac61006bdae14f2e/src/cgame/cg_draw_hud.c#L604-L619

src/cgame/cg_players.c
```diff
@@ -2999,6 +3005,85 @@ void CG_Player(centity_t *cent)
...
+
+		}
+
+		if (damagedState >= 0)
+		{
+			head.customShader = 0;
+			head.customSkin   = cgs.media.hudDamagedStates[damagedState];
+			trap_R_AddRefEntityToScene(&head);
+		}
+	}
```
Aranud: Store this is MOD Table https://github.com/etlegacy/etlegacy/blob/17ea8b747152254852b09127ac61006bdae14f2e/src/game/bg_misc.c#L248

Aranud (changes requested):

src/cgame/cg_players.c
Vorschreibung: > It is inconcistent with the value set for client

Good eye! That's on purpose tho after trying them out a lot, I think that my values are better - they are only visible for others as well, so the difference won't be noted.

I also used `cgs.media.hudDamagedStates[3]` for a while, but have found that it looks off and too intrusive - especially because the neck and the rest of the body won't ever be blooded, but only the face.

I think my proposal blends in much better.

Let me know if this suffices for you, or if you want to still change it.

src/cgame/cg_players.c
Vorschreibung: Good point! Done

PR #2938, opened 2024-10-29
cgame: Fix Garand/M7 RG attach animation quirk

Vorschreibung:
Blend out the right hand of garand/m7 while attaching the rifle grenade to fix a shipped animation quirk where the hand would reappear on the bottom right at the last parts of the animation.


Vorschreibung:
previously popped in the right hand at the end for the remaineder of the animation and then magically pop out again when transitioning to IDLE weaponstate
image

clean now:

https://github.com/user-attachments/assets/96a8dcb6-517f-4abd-82c2-1a93b370712a

Vorschreibung:
@u1ps you have raised this before, now fixed

PR #2940, opened 2024-10-30
mod: Add new muzzle flash sizes

Vorschreibung:
Removes 'cg_muzzleFlashScale', which was introduced in a previous snapshot (not in a release).

Adds 'cg_muzzleFlashOld', which defaults to '0'.

When not set, it scales muzzle flashes via first-person and third-person values set in `.weap` files.

The goal of new muzzle flashes is to have them not heavily obscure the player view such that players are incentivized to switch them off entirely via 'cg_muzzleFlash 0'.


Vorschreibung:
Ready for review.

PR #2941, opened 2024-10-30
cgame: Add 'Toggle Help' Button to HUD Editor

Vorschreibung:
Currently the Help Text always gets shown when opening the HUD Editor
for the first time since the game was started.

This makes sure that users take note of it, however it also becomes
cumbersome after you know about it's existance.

Instead of showing the help text always at the start, this commit adds a
`Toggle Help` Button and gives it a special color to make it easily
noticeable.

### DIFF
--- a/etmain/ui/menudef.h
+++ b/etmain/ui/menudef.h
@@ -439,4 +439,6 @@
 #define SCREEN_WIDTH            640
 #define SCREEN_HEIGHT           480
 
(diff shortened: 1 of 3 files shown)


Vorschreibung:
https://github.com/user-attachments/assets/a33f699f-21d3-4b6c-8cdc-f3f0257c6e85

Issue #2942, opened 2024-10-30
Crosshair is blurry / transparent since 2.82.0


S1RIUS1:
Crosshair is blurry / transparent since 2.82.0

Tried r_scalesvg 1 or 2
Tried cg_useCvarCrosshair 1
Tried with a fresh profile.

Files included show the difference between 2.82.1 and 2.81.1. 
Crosshairsize is 38, game is running at 1080p on 1080p screen.

Rendering the game at 1440p then downscaling it to 1080p _fix_ the issue (with nvidia SuperRes)

![2811](https://github.com/user-attachments/assets/ab992cdc-b90c-4363-807e-8a2206c76fc0)
![2821](https://github.com/user-attachments/assets/b0525da0-d44f-4d19-ac72-9ee59087a19c)

Vorschreibung:
Thanks for reporting! :heart: We'll look into it!

Vorschreibung:
@Aranud AFAIS this can be closed since we now have `cg_crosshairSVG`.

Aciz:
I wouldn't necessarily consider this fixed, the cvar is a duct tape solution to the underlying issue.

Aranud:
The duct tape will stay until crosshair editor will be in. Would say, svg will be no longer a discussion for crosshair all

Vorschreibung:
> The duct tape will stay until crosshair editor will be in. Would say, svg will be no longer a discussion for crosshair all

I agree.

PR #2943, opened 2024-10-30
cgame: Return to use `.tga` crosshairs, instead of `.svg`

Vorschreibung:
As we've received reports of issues with the SVG crosshairs (#2942),
this change stops using `.svg` crosshairs by default .

Also adds a 'cg_crosshairSVG' CVAR (default 0), which allows opting in
to the `.svg` crosshairs by setting it to 1.

As currently paths in the renderer always prioritize `.svg` files over
`.tga` files, no matter if we specify a file extension explicitly or
not, we have to give the `.svg` crosshairs unique file names when
stripping the file extension.


src/cgame/cg_main.c
```diff
@@ -641,6 +642,7 @@ static cvarTable_t cvarTable[] =
 	{ &cg_drawSpawnpoints,          "cg_drawSpawnpoints",          "0",           CVAR_ARCHIVE,                 0 },
 
 	{ &cg_useCvarCrosshair,         "cg_useCvarCrosshair",         "1",           CVAR_ARCHIVE,                 0 },
+	{ &cg_crosshairSVG,             "cg_crosshairSVG",             "0",           CVAR_ARCHIVE,                 0 },
```
Aciz: This needs to be a `CVAR_LATCH` as the shaders need to be re-registered to switch between TGA/SVG.

Vorschreibung: Thanks! Done

PR #2944, opened 2024-10-30
mod: add entity bit stream ET_EBS_SHOUTCAST

ryzyk-krzysiek:
- add server extension trap_SnapshotCallbackExt_Legacy for new systemcalls G_SNAPSHOT_CALLBACK_EXT and GAME_SNAPSHOT_CALLBACK_EXT.
  Without new snapshot callback it is impossible to send entities to select few clients.
- add new entity ET_EBS_SHOUTCAST replacing ET_LANDMINESPOT_HINT, enabled only if above engine extension exists and on stopwatch gametype.
  Updates only if there are present shoutcasters on server. Sends health, ammoclip, ammo, weapon and powerups. Sent only to shoutcasters.
- update shoutcast overlay with new data.
- add net_overhead_print and net_overhead_clear server commands to help evaluating network overhead, debug builds only.

Compared to previous version from a year ago:
- clients are always saved into same slots, to better utilize delta encoding.
- changed the way ebs is handled on client and server and thus all entityState fields are writeable beside eType (from 644 available int bits to 758 out of 766 total)

https://www.youtube.com/watch?v=gC4SchxcwFk

### DIFF
--- a/src/cgame/cg_ents.c
+++ b/src/cgame/cg_ents.c
@@ -2578,8 +2578,8 @@ void CG_EBS_Shoutcast(centity_t *cent)
 		ci        = &cgs.clientinfo[clientNum];
 
 		ci->health        = EBS_ReadBitsWithSign(&ebs, 9);
-		ci->ammoclip      = EBS_ReadBits(&ebs, 8);
-		ci->ammo          = EBS_ReadBits(&ebs, 8);
+		ci->ammoclip      = EBS_ReadBits(&ebs, 10);
+		ci->ammo          = EBS_ReadBits(&ebs, 10);
 		ci->currentWeapon = EBS_ReadBits(&ebs, 6);
 		ci->powerups      = EBS_ReadBits(&ebs, 16);
 	}
--- a/src/cgame/cg_ents.c
+++ b/src/cgame/cg_ents.c
@@ -34,6 +34,7 @@
  */
 
 #include "cg_local.h"
+#include "../game/bg_ebs.h"
 
 /**
  * @brief Modifies the entities position and axis by the given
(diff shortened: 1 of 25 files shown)


mittermichal:
awesome! adding player locations and stats (k,d,dmg,gibs) should be possible too with this right?

ryzyk-krzysiek:
> awesome! adding player locations and stats (k,d,dmg,gibs) should be possible too with this right?

Possible yes, worth it? No. Locations I omitted due to much higher bandwidth used but maybe it can be added but with better way. For stats I see no reason to add at all to this, you won't fit it into the overlay anyway, the ammo barely fits (it can overlay with powerups in an edge case already)

mittermichal:
> Locations I omitted due to much higher bandwidth used but maybe it can be added but with better way.

I see, not worth effort probably

you see these stats if you spectate someone but they are only refreshed in certain interval, but IMO it would be better if it was showed like XP gain, so that you easily see recent damage given.

mittermichal:
running this branch with ETLEGACY_DEBUG enabled on et.greatshot.xyz tv servers, but without mod changes.
Tried only tv connecting to empty server and connecting client to tv server.
```
net_overhead_print
 2702700 =========================
```

ryzyk-krzysiek:
> > Locations I omitted due to much higher bandwidth used but maybe it can be added but with better way.
> 
> I see, not worth effort probably
> 
> you see these stats if you spectate someone but they are only refreshed in certain interval, but IMO it would be better if it was showed like XP gain, so that you easily see recent damage given.

Whole stats feature is very lazy implementation by reusing `+stats` (default `alt` keybind) which sends a lot of stats that are irrelevant, but simply increasing the update rate would be very inefficient, it would require redoing it in another way so it is efficient (feature just for shoutcaster) and preferably backward compatible.

> running this branch with ETLEGACY_DEBUG enabled on et.greatshot.xyz tv servers, but without mod changes. Tried only tv connecting to empty server and connecting client to tv server.
> 
> ```
> net_overhead_print
>  2702700 =========================
> ```

You need to add commands/entities to track and enable them by yourself. Tracking is disabled by default (`net_overhead.numSlices = 0;`):

https://github.com/etlegacy/etlegacy/blob/463fe7ad9d31a87fce4693e032c911e24c076efa/src/server/sv_snapshot.c#L1438-L1446

PR #2946, opened 2024-10-31
renderer: Remove cheat protection from `r_lodScale`

Vorschreibung:
Removing cheat protection from `r_lodScale` as it became the CVAR set by
changing 'Geometric Detail' in Settings via
https://github.com/etlegacy/etlegacy/commit/78d5775690cbe4b8c60c159143279f3b9c11d717

### DIFF
--- a/src/renderer/tr_init.c
+++ b/src/renderer/tr_init.c
@@ -1150,7 +1150,7 @@ void R_Register(void)
 	r_skipBackEnd = ri.Cvar_Get("r_skipBackEnd", "0", CVAR_CHEAT);
 
 	r_measureOverdraw = ri.Cvar_Get("r_measureOverdraw", "0", CVAR_CHEAT);
-	r_lodScale        = ri.Cvar_Get("r_lodscale", "5", CVAR_CHEAT);
+	r_lodScale        = ri.Cvar_Get("r_lodscale", "5", CVAR_ARCHIVE_ND | CVAR_LATCH);
 	r_noreFresh       = ri.Cvar_Get("r_norefresh", "0", CVAR_CHEAT);
 	r_drawEntities    = ri.Cvar_Get("r_drawentities", "1", CVAR_CHEAT);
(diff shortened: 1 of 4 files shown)


Aciz:
FWIW this has implications for other mods ofc. Same was done to `r_wolffog` without much consideration years ago, which I'm not sure was a great move. This can be accomplished for just legacy mod by using cgame proxy cvars, as raw `trap_Cvar_Set` will bypass `CVAR_CHEAT` flag. We have such system in ETJump for unlocking cheat restriction from several cvars.

src/renderer/tr_init.c
```diff
@@ -1139,7 +1139,7 @@ void R_Register(void)
 
 	r_noCurves    = ri.Cvar_Get("r_nocurves", "0", CVAR_CHEAT);
 	r_drawWorld   = ri.Cvar_Get("r_drawworld", "1", CVAR_CHEAT);
-	r_drawFoliage = ri.Cvar_Get("r_drawfoliage", "1", CVAR_CHEAT);
+	r_drawFoliage = ri.Cvar_Get("r_drawfoliage", "1", CVAR_ARCHIVE_ND | CVAR_LATCH);
```
Aciz: `r_drawFoliage` in vanilla renderer does not require `CVAR_LATCH`

Edit: only R2 apparently requires `CVAR_LATCH`, others do not.

Vorschreibung:
> FWIW this has implications for other mods ofc. Same was done to `r_wolffog` without much consideration years ago, which I'm not sure was a great move. This can be accomplished for just legacy mod by using cgame proxy cvars, as raw `trap_Cvar_Set` will bypass `CVAR_CHEAT` flag. We have such system in ETJump for unlocking cheat restriction from several cvars.

Thanks for the input!

I don't consider changing this worth it then, at least not in this PR and have removed the change to `r_wolffog` (it can be done separately later if needed) - this PR is now only about `r_lodScale`.

Vorschreibung:
Ready for review again.

Issue #2333, opened 2023-06-30
Log engineer repairs


x0rnn:
Log engineer repairs with G_LogPrintf into the server log.

`G_LogPrintf("Repair: %d %s\n", ent->client->ps.clientNum, repaired_objective_name);`

Could maybe even add objective destructions, "Objective_Destroyed: %d %s" for engineer dynamites and covert satchels. Edit: to simplify and highlight only important destructions, log only engineer dynamites of objectives in TOIs.

It would be nice to have this information in the log to be able to parse it and calculate statistics and such and as far as I know these stats aren't being saved right now at all?

x0rnn:
It seems there already is a "Repair:" line being logged, but only for MG repairs for some reason. Could it be changed to log all repair events instead?

Vorschreibung:
Abusing ETL log lines as an API for the LUA side is a terrible and irresponsible idea, when adding a LUA Hook for this particular case is a matter of adding a function and creating a LUA bind for it, which can all be done within 5min to 1hr depending on your experience with the code and has been done dozens of times by one-time contributors in the past.

PR #2951, opened 2024-11-03
Log engineer repairs/constructions

x0rnn:
Addresses https://github.com/etlegacy/etlegacy/issues/2333, logs "Repair: clientNum" upon engineer repairs/constructions which currently logs it only for MG42/Browning repairs.

### DIFF
--- a/src/game/g_weapon.c
+++ b/src/game/g_weapon.c
@@ -1101,12 +1101,14 @@ static qboolean TryConstructing(gentity_t *ent, gentity_t *trigger)
 		{
 			// call script
 			G_Script_ScriptEvent(constructible, "built", "final");
+			G_LogPrintf("Repair: %d\n", (int)(ent - g_entities));
 		}
 		else
 		{
 			if (constructible->grenadeFired == constructible->count2)
 			{
 				G_Script_ScriptEvent(constructible, "built", "final");
+				G_LogPrintf("Repair: %d\n", (int)(ent - g_entities));
 			}
 			else
 			{


Vorschreibung:
Abusing ETL log lines as an API for the LUA side (as is described as the intention of this PR in https://github.com/etlegacy/etlegacy/issues/2333) is a terrible and irresponsible idea, when adding a LUA Hook for this particular case is a matter of adding a function and creating a LUA bind for it, which can all be done within 5min to 1hr depending on your experience with the code and has been done dozens of times by one-time contributors in the past.

Vorschreibung:
@jackeri @ryzyk-krzysiek as you have more experience with the code, please share your thoughts on this.

x0rnn:
Who mentioned anything about using it for lua or only for lua? As mentioned here, it's to be used with external stats parsers such as stats.hirntot.org and others as well, which need to read the etconsole log: https://github.com/etlegacy/etlegacy/pull/2949

Vorschreibung:
> Who mentioned anything about using it for lua or only for lua? As mentioned here, it's to be used with external stats parsers such as stats.hirntot.org and others as well, which need to read the econsole log: #2949

The same problem arises, because log lines are not an API and nothing in the code hints that you just elevated a log line into API status.

It additionally spams the logs with needless information that no human cares for, only your stats parser, making them increasingly verbose over time and thus harder to read.

Vorschreibung:
I'm at a loss why you can't just rely on the LUA API for your stat reporting, which is intended to provide a stable API.

Why are you mix and matching your server-side LUA scripts, with apparently another external program that parses the logs?

In case you do have an important use-case where you can't use the LUA API directly (which?) and you need to follow and parse lines from a file: it's trivial to write a LUA script that uses the LUA API itself and simply logs lines to a file itself.

x0rnn:
The current logging behaviour for engineer repairs is inconsistent - why only log MG repairs? This PR addresses that and is irrelevant to anything lua related.

I don't parse anything regarding repairs with lua, and don't intend to - don't assume what I do. In fact, having to use a lua to first hook this and then output to yet another log would be more troublesome than to just make the repair logging consistent in the first place and just be able to read the raw log file with whatever custom parser somebody is using. Again, why only log MG repairs?

Making a lua hook for this is irrelevant to this PR and the referenced issue, but if somebody wants it, sure, he can open an issue or push a PR about it.

PR #2952, opened 2024-11-04
client: Add 'CG_CVAR_SETDESCRIPTION' engine extension

Vorschreibung:
Allows setting descriptions for CVARs in cgame.

Also adds a 'trap_Cvar_SetDescription' shim to cgame and makes use of it by adding a description to `cg_autoCmd`.

### DIFF
--- a/src/cgame/cg_main.c
+++ b/src/cgame/cg_main.c
@@ -379,6 +379,14 @@ vmCvar_t cg_crosshairHealth;
 
 vmCvar_t cg_commandMapTime;
 
+// vmCvar_t Descriptions {{{
+#define cg_autoCmd_description \
+		"Allows setting a comand (or commands, ';'-seperated), that will automatically be executed when you first join a match or when the map restarts.\n" \
+		"\n" \
+		"Example:\n" \
+		"  /cg_autoCmd \"team r 2 23 3;say The sanctity of this place has been fouled.\""
+// {{{
+
 typedef struct
 {
 	vmCvar_t *vmCvar;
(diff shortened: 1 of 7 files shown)


Vorschreibung:
@Aciz @ryzyk-krzysiek Please let me know what you think about this and if I have possibly missed a way to set CVAR description without having to add an engine extension.

src/cgame/cg_main.c
```diff
@@ -378,280 +379,289 @@ vmCvar_t cg_crosshairHealth;
...
 vmCvar_t cg_commandMapTime;
 
+// vmCvar_t Descriptions {{{
+#define cg_autoCmd_description \
+		"Allows setting a comand (or commands, ';'-seperated), that will automatically be executed when you first join a match or when the map restarts.\n" \
+		"\n" \
+		"Example:\n" \
+		"  /cg_autoCmd \"team r 2 23 3;say The sanctity of this place has been fouled.\""
+// {{{
+
```
Aciz, line 389: Instead of this, I would add a struct + array, like `cvarTable_t`, which holds a pointer to a `vmCvar_t` + description for the cvar. You could then iterate this array in `CG_RegisterCvars`, where you currently check if the description is set, and register all the descriptions for cvars that way. I understand this is probably a temp solution that you've provided here, but this would be my suggestion for a more permanent solution, rather than adding hundreds of macros here, or writing out the descriptions to the cvartable itself.

Aciz (commented):
Looks okay, haven't tested but I don't see anything inherently wrong here.

ryzyk-krzysiek:
In my opinion adding extension like `trap_Cvar_RegisterAndDescribe` replacing `trap_Cvar_Register` would be better a solution. I think it makes more sense given how description is not something that will change past cvar registration. You could simply replace `trap_Cvar_Register` call with it in cgame and if extenstion is missing you simply call `trap_Cvar_Register` inside `trap_Cvar_RegisterAndDescribe`.

In engine you could either call `Cvar_Register` and then `Cvar_SetDescription_FromVmCvar`. Or extend `Cvar_Register` with description var and call `Cvar_SetDescription` inside, like so:
```diff @@ -1744,12 +1744,13 @@ void Cvar_SetDescription(cvar_t *cv, const char *varDescription)
  *
  * @param[in,out] vmCvar
  * @param[in] varName
  * @param[in] defaultValue
  * @param[in] flags
+ * @param[in] varDescription
  */
-void Cvar_Register(vmCvar_t *vmCvar, const char *varName, const char *defaultValue, cvarFlags_t flags)
+void Cvar_RegisterAndDescribe(vmCvar_t *vmCvar, const char *varName, const char *defaultValue, cvarFlags_t flags, const char *varDescription)
 {
 	cvar_t *cv;
 
 	// There is code in Cvar_Get to prevent CVAR_ROM cvars being changed by the
 	// user. In other words CVAR_ARCHIVE and CVAR_ROM are mutually exclusive
@@ -1837,10 +1838,24 @@ void Cvar_Register(vmCvar_t *vmCvar, const char *varName, const char *defaultVal
 		return;
 	}
 	vmCvar->handle            = cv - cvar_indexes;
 	vmCvar->modificationCount = -1;
 	Cvar_Update(vmCvar);
+	Cvar_SetDescription(cv, varDescription);
+}
+
+/**
+ * @brief Cvar_Register
+ *
+ * @param[in,out] vmCvar
+ * @param[in] varName
+ * @param[in] defaultValue
+ * @param[in] flags
+ */
+void Cvar_Register(vmCvar_t *vmCvar, const char *varName, const char *defaultValue, cvarFlags_t flags)
+{
+	Cvar_RegisterAndDescribe(vmCvar, varName, defaultValue, flags, NULL);
 }
 
 /**
  * @brief Updates an interpreted modules' version of a cvar
  * @param[in,out] vmCvar
```

PR #2960, opened 2024-11-07
mod: add cg_crosshairX/Y, cg_crosshairScaleX/Y for cvar crosshair

### LINKED ISSUE #2543: Can't set crosshair position
I used to use `cg_crosshairY "3"` in my config to move the crosshair position a few pixels down. It seems that the environment variable doesn't work anymore, and that I have to change it in the HUD. But when I do so, the crosshair position is reset after each map, even when I create a new HUD config and click "Save".

More discussion on Discord here: https://discord.com/channels/260750790203932672/260750790203932672/1230966199580364922

image

Aciz:
HUD editor is able to move the crosshair around and perform non-uniform scaling. This wasn't possible when using `cg_useCvarCrosshair`, as `cg_crosshairX/Y` were removed, and scaling options were never added. Crosshair customization should now have feature parity between HUD editor and cvars.

fixes #2543 
refs #1967

### DIFF
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -2973,22 +2973,56 @@ static void CG_EditHudComponentComplete(void)
 
 #define EDITCOMPONENT_CROSSHAIR_STRING "editcomponent crosshair"
 
-static void CG_CrosshairSize_f(void)
+static void CG_CrosshairSizePos_f(void)
 {
 	if (trap_Argc() > 1)
 	{
-		const char *token;
-
-		token = CG_Argv(1);
+		const char *token = CG_Argv(1);
 
 		if (Q_isanumber(token))
 		{
-			float size = Q_atof(token);
+			const float value = Q_atof(token);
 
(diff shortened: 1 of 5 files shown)


src/cgame/cg_consolecmds.c
```diff
@@ -2973,22 +2973,56 @@ static void CG_EditHudComponentComplete(void)
...
+				offsetY = value;
+			}
+			else if (!Q_stricmp(cmd, "cg_crosshairScaleX_f"))
+			{
+				scaleX = value;
+			}
+			else if (!Q_stricmp(cmd, "cg_crosshairScaleY_f"))
+			{
+				scaleY = value;
+			}
```
Aciz, line 3016: Rather than calling the same function here for multiple cvars, we could store all these in a struct and update values in that, and then just recompute the position with those values whenever they're modified, let me know if you want me to refactor this like that.

Initially I added separate functions for position/scale cvars, but that caused update collisions with `vid_restart`/map load, where the position/size updates would come from both this and the position/scale functions, which reset some of the values depending on which was called first, so I made this the SSOT function for computing size and position.

PR #2963, opened 2024-11-08
cgame: Allow midair `cg_weapaltMgAutoProne`

Vorschreibung:
Make airborne check dependent on `g_pronedelay`.


Vorschreibung:
`g_pronedelay` isn't currently sent in the CS, so this doesn't work

PR #2984, opened 2024-11-17
client: fix +strafe framerate dependency

Aciz:
Strafing around using `+strafe` was framerate dependant, and same mouse move with 125FPS would result in half the traveled distance with 250FPS. Normalize all movement to 125FPS, so `m_forward/side` behave the same regardless of framerate.

### DIFF
--- a/src/client/cl_input.c
+++ b/src/client/cl_input.c
@@ -954,6 +954,7 @@ void CL_MouseMove(usercmd_t *cmd)
 	float mx, my;
 	float accelSensitivity;
 	float rate;
+	float deltaStrafeSensitivity;
 
 	// allow mouse smoothing
 	if (m_filter->integer)
@@ -989,10 +990,13 @@ void CL_MouseMove(usercmd_t *cmd)
 		return;
 	}
 
+	// normalize '+strafe' sensitivity to 125FPS
+	deltaStrafeSensitivity = 0.008f * (1000.0f / (float)frame_msec);
+
 	// add mouse X/Y movement to cmd
 	if (kb[KB_STRAFE].active)
 	{
-		cmd->rightmove = ClampChar(cmd->rightmove + m_side->value * mx);
+		cmd->rightmove = ClampChar(cmd->rightmove + m_side->value * deltaStrafeSensitivity * mx);
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


chaq3l:
Why do you want to hard-normalize it to 125 Hz and not allow it to be normalized to a specific value as a server setting?

Aciz:
> Why do you want to hard-normalize it to 125 Hz and not allow it to be normalized to a specific value as a server setting?

That would make absolutely 0 sense? The sensitivity is already user-controllable with `m_forward` and `m_side`, this just normalizes the baseline so that e.g `m_forward 0.25` & `m_side 0.25` equal to same distance moved in game for a given physical mouse move, regardless of clients framerate. Without this, the higher your framerate, the slower your `+strafe` sensitivity is with same `m_forward` & `m_side` values.

chaq3l:
You wrote "same mouse move with 125FPS would result in half the traveled distance with 250FPS" and that what now you wrote is totally different. Should it be: same mouse move with 250FPS would result in half the traveled distance with 125FPS ??

Aciz:
Maybe I'm wording it poorly, but right now you basically have to double your `m_forward` and `m_side` sensivity if you go from 125 -> 250fps, in order for the same physical mouse move to result in same in-game move. This is what I'm addressing here.

chaq3l:
Right now I am just wondering how this change will affect gameplay because trick jump players prefer to use larger FPS rate with higher sensitivity to gain more speed while strafing. I will speak with some players to get better insight is this topic.

Aciz:
> Right now I am just wondering how this change will affect gameplay because trick jump players prefer to use larger FPS rate with higher sensitivity to gain more speed while strafing. I will speak with some players to get better insight is this topic.

Based off this comment I don't think you even know what `+strafe` does and what it's used for. I am in fact the main developer of [ETJump](https://github.com/etjump/etjump), `+strafe` is used pretty much exclusively by ETJump players to perform overbounce tricks, and having a precise, consistent mouse movement with `+strafe` is very important for that. The fact that it was fps meant that you'd have to keep adjusting your `m_forward/side` sensitivity based off what FPS you played at, which is just dumb, sensitivity should not be affected by framerate.

Outside of ETJump, I don't think anyone uses `+strafe` for anything, as it doesn't really make sense.

PR #2995, opened 2024-11-24
game: fix 'BG_PlayerTouchesObjective'

Liala1:
In v2.83, I discovered that in the map `password2`, the objective can pickuped before entering the password.
The current code picks up the objective when it is out of range, which is different from the v2.82.
fixed this.

ref: https://github.com/etlegacy/etlegacy/blob/0a24c70f84cf21940277ddabd736ce83535cb794/src/game/bg_misc.c#L2529

### DIFF
--- a/src/game/bg_misc.c
+++ b/src/game/bg_misc.c
@@ -2532,7 +2532,7 @@ qboolean BG_PlayerTouchesObjective(playerState_t *ps, entityState_t *item, int a
 	BG_EvaluateTrajectory(&item->pos, atTime, origin, qfalse, item->effect2Time);
 
 	// we are ignoring ducked differences here
-	return (ps->origin[0] - origin[0] > 36
+	return !(ps->origin[0] - origin[0] > 36
 	        || ps->origin[0] - origin[0] < -36
 	        || ps->origin[1] - origin[1] > 36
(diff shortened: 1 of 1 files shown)


Vorschreibung:
@Liala1 Thanks for the fix!

Vorschreibung (approved):

Aranud:
Thank you very much !

mittermichal:
seems odd to me that function that is supposed to check picking objective would be broken on only 1 map and then negating such function to not break picking objective on different maps. Am I missing something?

Aranud:
That's because there is at first a Box trace to identify if a player is "touching" a BBox ent. So there is a first check range then a second to ensure you are close enough to take the item https://github.com/etlegacy/etlegacy/blob/8ecd46b0a4a7d25ff89476ce84f189ee24d617c2/src/game/g_active.c#L439

mittermichal:
thanks.
so If I understand correctly 65e1e9aab84c6e32e0445d1c5836d5c85c313106 made players to auto-pickup objectives easier and this fixes that.
It should be communicated to competitive community since they use version that has that, so it will change for them if they update to more recent snapshot. When someone would just to do `git log` and summarize commit messages, he should somehow be aware of this.

For example on this spot you won't pick objective automatically anymore 
shot0011

PR #2999, opened 2024-11-27
cgame: Fix 1P Hand animation quirks

Vorschreibung:
Various hand parts of weapons hover, what was supposed to be, right outside of playerview.

However depending on 'cg_fov', 'cg_gun{X,Y,Z}' and 'cg_gunFovOffset' and display aspect ratio, they can sometimes pop into player view on some animation tags.

In order to fix these pop ins, some parts are now blended out altogether unless they are actually needed.


Vorschreibung:
To test:

- pick an engi with a rifle
- set `/cg_gunX 40`
- switch off/to the rifle
- {de-,at-}tach rg to it
- reload the rifle

Unless reloading, only the left arm/hand should show.

PR #3016, opened 2025-01-09
cgame: Clamp cg_fov for cg_gunFovOffset

Vorschreibung:
Stop cg_gunFovOffset "taking effect" below 90 - as this is assumed to be "zooming in".

### DIFF
--- a/src/cgame/cg_weapons.c
+++ b/src/cgame/cg_weapons.c
@@ -2626,15 +2626,16 @@ static void CG_ApplyViewWeaponShift(refEntity_t *hand, vec3_t shifts_90, vec3_t 
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


mittermichal:
IIRC some people use 75 keep it looking same as some previous version

Vorschreibung:
> IIRC some people use 75 keep it looking same as some previous version

I think that's okay, the commit isn't messing it up per se - it just "cuts off" parts of the gun model when `cg_gunFovOffset 0` as if you'd zoom/ADS.

I think the majority of players that go below 90 intend to zoom - for the ones that don't, this is not necessarily what they are after, but it also shouldn't be much of an issue for them AFAIS.

PR #3020, opened 2025-01-12
cgame: Reset 'VoiceMedic' Request on Max Health

Vorschreibung:
**DONT `Squash and Merge` THIS PR - BUT `Rebase and Merge` IT INSTEAD**

----

The 'VoiceMedic' Voice Chat Sprite (Need Medic/ Heal Me etc.) now
disappears once a player has been healed back up to Max Health.

If the 'VoiceMedic' Sprite is played when a player already is on Max
Health, the Sprite will remain until either the usual Voice Chat timeout
occurs or when the player dips below Max Health and then gets healed up
to it again.

'CG_PlayVoiceChat' is mildly refactored in order to set
'voiceChatSpriteTime' only once.

Furthermore 'identifyClientHealth' is now also recorded in ClientInfo, 
which improves the responsiveness of 'ci->health' values.

### DIFF
--- a/src/cgame/cg_snapshot.c
+++ b/src/cgame/cg_snapshot.c
@@ -306,6 +306,22 @@ static void CG_TransitionSnapshot(void)
 	BG_PlayerStateToEntityState(&cg.snap->ps, &cg_entities[cg.snap->ps.clientNum].currentState, cg.time, qfalse);
 	cg_entities[cg.snap->ps.clientNum].interpolate = qfalse;
 
+	// Also record 'ps.identifyClientHealth' (health of an alive team member
+	// you have your crosshair over) in 'cgs.clientinfo'.
+	//
+	// This is in addition to the passive health info we regularly get of all
+	// team members (for the Fireteam Window) - however 'identifyClientHealth'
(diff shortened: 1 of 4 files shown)


src/cgame/cg_snapshot.c
```diff
@@ -306,6 +306,22 @@ static void CG_TransitionSnapshot(void)
...
+	// or somesuch.
+	{
+		int identifyClientNum = cg.snap->ps.identifyClient;
+		if (identifyClientNum != cg.predictedPlayerState.clientNum)
+		{
+			clientInfo_t *ci = &cgs.clientinfo[identifyClientNum];
+			ci->health = cg.snap->ps.identifyClientHealth;
+		}
+	}
+
```
Aranud, line 324: Does it work in case an enemy is disguised as team mate ?
Does the identifyClient reflect the stolen client ?

Vorschreibung: @Aranud Seems to work fine AFAIS in tests - `identifyClient` does not reflect the stolen client, which seems however what we want, as we wouldn't want the health of the player the disguise is taken from, but the player who actually has disguised themselves as another player.

PR #3022, opened 2025-01-22
etmain: Reimplement 'medicrevive.tga'

Vorschreibung:
The OG one produced color artifacts with certain gamma settings.

Reimplemented in a vectorized form, then rasterized.


Vorschreibung:
Example of the gradient issues in the original one on certain gamma settings.
image
---
Can also be recreated by just ramping up the input levels in GIMP
image

Vorschreibung:
TODO: Same for Heal Me icon

PR #3029, opened 2025-02-19
etltv: add etpro support, refs #229

ryzyk-krzysiek:
- add entity bitstream
- add etpro commands support (working scoreboard, compass and other)
- remove unused g_debugBullets, team_riflegranades, g_pronedelay, g_debugAnim cvars
- add sv_etltv_netblast cvar (etpro ettv_netblast). Controls whether message fragments are sent all at once instead of delayed. Helps with lag introduced by fragmented messages. Default: ON
   
refs #229


src/tvgame/tvg_cmds.c
```diff
@@ -1421,6 +1423,116 @@ static void TVG_ConfigStringModified(void)
...
+		return qtrue;
+
+	// speaker state
+	case 13:
+		return qtrue;
+
+	default:
+		return qfalse;
+	}
+}
```
Aciz: These should probably be in an eunm?

src/game/bg_ebs.c
```diff
@@ -0,0 +1,548 @@
...
+
+typedef struct
+{
+	int byteOffset;
+	int numBits;
+} intField_t;
+
+// fields are in network order, same as the engine
+// selected fields do not provide forward compatibility
+#define ESF(Field, Bits) { (size_t)(&((entityState_t *)0)->Field), Bits }
```
Aciz: This can be simplified and made (arguably) more readable btw I believe (see `FOFS` in `g/tvg_local.h`):
```
#define ESF(Field, Bits) { offsetof(entityState_t, Field), Bits }
```

Edit: same deal for the re-definition below:
```
#define ESF(Field) (offsetof(entityState_t, Field))
```

src/tvgame/tvg_cmds.c
ryzyk-krzysiek: Changed to enum.

src/game/bg_ebs.c
ryzyk-krzysiek: Updated to use offsetof macro instead.

Issue #2400, opened 2023-11-08
Game does not remember last person you spectated when you die


x0rnn:
When you die and follow a certain player while waiting for respawn, it should remember that player the next time you die and automatically follow him if he is alive, but it doesn't.

x0rnn:
"IIRC it is basically vanilla code, who you spectate is saved into spectatorClient and when that person dies it is overwritten when it switches to someone else. I think new field is necessary just to store the client you want to spectate, at least that was my conclusion when I looked at it back then, maybe there is better way or I remember wrong.

Just a thought, maybe it could be possible (and simpler) to just check if the person we currently follow is dead, and instead update it to whoever that person is following (instead of overwriting spectatorClient), but it could get ugly I guess." - ryzyk_krzysiek

x0rnn:
@ryzyk-krzysiek any chance you take a look?

ryzyk-krzysiek:
> @ryzyk-krzysiek any chance you take a look?

No.

x0rnn:
With regards to https://github.com/etlegacy/etlegacy/pull/2924, maybe @Vorschreibung could take a look at this one? Similarly to that PR, when you intend to follow a specific player while waiting to respawn, you currently just spectate a random player, whereas in ETPro it always remembered the last player you spectated.

Vorschreibung:
> With regards to #2924, maybe @Vorschreibung could take a look at this one? Similarly to that PR, when you intend to follow a specific player while waiting to respawn, you currently just spectate a random player, whereas in ETPro it always remembered the last player you spectated.

![image](https://github.com/user-attachments/assets/77671802-e206-4727-b847-0877baf89629)

I'll have a look, but no promises.

Vorschreibung:
@x0rnn what exactly should happen if the player you previously spectated is dead themselves? - A simple implementation would just pick a random other allied player and override your previous choice.

x0rnn:
It would temporarily follow another random player, but the next time it would follow the one you originally followed.

Vorschreibung:
> It would temporarily follow another random player, but the next time it would follow the one you originally followed.

If you press `ATTACK` in this instance tho, it would still override the one you originally followed, yes?
Otherwise, how would you be able to set a new "target" in case the one you're following is dead?

x0rnn:
Correct, `ATTACK` would set a new target to follow.

PR #3032, opened 2025-02-19
game: remember user chosen person to spectate, refs #2400

ryzyk-krzysiek:
refs #2400


Vorschreibung:
LGTM - tested with bots

PR #3044, opened 2025-03-03
game: Fix 0-velocity missile bouncing scenarios

Vorschreibung:
In some scenarios missiles can trigger 'G_MissileImpact' and thus 'G_BounceMissile' over and over again, despite a bounce velocity of {0,0,0}.

This causes the entity to rotate and play the 'missileBouncingSound' over and over again, without actually breaking out of the loop.

This is addressed by exiting early when 'G_BounceMissile' calculates a velocity of {0,0,0} and the surrounding callsite in 'G_MissileImpact' to skip sending the 'EV_GRENADE_BOUNCE' event, which in turn would play 'missileBouncingSound' on the clients.

One way to reproduce:

1. Pick engineer
2. Give yourself Engineer 4 Flak Jacket Skill
3. Throw a landmine to your feet (but don't arm it)
4. Switch and throw/prime a grenade to your feet
5. Immediately switch to pliers and arm the landmine
6. The grenade will explode and down you. Immediately exit to limbo.
7. The landmine should now collide with your dead body and be stuck in the above missile bounce loop scenario until the body disappears (sinks into the ground).

### DIFF
--- a/src/game/g_missile.c
+++ b/src/game/g_missile.c
@@ -42,8 +42,9 @@
  * @brief G_BounceMissile
  * @param[in,out] ent
  * @param[in] trace
+ * @return true if missile actually bounced, otherwise false
  */
-void G_BounceMissile(gentity_t *ent, trace_t *trace)
+qboolean G_BounceMissile(gentity_t *ent, trace_t *trace)
 {
 	vec3_t    velocity;
 	float     dot;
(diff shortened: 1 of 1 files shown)


Vorschreibung:
Demo of how to reproduce: https://www.youtube.com/watch?v=27MKpluMHTo

Vorschreibung:
@ryzyk-krzysiek, @Aranud, @Aciz - feedback / review is appreciated

Aranud (approved):

ryzyk-krzysiek:
Looks good, just missing uncrustify.

PR #3047, opened 2025-03-03
cgame: Add 'cg_teamannouncer' CVAR

Vorschreibung:
A CVAR has been added to disable team announcer radio audio.

### DIFF
--- a/src/cgame/cg_event.c
+++ b/src/cgame/cg_event.c
@@ -2551,7 +2551,7 @@ void CG_EntityEvent(centity_t *cent, vec3_t position)
 	}
 	break;
 	case EV_GLOBAL_TEAM_SOUND:
-		if (cgs.clientinfo[cg.snap->ps.clientNum].team == es->teamNum)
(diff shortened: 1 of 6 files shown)


src/cgame/cg_event.c
```diff
@@ -2560,7 +2560,7 @@ void CG_EntityEvent(centity_t *cent, vec3_t position)
 		CG_PlayGlobalSound(cent, es->eventParm);
 		break;
 	case EV_GLOBAL_CLIENT_SOUND:
-		if (cg.snap->ps.clientNum == es->teamNum)
+		if (cg_teamannouncer.integer != 0 && cg.snap->ps.clientNum == es->teamNum)
```
Aciz: This makes `cg_teamannouncer` filter out the `playsound` command, which is probably not desired?

Vorschreibung: @Aciz Thanks for pointing it out! - This should fix it:
https://github.com/etlegacy/etlegacy/compare/8e3e88c5f83585a3f60eeba9423cde58a8d1e842..a3e57a5de8e41e345de4f1f387f98c4ece4e84a6

I repurposed the mostly unused `effect3Time` field to differentiate between team announcer voice lines and normal `playsound` commands.

PR #3051, opened 2025-03-03
cgame: fix invalid ci pointer when setting health

Aciz:
`ps.identifyClient` is not a valid clientNum if crosshair entity is an enemy, so ensure we don't do read/write ouf of bounds when setting health from `ps.identifyClientHealth`.

refs #3020


Aciz:
No longer required since the commit was reverted, but if we re-introduce this, we must ensure this safety check is still included.

Issue #2416, opened 2023-11-19
timegraph creates duplicates when resolution goes bigger?


u1ps:
![TIMEG](https://github.com/etlegacy/etlegacy/assets/99591943/091869af-49fb-4461-be77-63d7969877c9)

PR #3054, opened 2025-03-05
engine: improve time resolution from milliseconds to microseconds

ryzyk-krzysiek:
- meant as a fix for input timestamp, so when frametime drops below 1ms keydown inputs don't start being registered as not pressed ( https://www.youtube.com/watch?v=mIpGvIDW5-8 ). Inputs are "scaled" using time, so this will affect both mouse movement and keypresses. Might not be really noticeable though.
- only windows
- client and server code still uses milliseconds, only input and sleep is using more accurate time
- fix timegraph duplicating and center it, refs #2416

### DIFF
--- a/src/sys/sys_unix.c
+++ b/src/sys/sys_unix.c
@@ -67,60 +67,60 @@ static char homePath[MAX_OSPATH] = { 0 };
 #ifdef  __ANDROID__
 char *Sys_CdToExtStorage(void)
 {
-    JNIEnv *env = (JNIEnv *) SDL_AndroidGetJNIEnv();
-    jthrowable exception;
-    const char *path;
+	JNIEnv     *env = (JNIEnv *) SDL_AndroidGetJNIEnv();
+	jthrowable exception;
+	const char *path;
 
-    // Get File object for the external storage directory.
-    jclass classEnvironment = (*env)->FindClass(env, "android/os/Environment");
-    if (!classEnvironment)
-    {
-        return NULL;
-    }
-    jmethodID methodIDgetExternalStorageDirectory = (*env)->GetStaticMethodID(env, classEnvironment, "getExternalStorageDirectory", "()Ljava/io/File;"); // public static File getExternalStoragePublicDirectory ()
-    if (!methodIDgetExternalStorageDirectory)
-    {
-        return NULL;
-    }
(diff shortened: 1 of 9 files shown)


mittermichal:
would be nice if the pr-build workflow created artifacts so people could try this out :smile:

Aranud:
Doesn't snapshot made for this ?

mittermichal:
only for master branch ?

ryzyk-krzysiek:
Improved sleep time would allow for more `com_maxfps` settings. 
For example 190:
obraz
Or even something like 194:
obraz

Game simulation still doesn't allow fractional frametimes (microseconds), so when you set 190 fps your frametime is 5ms and occasional 6ms and so I wouldn't exactly say it is something to strive for. Timegraph: 
obraz

I managed to make cgame run at any frametime (qagame, server and pmove still runs at full milliseconds though and I think real solution probably doesn't exist without way too big changes that makes engine incompatible with all mods) 

https://github.com/user-attachments/assets/c0544233-0ff4-4307-a2c5-4c5e2dafba42

src/sys/sys_unix.c
```diff
@@ -269,6 +269,15 @@ int Sys_Milliseconds(void)
 	return curtime;
 }
 
+/**
+ * @brief Sys_Microseconds
+ * @return
+ */
+int64_t Sys_Microseconds(void)
+{
+	return Sys_Milliseconds() * 1000;
```
jackeri, line 278: This is more of a note (I can implement it after the PR is merged). The Unix side also needs to have the Microseconds function to do the actual time calc and the milliseconds just needs to return that value subdivided by a 1000.

ensiform:
Microseconds time function available in ETe for quite some time, and doesn't need to influence the regular milliseconds function fwiw.

https://github.com/etfdevs/ETe/blob/master/src%2Fqcommon%2Fcommon.c#L951-L980

Also works for Linux/macos not sure on others.

src/sys/sys_unix.c
x0rnn: > This is more of a note (I can implement it after the PR is merged). The Unix side also needs to have the Microseconds function to do the actual time calc and the milliseconds just needs to return that value subdivided by a 1000.

Like https://github.com/etlegacy/etlegacy/pull/3099 ?

PR #3039, opened 2025-02-26
game: Show resolved spawnpoint in Limbo Menu

Vorschreibung:
Highlight the Spawnpoint that is selected/resolved in Limbo Menu.

PR #3057, opened 2025-03-17
cgame: fix incorrect player position at the start of intermission

### LINKED ISSUE #3055: On some maps after map end scripted events do not work anymore.
Missiles and Special Deliver doors for example

2.83 - not working 

https://github.com/user-attachments/assets/3847803a-65ce-4926-8309-5f6b4ade490b

2.82 - working

https://github.com/user-attachments/assets/9e90f46e-15c1-4d38-a3c1-478f8aa6152c

Aciz:
Don't run full limbopanel setup when client info is updated, only run the relevant part (weapons), otherwise we get placed to a wrong position when intermission starts and potentially break any intermission scenes involving entities, because they are no longer in our PVS.

This also fixes `getspawnpt` being run immediately when intermission starts, similar to what `obj` did, which prompted the change to allow it during intermission (which broke this whole thing). We probably should allow it to run during intermission though, as limbo panel can be opened during intermission to switch teams, which will trigger the prompt that the command is not allowed during intermission.

Opening the limbo panel during intermission will still predictably break the intermission screen (as it triggers sending `obj -1`, which is what broke this), but fixing that is out of scope for this commit (tracked in https://github.com/etlegacy/etlegacy/issues/3058).

fixes #3055 
refs #2731 
refs #3039

### DIFF
--- a/src/cgame/cg_limbopanel.c
+++ b/src/cgame/cg_limbopanel.c
@@ -3525,9 +3525,9 @@ void CG_LimboPanel_RenderCounter(panel_button_t *button)
 }
 
 /**
- * @brief CG_LimboPanel_SetWeapons
+ * @brief CG_LimboPanel_SelectLoadout
  */
-void CG_LimboPanel_SetWeapons(void)
+void CG_LimboPanel_SelectLoadout(void)
 {
 	const clientInfo_t *ci = &cgs.clientinfo[cg.clientNum];
 	int                i;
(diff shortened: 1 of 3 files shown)


src/cgame/cg_limbopanel.c
```diff
@@ -3524,15 +3524,55 @@ void CG_LimboPanel_RenderCounter(panel_button_t *button)
 	}
 }
 
+/**
+ * @brief CG_LimboPanel_SetWeapons
+ */
+void CG_LimboPanel_SetWeapons(void)
```
Vorschreibung: This seems to set `Team` and `Class` as well, rename to: `CG_LimoPanel_SelectLoadout` ? (akin to `cgs.limboLoadoutSelected`)

Vorschreibung:
LGTM! Thanks!

PR #3061, opened 2025-03-20
ui: Make Map/Player Callvote Menu taller

Vorschreibung:
I've seen players scroll a lot, as both lists tend to be larger than the 8 or so items it can hold by default.

The vertical screen space is already there, so let's just use it.


Vorschreibung:
image

image

PR #3069, opened 2025-04-02
cgame: hide affix for ping and speedo

fghvgsx:


fghvgsx:
this will only draw the heat bar and the grenade cooking indicator

src/cgame/cg_draw_hud.c
```diff
@@ -86,8 +86,8 @@ const hudComponentFields_t hudComponentFields[] =
 	{ HUDF(weaponchargetext),   CG_DrawWeaponCharge,              0.25f,  { "Draw Suffix" } },
 	{ HUDF(fps),                CG_DrawFPS,                       0.19f,  { 0 } },
 	{ HUDF(snapshot),           CG_DrawSnapshot,                  0.19f,  { 0 } },
-	{ HUDF(ping),               CG_DrawPing,                      0.19f,  { 0 } },
-	{ HUDF(speed),              CG_DrawSpeed,                     0.19f,  { "Max Speed" } },
+	{ HUDF(ping),               CG_DrawPing,                      0.19f,  { "Draw Suffix" } },
```
Aranud: Ping label is before the value, it is a prefix !

src/cgame/cg_draw_hud.c
```diff
@@ -57,7 +57,7 @@ const hudComponentFields_t hudComponentFields[] =
 	{ HUDF(xptext),             CG_DrawXP,                        0.25f,  { "Draw Suffix" } },
 	{ HUDF(ranktext),           CG_DrawRank,                      0.20f,  { 0 } },
 	{ HUDF(statsdisplay),       CG_DrawSkills,                    0.25f,  { "Column" } },
-	{ HUDF(weaponicon),         CG_DrawGunIcon,                   0.19f,  { "Icon Flash",    "Dynamic Heat Color" } },
+	{ HUDF(weaponicon),         CG_DrawGunIcon,                   0.19f,  { "Icon Flash",    "Dynamic Heat Color", 		   "Only Charge" } },
```
Aranud: "Only charge" doesn't sound correct as it could be confused with other weapon using charge, which seems odd. 
May I suggest separate in 2 differents options "Only ticking" and "Show overheat" indicator ? or another wording.

Moreover, it could a nice plus to separate overheat indicator in his own comp, so it could be move around separatly from weapon icon (or stack up like currently)

Aranud (changes requested):

src/cgame/cg_local.h
Aranud, line 1: This need to be incremented as well !

Aranud (changes requested):

src/cgame/cg_local.h
```diff
@@ -4431,6 +4431,8 @@ typedef struct hudStructure_s
 	hudComponent_t scTeamNamesAxis;
 	hudComponent_t scTeamNamesAllies;
 
+	hudComponent_t weaponheatbar;
+
 	hudComponent_t *components[HUD_COMPONENTS_NUM];
```
Aranud, line 4435: HUD_COMPONENTS_NUM need to be incremented as well !

src/cgame/cg_draw_hud.c
```diff
@@ -192,6 +193,7 @@ void CG_setDefaultHudValues(hudStucture_t *hud)
 	hud->ranktext           = CG_getComponent(167, 465, 57, 14, qfalse, 0, 100.f, colorWhite, colorWhite, qfalse, HUD_Background, qfalse, HUD_Border, ITEM_TEXTSTYLE_SHADOWED, ITEM_ALIGN_CENTER, qfalse, 0.20f, CG_DrawRank);    // disable
 	hud->statsdisplay       = CG_getComponent(116, 394, 42, 70, qtrue, 0, 100.f, colorWhite, colorWhite, qfalse, HUD_Background, qfalse, HUD_Border, ITEM_TEXTSTYLE_SHADOWED, ITEM_ALIGN_CENTER, qfalse, 0.25f, CG_DrawSkills);
 	hud->weaponicon         = CG_getComponent(SCREEN_WIDTH - 88, SCREEN_HEIGHT - 52, 60, 32, qtrue, 1 | 2, 100.f, colorWhite, colorWhite, qfalse, HUD_Background, qfalse, HUD_Border, ITEM_TEXTSTYLE_SHADOWED, ITEM_ALIGN_CENTER, qfalse, 0.19f, CG_DrawGunIcon);
```
Aranud: Style value should stay at 1 (icon visible by default like vanilla)

src/cgame/cg_draw_hud.c
```diff
@@ -57,7 +57,8 @@ const hudComponentFields_t hudComponentFields[] =
 	{ HUDF(xptext),             CG_DrawXP,                        0.25f,  { "Draw Suffix" } },
 	{ HUDF(ranktext),           CG_DrawRank,                      0.20f,  { 0 } },
 	{ HUDF(statsdisplay),       CG_DrawSkills,                    0.25f,  { "Column" } },
-	{ HUDF(weaponicon),         CG_DrawGunIcon,                   0.19f,  { "Icon Flash",    "Dynamic Heat Color" } },
+	{ HUDF(weaponicon),         CG_DrawGunIcon,                   0.19f,  { "Icon Flash",    "Only Ticking" } },
+	{ HUDF(weaponheatbar),      CG_DrawGunHeatBar,                0.19f,  { "Dynamic Heat Color" } },
```
Aranud: weapon heat bar must be draw before weapon icon, otherwise it will be draw on top of icon, instead like in vanilla as bakground.
Declaration and assignment just need to happen before

Aranud (changes requested):

Aranud:
Otherwise, nice work !

PR #3085, opened 2025-05-22
game: fix omnibot GetEntityName null value

plastovicka:
Regression from 0f63512a5371ded87b8ae67289706e1c14209ce0


Aranud:
Nice catch, thank you very much

PR #3092, opened 2025-05-31
Headshot obituaries by Vorschreibung

x0rnn:
Continuing https://github.com/etlegacy/etlegacy/pull/2947 by @Vorschreibung, added missing headshot logic and drew an arrow through the skull as well.


src/game/bg_public.h
```diff
@@ -3038,6 +3038,7 @@ typedef enum popupMessageType_e
 	PM_CONSTRUCTION,
 	PM_MINES,
 	PM_DEATH,
+	PM_DEATH_HEADSHOT,
```
mittermichal: some backward compatibility for playing older demos would be good to have

Aciz: > some backward compatibility for playing older demos would be good to have

You can't retroactively add this to correctly display in old demos because the data where the headshot is transmitted won't be there, but this needs to be moved at the bottom of the enum, before `PM_NUM_TYPES`, in order to not mess up old demos.

PR #3099, opened 2025-06-10
engine: implement Sys_Microseconds for Unix

x0rnn:
Refs https://github.com/etlegacy/etlegacy/pull/3054

### DIFF
--- a/src/sys/sys_unix.c
+++ b/src/sys/sys_unix.c
@@ -237,6 +237,8 @@ int64_t Sys_Microseconds(void)
 	static qboolean initialized = qfalse;
 	static int64_t  timeBase_us = 0;
 	struct timespec ts;
+	int64_t currentTime_us;
+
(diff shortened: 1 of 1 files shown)


src/sys/sys_unix.c
```diff
@@ -275,7 +243,10 @@ int Sys_Milliseconds(void)
  */
 int64_t Sys_Microseconds(void)
 {
-	return Sys_Milliseconds() * 1000;
+	struct timeval curr;
+	gettimeofday(&curr, NULL);
```
Vorschreibung: This isn't monotonic (it'll be affected by timezone changes and can jump back/forth) and is comparatively slow.

If you take a glimpse at the Windows impl, it's using QueryPerformanceCounter - so we probably should likewise go for man 2 clock_gettime with CLOCK_MONOTONIC - which you have removed?

Vorschreibung (changes requested):

PR #3113, opened 2025-06-12
game: Add 'debug_crash' command

Vorschreibung:
/debug_crash
	debug_crash: confirm via calling this command again with: 2102983296 <crashType>
	             <crashType> is one of these integers: {1: Segfault, 2: Stack Overflow}]

	/debug_crash 2102983296 1
	debug_crash: Crash code confirmed, producing a Segfault ...

	Thread 1 "etl.x86_64" hit Catchpoint 2 (signal SIGSEGV), 0x00007fffa1a5e671 in cmd_debug_crash (crashType=1) at /home/vor/workspace/etlegacy/src/game/g_cmds.c:563

The Crash code is initialized randomly each time.

### DIFF
--- a/src/game/g_cmds.c
+++ b/src/game/g_cmds.c
@@ -534,6 +534,94 @@ void G_SendScore(gentity_t *ent)
 	trap_SendServerCommand(ent - g_entities, va("%s %i%s", startbuffer, count, buffer));
 }
 
+
+/**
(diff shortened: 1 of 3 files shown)


Aranud:
Perhaps this command should be executed only on /devmap condition ?

PR #3119, opened 2025-06-13
cgame: Slightly bend player corpses mid-air

Vorschreibung:


Vorschreibung:
https://github.com/user-attachments/assets/c93ddb4b-80ce-4f35-b6a0-14a06cfd723a

PR #3139, opened 2025-07-17
Rotating map compass

Aranud:


Aranud:
WIP : Players icons need adjustment

PR #3154, opened 2025-08-01
Update README.md

hugok79:
[Translation] Fix Transifex URL


dr-linetco:
Why would you only correct the link text, and not the link itself?

hugok79:
@dr-linetco Done

Aranud:
It work ! Thanks you

PR #3155, opened 2025-08-06
Fireteam spawnpoints

Helldembez:
After LAN I noticed multiple times people still struggling with spawnpoints. The spawnmenu should already help with this. But for people still using binds I think a tool that can help them is seeing who in their team has what spawnpoint. I was playing around with latchedSpawnPt and minorSpawnPt, but I think its best to keep it simple by using only majorspawnpt.

I have added the following:
- A entry for the fireteam hud component
- A entry in the clientInfo string
- A call to update client info when changing spawnpt
- Showing the spawnpt in front of the name in the fireteam under a hudcomponent check

See: https://streamable.com/nj76a8

I am not sure if I need to add anything for hud versioning.

### DIFF
--- a/src/cgame/cg_fireteamoverlay.c
+++ b/src/cgame/cg_fireteamoverlay.c
@@ -521,7 +521,7 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 			}
 			if (comp->style & BIT(6))
 			{
-				spawnPtStr[i] = va("^7%s", Q_CleanStr(CG_GetLocationMsg(ci->clientNum, cgs.majorSpawnpointEnt[ci->spawnpt - 1].origin)));
+				spawnPtStr[i] = va("%s", Q_CleanStr(CG_GetLocationMsg(ci->clientNum, cgs.majorSpawnpointEnt[ci->spawnpt - 1].origin)));
 				if (cg_locationMaxChars.integer)
 				{
 					spawnwidth  = spacing;
@@ -842,8 +842,6 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 		x += spacing;
 		float widthLocationLeft = w - (x - comp->location.x) - spacing;
 		float charWidth         = CG_Text_Width_Ext_Float("A", scale, 0, FONT_TEXT);
-
-		// paint
 		if (cg_locations.integer & LOC_FTEAM)
 		{
 			float locWidthLocationLeft = widthLocationLeft;
@@ -879,6 +877,22 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 		}
 		if (comp->style & BIT(5))
 		{
+			vec4_t *spawnPtColor;
+			float  timeDiff = cg.time - ci->spawnChangedTime;
+			if (timeDiff < 1500.0f)
+			{
+				float t = timeDiff / 1500.0f;
+				spawnPtColor = (vec4_t[]) { { colorOrange[0] + (colorWhite[0] - colorOrange[0]) * t,
+											  colorOrange[1] + (colorWhite[1] - colorOrange[1]) * t,
+											  colorOrange[2] + (colorWhite[2] - colorOrange[2]) * t,
+											  colorOrange[3] + (colorWhite[3] - colorOrange[3]) * t } };
+			}
+			else
+			{
+				spawnPtColor = colorWhite;
+			}
+
+
 			if (cg_locations.integer & LOC_FTEAM)
 			{
 				widthLocationLeft = widthLocationLeft / 2;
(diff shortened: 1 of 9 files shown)


mittermichal:
awesome!
why not also minor spawn point?

x0rnn:
Yea, I think minor spawn info would be useful too - you can see where teammates spawn so you can communicate or change to another minor spawn.

Helldembez:
So I added minor spawnpoints and changed it to be the live spawnpoint. E.g. when cp gets blown it shows the correct one. It also shows by color that one was changed.

I was struggling with keeping the old behavior of locations and adding this as a new last number/text. I'm not sure but it can be its not the exact same behavior anymore. Especially with left/right alignment.

Also, if anyone notices I do things in the "Wrong" place let me know. As well as the relevance of these 2 failing Github checks(seems a lot is code i didnt touch?).

src/cgame/cg_fireteamoverlay.c
```diff
@@ -802,26 +840,91 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
...
 			}
+			else
+			{
+				CG_Text_Paint_Ext(x, y + heightTextOffset, scale, scale, *spawnPtColor, Q_TruncateStr(spawnPtStr[i], 0), 0, 0, comp->styleText, FONT_TEXT);
+				if (comp->style & BIT(7) && ci->mspawnpt > 0)
+				{
+					CG_Text_Paint_Ext(x + charWidth, y + heightTextOffset, scale / 1.5, scale / 1.5, *spawnPtColor, va("%i", ci->mspawnpt), 0, 0, comp->styleText, FONT_TEXT);
+				}
+
+			}
```
Aciz:
```cpp
vec4_t *spawnPtColor;
```

This being a pointer to an array is wrong. First off, you cannot just assign an array directly into a pointer to an array, you must assign the address of the array. This would still be very bad here, because now you'd be mutating the global `colorWhite` array that is defined in `q_math.h`, this is just language semantics.

```diff
- spawnPtColor = colorWhite;
+ spawnPtColor = &colorWhite;
```

Because `spawnPtColor` was a pointer to an array, you now had to dereference it in all the calls to the text paint functions. This is fine and works, and is what you need to do if you have a pointer to an array, but we don't want to modify the global white color array here, so just make `spawnPtColor` an array instead of a pointer to an array. I also simplified the color lerping by just using `LerpColor` function.

```diff
diff --git a/src/cgame/cg_fireteamoverlay.c b/src/cgame/cg_fireteamoverlay.c
index e1ad7994b..840fc0ba8 100644
--- a/src/cgame/cg_fireteamoverlay.c
+++ b/src/cgame/cg_fireteamoverlay.c
@@ -877,19 +877,15 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 		}
 		if (comp->style & BIT(5))
 		{
-			vec4_t *spawnPtColor;
-			float  timeDiff = cg.time - ci->spawnChangedTime;
+			const float timeDiff = cg.time - ci->spawnChangedTime;
+
+			vec4_t spawnPtColor;
+			Vector4Copy(colorWhite, spawnPtColor);
+
 			if (timeDiff < 1500.0f)
 			{
-				float t = timeDiff / 1500.0f;
-				spawnPtColor = (vec4_t[]) { { colorOrange[0] + (colorWhite[0] - colorOrange[0]) * t,
-											  colorOrange[1] + (colorWhite[1] - colorOrange[1]) * t,
-											  colorOrange[2] + (colorWhite[2] - colorOrange[2]) * t,
-											  colorOrange[3] + (colorWhite[3] - colorOrange[3]) * t } };
-			}
-			else
-			{
-				spawnPtColor = colorWhite;
+				const float t = timeDiff / 1500.0f;
+				LerpColor(colorOrange, colorWhite, spawnPtColor, t);
 			}
 
 
@@ -918,10 +914,10 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
 			}
 			else
 			{
-				CG_Text_Paint_Ext(x, y + heightTextOffset, scale, scale, *spawnPtColor, Q_TruncateStr(spawnPtStr[i], 0), 0, 0, comp->styleText, FONT_TEXT);
+				CG_Text_Paint_Ext(x, y + heightTextOffset, scale, scale, spawnPtColor, Q_TruncateStr(spawnPtStr[i], 0), 0, 0, comp->styleText, FONT_TEXT);
 				if (comp->style & BIT(7) && ci->mspawnpt > 0)
 				{
-					CG_Text_Paint_Ext(x + charWidth, y + heightTextOffset, scale / 1.5, scale / 1.5, *spawnPtColor, va("%i", ci->mspawnpt), 0, 0, comp->styleText, FONT_TEXT);
+					CG_Text_Paint_Ext(x + charWidth, y + heightTextOffset, scale / 1.5, scale / 1.5, spawnPtColor, va("%i", ci->mspawnpt), 0, 0, comp->styleText, FONT_TEXT);
 				}
 
 			}

```

src/cgame/cg_local.h
```diff
@@ -1418,6 +1421,9 @@ typedef struct
 	int numMiscGameModels;
 	int numCoronas;
 	int numSpawnpointEnts;
+	int numMajorSpawnpointEnts;
+
+	qboolean supportsMinorSpawnPoints;
```
Aciz: Nitpick, but should imo be named something like `hasMinorSpawnPoints`, every map "supports" minor spawnpoints, that's a standard feature in legacy mod, it's about whether they are defined in the map or not. "Support" to me indicates something that needs to be explicitly enabled, but minor spawn points are implicitly supported.

src/cgame/cg_fireteamoverlay.c
```diff
@@ -802,26 +840,91 @@ void CG_DrawFireTeamOverlay(hudComponent_t *comp)
...
+				if (comp->style & BIT(6))
+				{
+					locWidthLocationLeft = (widthLocationLeft - spacing) / 2;
+				}
+				else
+				{
+					locWidthLocationLeft -= bestSpawnWidth + spacing;
+				}
+			}
+			int lim = locWidthLocationLeft / charWidth;
```
Aciz: `lim` and `widthLocationLeft` should be declared earlier, this now violates `-Wdeclaration-after-statement`.

Aciz (changes requested):
Seems to work and is fine other than the mentioned issues :+1:

mittermichal:
shot0000
If you select "non-available" spawn such as enemy spawn or some disabled one, it shows where you actually spawn which in  can be useful, but it also hides the information that you chose such spawn and will be placed on particular minor spawn - in the screenshot I chose spawn point 2, which will place me at right side of axis tank spawn in sw_goldrush_te but my fireteam teammates won't know that.

Helldembez:
The specific spawn point of a major spawn is resolved at moment of spawning. So you cant accurately show it in fireteam. If 2 people do this, how will you predict who will spawn where?

This problem also happens if you are on a minor spawnpoint map and select a minor spawnpoint, then the cp gets blown you still spawn on that minor spawn point, but on a other major one. Also when someone is standing on your spawn point, not even spawning there, will spawn you on a other point.

mittermichal:
> The specific spawn point of a major spawn is resolved at moment of spawning. 

Yes, but if I saw 1 teammate with spawnpt 2 in fireteam, I could expect him to spawn there.

I suggest to have another toggle to be able to see requested major spawn point.

Helldembez:
I think a better solution would be fixing minor spawnpoints in all maps.

src/cgame/cg_local.h
```diff
@@ -649,6 +649,9 @@ typedef struct clientInfo_s
 	int breathPuffTime;
 	int cls;
 	int latchedcls;
+	int spawnpt;
+	int mspawnpt;
```
Aciz, line 653: I do wonder if this should be renamed to something else (and also the key that is written to client info string), `mspawnpt` can be interpreted as either major or minor spawnpoint :D This would be easy to just change to `int minorspawnpt`, but I don't really have a suggestion for the client info string, maybe just `misp`, dunno. But this isn't required really, it's not very hard to figure out what this is for someone who's gonna look at this in few months.

Aciz (approved):

src/cgame/cg_local.h
Aranud: The variable can be very descriptive as `majorSpawnPoint` and `minorSpawnPoint`, so as you suggest later on you can quickly to the link between the variable name and the key is assigned to.
The key, in contrary, must be shorter : `masp` and `misp` sound good. They are not meant to be quickly descriptive at first look.

mittermichal:
> I think a better solution would be fixing minor spawnpoints in all maps.

You mean adding them to every map? Even then the issue I described would still exists.

I think this feature is great, and if it were also extended with the requested spawn point, it would be even better. But I understand that adding it is a bit of a chore.

PR #3174, opened 2025-09-09
etmain: revamp blood puffs/splashes

Vorschreibung:
Blood puffs now rival smoke puffs in visiblity - giving less reason to turn them off.

Blood spurts are mildly less obtrusive.


Vorschreibung:
Some random shots - buffs and spurts

image

image

image

PR #3214, opened 2025-11-26
cgame: fix needle bar placement in HUD for certain combinations

MijPeter:
Hey, position of needle bar is incorrect when LEFT checkbox is unticked.

image

Below is after fix:
image


Aranud:
Nice catch. Thank you very much !

PR #3216, opened 2025-11-29
etmain: add shader scripts for cm_flagshadow so that its not drawn ou…

MijPeter:
…tside of compass

I figured it should have same shader script as allies/axis flags (as this is basically an extension of flag).

Screen that shows how it was before fix:
image

And screen that shows how it is after:
image


Aranud:
Thank you very much !

PR #3219, opened 2025-12-02
cgame: add cg_bloodForcePuffsForDamage

Vorschreibung:
Default "1" - if enabled always shows smoke puffs for shooting
invulnerable players (spawn prot, rez shield) otherwise always shows
blood puffs.


Vorschreibung:
image
image

PR #3220, opened 2025-12-02
cgame: fix cursor timeout when interacting with demo UI

Aciz:
Cursor timeout wasn't correctly accounting for the timestamp changes that happened as a result of rewind/fast-forward/seek, resulting in interacting with the UI making the cursor disappear until it was moved again. This only works while interacting with the UI and not when inputting commands in console, as the commands are implemented in engine, so cgame doesn't know about them. They should probably be part of cgame at some point, because the whole rewind stuff is only supported in legacy mod demos anyway.


Aciz:
Included in #3254

PR #3225, opened 2025-12-17
Add CvarRegisterExt Extension

Vorschreibung:
Ought to be usable by cgame etc. to register a CVAR alongside with
Description and Validation in a single call.


Vorschreibung:
superseded

PR #3230, opened 2026-01-03
rendercommon: shrink splash image < 512 res

Vorschreibung:


Vorschreibung:
Thanks to @Aciz for finding and fixing!

PR #3234, opened 2026-01-04
README: add troubleshooting steps for X11 audio and x11 input issues

mittermichal:
based on investigation in private discord channel


Vorschreibung:
Thanks!

PR #3238, opened 2026-01-07
annotate compat structs

Vorschreibung:


src/game/g_local.h
```diff
@@ -950,7 +950,7 @@ typedef struct damageReceivedStats_s
  * @brief This structure is cleared on each ClientSpawn(),
  * except for 'client->pers' and 'client->sess'
  */
-struct gclient_s
+struct ETL_260B_COMPAT gclient_s
```
Aciz: This doesn't need an annotation, you are free to modify this struct in any way.

Aciz (changes requested):
This isn't exhaustive ofc, but it does cover all structs in `q_shared.h` correctly, so that's a good start. Technically I believe most of them are extendable if the extended fields are only used in mod code, but regardless annotating them is good so that you don't accidentally do stuff that breaks 2.60b compatibility.

I touched on a bit on the naming in Discord already, that I don't think "compatibility" is really a good term to use here, as it implies you're building a compatibility layer. I'd prefer something that rather tells you that you should not edit this, like `ETL_IMMUTABLE` or `ETL_NOEDIT` or something.

Vorschreibung:
> I touched on a bit on the naming in Discord already, that I don't think "compatibility" is really a good term to use here, as it implies you're building a compatibility layer. I'd prefer something that rather tells you that you should not edit this, like `ETL_IMMUTABLE` or `ETL_NOEDIT` or something.

Renamed to `ETL_260B_NOEDIT`

PR #3239, opened 2026-01-08
ui: add last server in main menu

Vorschreibung:


src/ui/ui_main.c
```diff
@@ -1001,6 +1003,7 @@ void UI_Refresh(int realtime)
 	}
 
 	UI_UpdateCvars();
+	UI_UpdateLastServerPing();
```
Aciz, line 1054: This should only update when client `connState` is `<= CA_DISCONNECTED`, otherwise you're unnecessarily pinging servers when you bring up e.g. in-game menu while on server, yet you're only ever able to see the window in main menu.

Aciz (commented):
I'm not sure what I'm doing wrong when testing, but this just doesn't seem to work for me? No matter what I do, the last server info never gets saved.

I'm also not sure what the point of saving things like player counts and map is, because as far as I can tell, those fields are only ever saved when you connect to a server, so by the time you've disconnected and want to reconnect, the info is going to be out of date. The UI refresh seems to only update ping of the last visited server, and that's it.

PR #3287, opened 2026-02-07
etmain: skip single option menus in classmenu

MijPeter:
Currently classmenu displays options to player even when there is only one option to choose from.

In such cases there is no point waiting for player selection. This improvement is especially useful in gathers setting as it reduces clicks required to change to medic from 4 to 2.

I've made the code edits using onOpen, hoping it's an ok solution. I've tried to maintain previous structure of code to make reviewing and updating easier.

Disclaimer: Similar improvement CANNOT be done for spawnptmenu as there we sometimes want to select 'unselectable' spawnpts.

### DIFF
--- a/etmain/ui/wm_class.menu
+++ b/etmain/ui/wm_class.menu
@@ -107,7 +107,7 @@ QM_MENU_START( "wm_class_soldier" )
 QM_MENU_END
 
 QM_MENU_START( "wm_class_soldier_p1_s1" )
-	QM_MENU_ITEM_TEAM(	"P. Pistol",			exec "class s 1 1";	close wm_class_soldier_p1_s1,		"p", 0 )
+	onOpen { exec "class s 1 1";	close wm_class_soldier_p1_s1 }
 QM_MENU_END
 
 QM_MENU_START( "wm_class_soldier_p1_s2" )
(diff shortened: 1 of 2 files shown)


etmain/ui/wm_class.menu
```diff
@@ -256,13 +256,13 @@ QM_MENU_START( "wm_class_engineer_p2_s2" )
 QM_MENU_END
 
 QM_MENU_START( "wm_class_fieldops" )
-	QM_MENU_ITEM_TEAM(	"S. SMG",				uiScript clientCheckSecondaryWeapon ; conditionalscript cg_ui_secondary_lw 0
-												( "close wm_class_fieldops ; open wm_class_fieldops_p1_s2" )
-												( "close wm_class_fieldops ; open wm_class_fieldops_p1_s1" ),	"s", 0 )
+   onOpen { uiScript clientCheckSecondaryWeapon ; conditionalscript cg_ui_secondary_lw 0
+            ( "close wm_class_fieldops ; open wm_class_fieldops_p1_s2" )
+            ( "close wm_class_fieldops ; open wm_class_fieldops_p1_s1" ),	"s", 0 ) }
```
MijPeter: ah, this is not entirely correct

Aranud:
Thanks you !

PR #3295, opened 2026-02-11
qcommon: add 'dumpcvars' command

Vorschreibung:
Add a command to export the full cvar registry in .jsonl form.

Added a new `dumpcvars` command in `src/qcommon/cvardump.c` that writes `cvardump.jsonl` under `fs_basepath`.

Each cvar is emitted as a json object in a single line (JSONL).

### DIFF
--- a/src/qcommon/cvardump.c
+++ b/src/qcommon/cvardump.c
@@ -1,0 +1,213 @@
+/*
+ * Wolfenstein: Enemy Territory GPL Source Code
+ * Copyright (C) 1999-2010 id Software LLC, a ZeniMax Media company.
+ *
+ * ET: Legacy
+ * Copyright (C) 2012-2024 ET:Legacy team <mail@etlegacy.com>
+ *
+ * This file is part of ET: Legacy - http://www.etlegacy.com
+ *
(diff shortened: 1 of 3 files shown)


src/qcommon/cvardump.c
```diff
@@ -0,0 +1,213 @@
...
+void Cvar_Dump_f(void)
+{
+	char        dumpPath[MAX_OSPATH];
+	char        flagNames[512];
+	const char  *lastName = NULL;
+	cvar_t      *var;
+	FILE        *f;
+	int         dumpedCount = 0;
+
+	Q_strncpyz(dumpPath, FS_BuildOSPath(Cvar_VariableString("fs_basepath"), "", "cvardump.jsonl"), sizeof(dumpPath));
```
Aciz: Why? This by default (aka default install location) is a write protected directory on Windows, so this command will just not work in that case.

Vorschreibung: What directory can I use instead to dump this file to?

Aciz: `fs_homepath`, that is the standard for any files generated during runtime.

PR #3297, opened 2026-02-11
cgame: display revivable bodies of fireteam members in command map ev…

MijPeter:
…en when they're outside of pvs.

This change shouldn't have any influence on public servers and if you decide to join same fireteam with some1 on public then Id say it's an improvement for public as well.

But the main reason I would like this to be added is of course gathers. Currently ur teammate can be revivable 10m next to you but you have no way of telling where they are because they are outside of pvs.
Furthermore location information is already sent to fireteam hud element so imo this logic makes sense even more.

I really think this change is non controversial and an improvement.


Aranud:
Thanks you!

PR #3307, opened 2026-02-13
Reinforce ticktock

Vorschreibung:


Aranud:
The idea is great and I think we may push the concept a bit farther : 
- The tick sound length shall be 1 second length so player can control the number of tick before reinforcement
- By doing a 1 second sample we may increment for each seconds the volume of each tick until reinforcement happen
- It will ease to fix the issue when player join between starting timer sound and reinforcement happen
- It will ease to fix the issue when player leave the team and sound still playing
- Extra: add an extra sound for the last tick beep to be different from the first regular one, so it underline the reinforcement happen now

Aranud:
Just perfect. It work flawlessly.

It seem the PR contain other changes which are not related to reinforcement ticktock feature.
They work as well, but I guess they are not supposed to be here ?

PR #3319, opened 2026-02-20
weapzoom (fov toggle) feature

utrolig:
## Overview
Implemented a new client-side **weapzoom** feature (renamed from ADS) with smooth FOV transitions, sensitivity scaling, optional command hooks, and proper restrictions for mounted/set weapons. The feature is fully bindable and integrates with existing zoom/scope behavior.

## Key Features
- **Bindable commands**
  - `+weapzoom` / `-weapzoom` (hold)
  - `toggleweapzoom` (toggle)

- **Weapzoom FOV + timing**
  - `cg_weapzoomFov` (default `75`)
  - `cg_weapzoomInTimeMs` (default `100`)
  - `cg_weapzoomOutTimeMs` (default `100`)

- **Sensitivity scaling**
  - `cg_weapzoomSensitivityScale` (default `1.0`)
  - `cg_weapzoomSensitivityOverride` (default `0`, overrides scaling when > 0)
  - Scaling uses FOV ratio: `tan(weapzoomFov/2) / tan(cg_fov/2)`

- **Event hooks**
  - `cg_onWeapzoomStart` / `cg_onWeapzoomEnd`
  - Executed on weapzoom start/end with a newline appended to avoid concatenation issues

- **Gun FOV alignment**
  - Weapon model offsets use weapzoom‑interpolated FOV by default
  - Gun FOV follows weapzoom only when `cg_gunFovOffset` is `0`. If `cg_gunFovOffset` is non‑zero, weapon offsets use base `cg_fov`

- **Restrictions / compatibility**
  - Weapzoom is disabled while scoped (`cg.zoomed` / `cg.zoomedBinoc` / `cg.zoomval`)
  - Weapzoom is disabled for mounted weapons and set MGs:
    - `PERS_HWEAPON_USE`
    - `EF_MG42_ACTIVE` / `EF_AAGUN_ACTIVE` / `EF_MOUNTEDTANK`
    - `WEAPON_TYPE_SET` + `WEAPON_TYPE_MG`

## Example Config
```cfg
set cg_weapzoomFov 75
set cg_weapzoomInTimeMs 100
set cg_weapzoomOutTimeMs 150
set cg_weapzoomSensitivityScale 1.0
set cg_weapzoomSensitivityOverride 0
set cg_onWeapzoomStart "cg_crosshairSize 32; cg_crosshairColor blue"
set cg_onWeapzoomEnd "cg_crosshairSize 40; cg_crosshairColor red"
bind mouse2 "+weapzoom"
bind mouse3 "toggleweapzoom"
```

[Updated video](https://www.youtube.com/watch?v=SM2_lrrtnog)


Vorschreibung:
It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.

src/cgame/cg_cvars.c
```diff
@@ -80,6 +80,12 @@ vmCvar_t cg_tracerLength;
 vmCvar_t cg_tracerSpeed;
 vmCvar_t cg_autoswitch;
 vmCvar_t cg_fov;
+vmCvar_t cg_adsFov;
+vmCvar_t cg_adsInTimeMs;
```
Vorschreibung: In & Out should be coalesced into one.

utrolig:
> It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.

The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line

Vorschreibung:
> > It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.
> 
> The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line

Well it's important to me so you can either address it, discuss it further or it won't get merged

src/cgame/cg_cvars.c
```diff
@@ -338,6 +345,12 @@ static cvarTable_t cvarTable[] =
 	{ &cg_zoomDefaultSniper,                  "cg_zoomDefaultSniper",                  "20",          CVAR_ARCHIVE,                 0 }, // changed per atvi req
 	{ &cg_zoomStepSniper,                     "cg_zoomStepSniper",                     "2",           CVAR_ARCHIVE,                 0 },
 	{ &cg_fov,                                "cg_fov",                                "90",          CVAR_ARCHIVE,                 0 },
+	{ &cg_adsFov,                             "cg_adsFov",                             "75",          CVAR_ARCHIVE,                 0 },
+	{ &cg_adsInTimeMs,                        "cg_adsInTimeMs",                        "100",         CVAR_ARCHIVE,                 0 },
+	{ &cg_adsOutTimeMs,                       "cg_adsOutTimeMs",                       "100",         CVAR_ARCHIVE,                 0 },
+	{ &cg_adsSensitivityRatio,                "cg_adsSensitivityRatio",                "1.1",         CVAR_ARCHIVE,                 0 },
```
Vorschreibung: 1.1 seems to increase sensitivity, not decrease it; should probably be 0.9 instead with 0.5 halving sensitivity

src/cgame/cg_cvars.c
```diff
@@ -338,6 +345,12 @@ static cvarTable_t cvarTable[] =
...
+	{ &cg_adsSensitivity,                     "cg_adsSensitivity",                     "0",           CVAR_ARCHIVE,                 0 },
```
Vorschreibung: rename cg_adsSensitivity to `cg_adsSensitivityOverride` to make clear that it overrides ratio

src/cgame/cg_cvars.h
```diff
@@ -80,6 +80,12 @@ extern vmCvar_t cg_tracerLength;
 extern vmCvar_t cg_tracerSpeed;
 extern vmCvar_t cg_autoswitch;
 extern vmCvar_t cg_fov;
+extern vmCvar_t cg_adsFov;
+extern vmCvar_t cg_adsInTimeMs;
+extern vmCvar_t cg_adsOutTimeMs;
+extern vmCvar_t cg_adsSensitivityRatio;
+extern vmCvar_t cg_adsSensitivity;
+extern vmCvar_t cg_adsAffectsGunFov;
```
Vorschreibung: Why is this a separate CVAR? What for?

src/cgame/cg_local.h
```diff
@@ -1251,6 +1251,13 @@ typedef struct
 	int zoomTime;
 	float zoomSensitivity;
 	float zoomval;
+	qboolean adsActive;
+	qboolean adsWasActive;
+	qboolean adsStartValid;
+	int adsTime;
+	float adsStartFov;
+	float adsCurrentFov;
+	float adsBlend;
```
Vorschreibung: Either better name or comment what it does. (`adsBlend`)

Vorschreibung (changes requested):

Aciz:
> > > It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.
> > 
> > 
> > The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line
> 
> Well it's important to me so you can either address it, discuss it further or it won't get merged

Just add `cg_onAdsStart/cg_onAdsEnd` cvars so you don't need to fuck around with million variables and everyone can do whatever they want with their crosshair on ads on/off. This would also eliminate the need for separate sensitivity scaler cvar, as you can just set a different sensitivity using these.

Edit: OTOH, maybe having the sensitivity scaler cvar is still valuable, as you then only need to worry about a single sensitivity value if you adjust your base sensitivity.

src/cgame/cg_cvars.c
Aciz: Why? I think it's totally viable to want to have different in/out zoom times?

Vorschreibung:
@utrolig 
ADS is broken when looking down a deployed MG42/Browning; needs to be fixed

utrolig:
> > > > It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.
> > > 
> > > 
> > > The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line
> > 
> > 
> > Well it's important to me so you can either address it, discuss it further or it won't get merged
> 
> Just add `cg_onAdsStart/cg_onAdsEnd` cvars so you don't need to fuck around with million variables and everyone can do whatever they want with their crosshair on ads on/off. This would also eliminate the need for separate sensitivity scaler cvar, as you can just set a different sensitivity using these.
> 
> Edit: OTOH, maybe having the sensitivity scaler cvar is still valuable, as you then only need to worry about a single sensitivity value if you adjust your base sensitivity.

I would want to keep the sensitivity ratio since it also goes between the two values smoothly. The ratio is not just a multiplier, it's some calculation that lowers the sensitivity based on the difference of your fov and your adsFov. So if you have `cg_fov 90` and `cg_adsFov 80` it will adjust it differently than if you have `cg_adsFov 75`. Does that make sense? @Vorschreibung

utrolig:
Also, do we want to rename the whole feature to `weapzoom` instead of `ads` ?

utrolig:
> @utrolig ADS is broken when looking down a deployed MG42/Browning; needs to be fixed

Fixed in my latest commit.

src/cgame/cg_cvars.h
utrolig: Which one of them do you mean?

src/cgame/cg_cvars.c
utrolig: It does not. 

```
 scale = tan(cg_adsFov/2) / tan(cg_fov/2)
        = tan(75/2) / tan(90/2)
        = tan(37.5°) / tan(45°)
        = tan(37.5°) / 1
        ≈ 0.767
```

With `cg_fov 90`, `cg_adsFov 75`, `sensitivity 1` and `cg_adsSensitivityRatio 1`,  your ADS sensitivity will be about 0.77 once fully zoomed in (and it blends toward that during the ADS transition). I will make it `1` by default instead of `1.1`

Vorschreibung:
> Also, do we want to rename the whole feature to `weapzoom` instead of `ads` ?

Yes.

Vorschreibung:
> > > > > It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.
> > > > 
> > > > 
> > > > The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line
> > > 
> > > 
> > > Well it's important to me so you can either address it, discuss it further or it won't get merged
> > 
> > 
> > Just add `cg_onAdsStart/cg_onAdsEnd` cvars so you don't need to fuck around with million variables and everyone can do whatever they want with their crosshair on ads on/off. This would also eliminate the need for separate sensitivity scaler cvar, as you can just set a different sensitivity using these.
> > Edit: OTOH, maybe having the sensitivity scaler cvar is still valuable, as you then only need to worry about a single sensitivity value if you adjust your base sensitivity.
> 
> I would want to keep the sensitivity ratio since it also goes between the two values smoothly. The ratio is not just a multiplier, it's some calculation that lowers the sensitivity based on the difference of your fov and your adsFov. So if you have `cg_fov 90` and `cg_adsFov 80` it will adjust it differently than if you have `cg_adsFov 75`. Does that make sense? @Vorschreibung

Yes, keep sensitivity ratio, it's good.

Aciz wanted something different tho, namely to have `onAdsStart` / `onAdsEnd` hooks.
I think it's overkill, but don't have a huge opinion, talk it out with @Aciz

src/cgame/cg_cvars.c
Vorschreibung: @Aciz  Are you sure? Too many CVARs already, I don't see the use case

src/cgame/cg_cvars.c
Vorschreibung: I wrote badly, what I meant is that it's unintuitive I think. Currently the higher this CVAR is set, the lower the resulting sensitivity will be.

It arguably would be more intuitive to have a CVAR where the lower you set the lower the resulting sensitivity will be.

I'm not sure "ratio" is a good term either. Ratio of what? Users don't know of which.

I suggest to rename it `cg_adsSensitivityScale` maybe and have lower values actually mean lower resulting sensitivity with 1.0 meaning no change in sensitivity.

This seems much more intuitive, scales can be non-linear just fine so it should be understandable to most users.

src/cgame/cg_cvars.h
Vorschreibung: line 88; `cg_adsAffectsGunFov` I reviewed closer, I'm in favor of removing it, we don't want too many CVARs.

Weapon Zooming should IMO always affect gun fov if dynamic gun fov is set (if cg_gunFov is 0).
Otherwise it should not affect gun fov.

i.e. the CVAR seems to have no value

Vorschreibung (changes requested):

Aciz:
> > > > > > It's not bad, I'll review in a while, Crosshairsize change e.g. is not that trivial because we have customCrosshairs as well, will post more later.
> > > > > 
> > > > > 
> > > > > The crosshair part is not super important to me, I can already feel that people will start asking for more things to change etc and it's hard to know where to draw the line
> > > > 
> > > > 
> > > > Well it's important to me so you can either address it, discuss it further or it won't get merged
> > > 
> > > 
> > > Just add `cg_onAdsStart/cg_onAdsEnd` cvars so you don't need to fuck around with million variables and everyone can do whatever they want with their crosshair on ads on/off. This would also eliminate the need for separate sensitivity scaler cvar, as you can just set a different sensitivity using these.
> > > Edit: OTOH, maybe having the sensitivity scaler cvar is still valuable, as you then only need to worry about a single sensitivity value if you adjust your base sensitivity.
> > 
> > 
> > I would want to keep the sensitivity ratio since it also goes between the two values smoothly. The ratio is not just a multiplier, it's some calculation that lowers the sensitivity based on the difference of your fov and your adsFov. So if you have `cg_fov 90` and `cg_adsFov 80` it will adjust it differently than if you have `cg_adsFov 75`. Does that make sense? @Vorschreibung
> 
> Yes, keep sensitivity ratio, it's good.
> 
> Aciz wanted something different tho, namely to have `onAdsStart` / `onAdsEnd` hooks. I think it's overkill, but don't have a huge opinion, talk it out with @Aciz

That was mainly with regards to `cg_adsCrosshairSize`, this will very very quickly turn into "can we also have cvar for X/Y/Z when you ADS", my proposal was to just add a cvars that hook into the ADS on/off, so we don't need cvar for every single thing one might want to adjust with this, and everyone can just decide themselves on what they want to happen, if any.

While you could of course script this, the downside of that is that you can't track the state while scripting, e.g. if you die while ADS-ing, your "ADS off" script will never execute automatically. Having an internal "event listener" for this makes this a non-issue.

src/cgame/cg_cvars.c
utrolig: Did you compile and test this? Because it works the way you want. Lower scale = lower sensitivity. 

1.0 does not mean no change in sensitivity, it means it scales it down by the amount of "zoom". 

So essentially a 2x zoom should give half the sensitivity.

Vorschreibung: I've checked again and you are correct.

This leaves renaming it to `cg_adsSensitivityScale` for the reasons above, unless you disagree then I'd like to know why.

utrolig: I used the name `cg_weapzoomSensitivityRatio` because that's the wording that is used in source/source2 engine which is where I took the inspiration from. I can rename it to `cg_weapzoomSensitivityScale` if you feel that is more correct.

utrolig: https://github.com/ValveSoftware/source-sdk-2013/blob/11a677c349b149b2f77184dc903e6bb17f8df69b/src/game/client/view.cpp#L86

Vorschreibung: Yeah, we're not source engine and our users don't expect it either, please rename it to `cg_weapzoomSensitivityScale`

src/cgame/cg_cvars.h
utrolig: `cg_gunFov` doesn't seem to be a valid cvar

utrolig: I've removed `cg_adsAffectsGunFov` and made it the default

src/cgame/cg_local.h
utrolig: Renamed it to `weapzoomLerp` and added a short comment about what it does

src/cgame/cg_cvars.c
utrolig: done

src/cgame/cg_cvars.c
utrolig: done

utrolig:
This is ready for review now @Aciz @Vorschreibung

src/cgame/cg_consolecmds.c
```diff
@@ -86,6 +86,62 @@ static void CG_Viewpos_f(void)
 	          round(cg.refdefViewAngles[PITCH]), round(cg.refdefViewAngles[YAW]), round(cg.refdefViewAngles[ROLL]));
 }
 
+/**
+ * @brief CG_WeapzoomDown_f
+ */
+static void CG_WeapzoomDown_f(void)
+{
+	qboolean weapzoomAllowed = !(cg.zoomed || cg.zoomedBinoc || cg.zoomval != 0.f)
```
Vorschreibung: factor this and line 131 into a new function with ID_INLINE and then use it in line 100 and line 137 to deduplicate and to get rid of the the qboolean local

src/cgame/cg_local.h
```diff
@@ -1251,6 +1251,14 @@ typedef struct
 	int zoomTime;
 	float zoomSensitivity;
 	float zoomval;
+	qboolean weapzoomActive;
+	qboolean weapzoomWasActive;
+	qboolean weapzoomStartValid;
+	int weapzoomTime;
+	float weapzoomStartFov;
+	float weapzoomCurrentFov;
+	// Lerp factor [0..1] for weapzoom transition/sensitivity ramp.
```
Vorschreibung: remove this comment

src/cgame/cg_consolecmds.c
Vorschreibung: like this:

```
static ID_INLINE qboolean CG_WeapzoomAllowed_f() {
	return !(cg.zoomed || cg.zoomedBinoc || cg.zoomval != 0.f)
	                           && !(cg.snap->ps.persistant[PERS_HWEAPON_USE])
	                           && !(cg.snap->ps.eFlags & (EF_MG42_ACTIVE | EF_AAGUN_ACTIVE | EF_MOUNTEDTANK))
	                           && !((GetWeaponTableData(cg.snap->ps.weapon)->type & WEAPON_TYPE_SET)
	                                && (GetWeaponTableData(cg.snap->ps.weapon)->type & WEAPON_TYPE_MG));
}
```

src/cgame/cg_view.c
```diff
@@ -1089,6 +1090,94 @@ static int CG_CalcFov(void)
 		}
 	}
 
+	if (cg.predictedPlayerState.pm_type != PM_INTERMISSION && !cg.showGameView)
+	{
+		qboolean weapzoomAllowed = !(cg.zoomed || cg.zoomedBinoc || cg.zoomval != 0.f)
```
Vorschreibung: can also use `CG_WeapzoomAllowed_f()` like above

src/cgame/cg_view.c
```diff
@@ -1089,6 +1090,94 @@ static int CG_CalcFov(void)
...
+		                                && (GetWeaponTableData(cg.snap->ps.weapon)->type & WEAPON_TYPE_MG));
+
+		if (!weapzoomAllowed)
+		{
+			cg.weapzoomStartValid = qfalse;
+		}
+		else
+		{
+			float weapzoomTarget = cg_weapzoomFov.value;
+			float f              = 1.0f;
```
Vorschreibung: find a better, more explicit name than just `f`

Vorschreibung (changes requested):

Vorschreibung:
@Aciz do you really insist on `onAdsStart` / `onAdsEnd` hooks?
I'm very against introducing hooks for this prematurely and would wait for enough users to actually request it before adding them.

Oksii:
> @Aciz do you really insist on `onAdsStart` / `onAdsEnd` hooks? I'm very against introducing hooks for this prematurely and would wait for enough users to actually request it before adding them.

I think it makes sense to include from the start. Most/all players with fov toggles currently will want extra customization alongside it. E.g different crosshair sizes/colors, or drawgun on vs off, or might just not want the sens scaling and define a start and end sensitivity. You'd be hard pressed for any player to just have a simple fov bind to switch between cg_fov X and cg_fov Y value only

Vorschreibung:
Hear you, if the above codespots are addressed I think we can merge this then

src/cgame/cg_consolecmds.c
utrolig: Fixed.

src/cgame/cg_view.c
utrolig: Fixed

src/cgame/cg_view.c
utrolig: Fixed

Aciz:
> > @Aciz do you really insist on `onAdsStart` / `onAdsEnd` hooks? I'm very against introducing hooks for this prematurely and would wait for enough users to actually request it before adding them.
> 
> I think it makes sense to include from the start. Most/all players with fov toggles currently will want extra customization alongside it. E.g different crosshair sizes/colors, or drawgun on vs off, or might just not want the sens scaling and define a start and end sensitivity. You'd be hard pressed for any player to just have a simple fov bind to switch between cg_fov X and cg_fov Y value only

Yeah this was my thinking as well, and I already explained earlier why having people rely on scripting on top of this is a bad experience. If the implementation side is an issue for @utrolig, I can add the hooks myself.

I should have time to review and test this over the weekend.

utrolig:
> > > @Aciz do you really insist on `onAdsStart` / `onAdsEnd` hooks? I'm very against introducing hooks for this prematurely and would wait for enough users to actually request it before adding them.
> > 
> > 
> > I think it makes sense to include from the start. Most/all players with fov toggles currently will want extra customization alongside it. E.g different crosshair sizes/colors, or drawgun on vs off, or might just not want the sens scaling and define a start and end sensitivity. You'd be hard pressed for any player to just have a simple fov bind to switch between cg_fov X and cg_fov Y value only
> 
> Yeah this was my thinking as well, and I already explained earlier why having people rely on scripting on top of this is a bad experience. If the implementation side is an issue for @utrolig, I can add the hooks myself.
> 
> I should have time to review and test this over the weekend.

It has already been implemented. It would be good to take a look at it though, as I had some weird hard to reproduce issues with commands in the `cg_onWeapzoomStart/End` get concatenated with other commands like `+left`, but I solved it by adding a newline after the commands in `cg_onWeapzoomStart/End` and I haven't been able to reproduce it after that.

Aciz:
> > > > @Aciz do you really insist on `onAdsStart` / `onAdsEnd` hooks? I'm very against introducing hooks for this prematurely and would wait for enough users to actually request it before adding them.
> > > 
> > > 
> > > I think it makes sense to include from the start. Most/all players with fov toggles currently will want extra customization alongside it. E.g different crosshair sizes/colors, or drawgun on vs off, or might just not want the sens scaling and define a start and end sensitivity. You'd be hard pressed for any player to just have a simple fov bind to switch between cg_fov X and cg_fov Y value only
> > 
> > 
> > Yeah this was my thinking as well, and I already explained earlier why having people rely on scripting on top of this is a bad experience. If the implementation side is an issue for @utrolig, I can add the hooks myself.
> > I should have time to review and test this over the weekend.
> 
> It has already been implemented. It would be good to take a look at it though, as I had some weird hard to reproduce issues with commands in the `cg_onWeapzoomStart/End` get concatenated with other commands like `+left`, but I solved it by adding a newline after the commands in `cg_onWeapzoomStart/End` and I haven't been able to reproduce it after that.

Ah I see :D yeah adding a newline or alternatively `;` after the command in `trap_SendConsoleCommand` is a requirement, as that serves as a splitter for the command parser to differentiate all the separate commands that have been executed during a single frame.

Vorschreibung (approved):
From my side the review is done, LGTM. @utrolig  Thanks for your PR here.
@Aciz  will review it over the weekend and then we can merge it.

Aciz (commented):
Should this reset on death? Currently if you self kill while zooming, you respawn with the zoomed state. It can also be used while wounded and while spectating, which I'm not 100% sure is intentional? Do people use regular FOV scripts in these situations to warrant this working in those scenarios as well?

Oksii:
Probably a non issue on spec or wounded state, people will still have access to their fov toggles during those too typically. 
A reset on death/spawn would feel more polished for sure.

utrolig:
> Should this reset on death? Currently if you self kill while zooming, you respawn with the zoomed state. It can also be used while wounded and while spectating, which I'm not 100% sure is intentional? Do people use regular FOV scripts in these situations to warrant this working in those scenarios as well?

I'll leave this up to you guys, I don't mind either way. I don't think it is intrusive to keep it the way it is, and this only applies 
to using the toggle. I think zooming when spectating or downed doesn't really matter that much, is there a particular reason you would want to disable it in those states?

Aciz:
> > Should this reset on death? Currently if you self kill while zooming, you respawn with the zoomed state. It can also be used while wounded and while spectating, which I'm not 100% sure is intentional? Do people use regular FOV scripts in these situations to warrant this working in those scenarios as well?
> 
> I'll leave this up to you guys, I don't mind either way. I don't think it is intrusive to keep it the way it is, and this only applies to using the toggle. I think zooming when spectating or downed doesn't really matter that much, is there a particular reason you would want to disable it in those states?

I don't mind it, FOV scripts work all the same in those situations so for this to be a proper alternative to them, it should not at least take away properties of them that are not negative in any way. I was just thinking that it's a bit odd that a feature called "weapon zoom" works while not having a weapon, but that's just semantics.

Aciz (approved):
send it

PR #3320, opened 2026-02-21
game: add resolvedMinorSpawnPt information to client session

MijPeter:
Hello,

This PR aims to add information about minor spawnpoint into fireteam in case of a resolved spawnpt.
It also tries to fix some minor inaccuracies in fireateam major spawn display.

I'm not really happy with this PR as code is pretty spaghetti like - especially the part where any spawn blob is searched for (regardles of team classname which for some reason doesnt update when capturing flag).

I've tested this locally and here is example of how it works:
image

(here is viedo link for that test on PR https://streamable.com/okkmiy)

Code ain't pretty but if it works - then maybe it's worth to have this addition included?
Other than what I did it would be also nice to add information about target spawn pt on maps where there are no minor spawn pts mapped.

Here is some other scenario of how this PR works - https://streamable.com/3y6g96

Thanks for reading and sorry that this code is so messy.

### DIFF
--- a/src/game/g_team.c
+++ b/src/game/g_team.c
@@ -455,6 +455,7 @@ void G_clientFlagIndicator(gentity_t *ent)
 gentity_t *SelectRandomTeamSpawnPoint(int teamstate, team_t team, int majorSpawn, int minorSpawn)
 {
 	gentity_t *spot;
+	gentity_t *fallbackSpot;
 	gentity_t *spots[MAX_TEAM_SPAWN_POINTS];
 	int       count, closest, spawnpoint = -1;
 	int       i = 0;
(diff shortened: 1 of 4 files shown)


src/game/g_team.c
```diff
@@ -454,7 +454,7 @@ void G_clientFlagIndicator(gentity_t *ent)
  */
 gentity_t *SelectRandomTeamSpawnPoint(int teamstate, team_t team, int majorSpawn, int minorSpawn)
```
MijPeter, line 455: Changes to this method are not really needed. I just had hard time of understanding logic and once I understood what happens here I decided I could simplify it.

src/game/g_team.c
```diff
@@ -2104,31 +2088,76 @@ qboolean G_desiredFollow(gentity_t *ent, int nTeam)
...
+ * This code is inspired by an actual algorithm used to pick spawn blob.
+ * @param[in] team
+ * @param[in] origin Position to search from.
+ * @return Closest spawn entity, or NULL if none found.
+ */
+static gentity_t *G_FindClosestTeamSpawnEntity(team_t team, vec3_t origin, int minorSpawn) {
+	gentity_t *spot = NULL;
+	gentity_t *closest = NULL;
+	char *classname;
+	float shortest = 1000000, tmp;
```
MijPeter: is there some kind of float_max const that I could reuse?

Aciz: > is there some kind of float_max const that I could reuse?

You could use `MAX_MAP_SIZE * 2`, which would be the maximum length for a single axis in a map.

MijPeter:
Hmm, it doesn't work entirely correctly. Need to fix one thing (when u set unreachable spawnpt but then capture it resolvedminorspawnpt doesnt reset).

MijPeter:
Hey, I've fixed some bugs this PR had. I'd be grateful if anybody could review it.

PR #3330, opened 2026-02-23
game: allow syringe hits to pass through deployables

Vorschreibung:
- Syringe revive traces no longer get blocked by satchels, dynamite or landmines.

- Implementation wise, re-trace if blocking deployables are actually hit until we hit a revivable body or a non-blocking deployable solid.

### DIFF
--- a/src/game/g_weapon.c
+++ b/src/game/g_weapon.c
@@ -431,6 +431,7 @@ gentity_t *Weapon_Syringe(gentity_t *ent)
 	vec3_t    end;
 	trace_t   tr;
 	gentity_t *traceEnt;
+	int       i;
 
 	AngleVectors(ent->client->ps.viewangles, forward, right, up);
 	CalcMuzzlePointForActivate(ent, forward, right, up, muzzleTrace);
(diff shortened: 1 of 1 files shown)


src/game/g_weapon.c
```diff
@@ -439,15 +440,39 @@ gentity_t *Weapon_Syringe(gentity_t *ent)
...
+		if (tr.startsolid)
+		{
+			VectorMA(muzzleTrace, 8, forward, end);
+			trap_Trace(&tr, muzzleTrace, NULL, NULL, end, ent->s.number, MASK_SHOT);
+		}
+
+		if (tr.fraction == 1.0f)
+		{
+			break;
+		}
```
Aciz: This break should probably also be hit if `tr.entityNum` is `ENTITYNUM_WORLD` or `ENTITYNUM_NONE`, so you don't unnecessarily keep iterating over entities if you hit nothing. `tr.fraction` isn't necessarily `1.0` if you hit the world, e.g. if you're aiming to ground/wall.

Vorschreibung: Thanks, fixed!

x0rnn:
I think uips mentioned this issue in regards to corpses - does this fix that also?

Vorschreibung:
> I think uips mentioned this issue in regards to corpses - does this fix that also?

Corpses always were ignored by the hit trace AFAIS

Aciz:
> > I think uips mentioned this issue in regards to corpses - does this fix that also?
> 
> Corpses always were ignored by the hit trace AFAIS

`MASK_SHOT` contains `CONTENTS_CORPSE`, so I don't think they are ignored? I'm not super familiar with the syringe trace setup in general, but maybe worth checking.

Vorschreibung:
> > > I think uips mentioned this issue in regards to corpses - does this fix that also?
> > 
> > 
> > Corpses always were ignored by the hit trace AFAIS
> 
> `MASK_SHOT` contains `CONTENTS_CORPSE`, so I don't think they are ignored? I'm not super familiar with the syringe trace setup in general, but maybe worth checking.

@Aciz `G_TempTraceIgnoreBodies` ignores corpses

Vorschreibung:
Maybe we should ignore all friendly bodies, no matter if alive or corpse; in either case, this is out of scope for this PR.

PR #3341, opened 2026-02-26
cgame: draw custom crosshair smallcross for no-spread weapons

mittermichal:
so when cg_customCrosshair = 3 and cg_customCrosshairDotWidth = 0 player can always see the cross

related conversation: https://discord.com/channels/260750790203932672/260750790203932672/1476652997298880624


Vorschreibung:
don't think that's a good solution, will write more later

mittermichal:
solved in https://github.com/etlegacy/etlegacy/pull/3342

PR #3346, opened 2026-02-27
cgame: change cg_centertime unit from s to ms

Aranud:
Consistencies with other cvar using millisecond for duration unit


Aciz:
This seems like a really bad idea, you're now breaking everyone's configs, because they have the second value archived in `etconfig.cfg`, which means their centerprints will now be basically not visible.

PR #3349, opened 2026-02-27
cgame: fix spawn point location crash in fireteam

Aranud:
`CG_GetLocationMsg` returns a temporary string which is invalid once pass to argument in Q_CleanStr. Do a proper copy of the string to let  Q_CleanStr manipulate it without isssue.


Aciz:
refs #3217

Issue #3028, opened 2025-01-30
missing reset of weaponstats if stopwatch round 2 never occured


Oksii:
player weaponstats of stopwatch round 1 don't get reset if stopwatch round 2 is never played 

steps to reproduce: 
```
callvote map supply
1. complete first round of stopwatch
2. callvote/ref/rcon map supply (restart map/server, possibly any map; did not test)
3. Player will still have stopwatch round1 stats attributed despite map switch/restart. 
4. Ready up; Stopwatch round 1 starts again -> Player still retains previous map stats. They will carry over into Round 2 as normal.
```

x0rnn:
Similar issue: https://github.com/etlegacy/etlegacy/issues/2485, but callvoting matchreset or a different map will reset it there.

x0rnn:
Needs to be reopened.

Issue #2406, opened 2023-11-12
/scores and wstats in warmup doesn't show scores from previous map


x0rnn:
Don't know about VET, but at least in ETPro, doing /scores in warmup would still show the scores and weaponstats of the previous map.

x0rnn:
Probably caused by https://github.com/etlegacy/etlegacy/commit/14d4fab3a81dd47ca187629f40a62cb5b1fd6eaf which fixed #2383

PR #3357, opened 2026-03-01
game: prevent stats from SW round 2 carry over into new match on same map

### LINKED ISSUE #2485: Stats carries over to 1st round of 2nd match on SW
After the 2nd round of the 1st match, if a new match begins on the same map, the stats from the previous round/match will carry over to the 1st round of the 2nd match, and then resets to 0 on the 2nd round of the 2nd match. Happens when the same map is played twice in a row - can be fixed if callvoting matchreset or a different map.

x0rnn:
Fixes #2485, #3028 and #2406. It also allows to see the stats/scores of the previous map while in warmup in campaign mode.

### DIFF
--- a/src/game/g_main.c
+++ b/src/game/g_main.c
@@ -3671,6 +3671,14 @@ void CheckWolfMP(void)
 			{
 				level.warmupTime += 10000;
 				trap_Cvar_Set("g_restarted", "1");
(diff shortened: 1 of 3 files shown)


Aciz:
Could you name this something a bit more descriptive, just for future log reading/bisecting and searchability.

PR #3359, opened 2026-03-01
ui: add "Only humans..." filter option to server browser

FalconNL93:
Added a 4th state to the server browser filter that hides servers with bots currently playing


Aciz:
So how is this actually going to work? A tristate checkbox has 3 visual states, unchecked, yes and no. How is the 4th state visualized here?

Also not sure what you've done here with formatter, seems you've formatted some completely unrelated files?

FalconNL93:
> So how is this actually going to work? A tristate checkbox has 3 visual states, unchecked, yes and no. How is the 4th state visualized here?
> 
> 
> 
> Also not sure what you've done here with formatter, seems you've formatted some completely unrelated files?

I accidentally committed too much files. Just reverted it.

You can click to the 4th state just like it does with the third state. The name is a little bit misleading, i agree. 4th state will show filter "Only humans..." which filters out servers who have bots playing

PR #3373, opened 2026-03-06
cgame: fix riflegrenade switch exploit with cg_weapaltSwitches 0

Vorschreibung:
The previous change in ca86ce23985902d815ec289b8168e67d5a397884 made the attachment/silencer state correction in CG_FinishWeaponChange conditional on cg_weapaltSwitches.

In practice, an engineer could switch away from rifle and back and end up on riflegrenade without the required attach animation, enabling rapid repeated grenade shots.


mittermichal:
seem to work fine, thanks for quick fix!

PR #3379, opened 2026-03-06
cgame: add cg_disableComplaints cvar

Vorschreibung:
Introduce a new archived client cvar, cg_disableComplaints, to disable complaint popups entirely when set to 1.


Vorschreibung:
in HUD editor, votetext has a style flag to disable complaint popups

PR #3385, opened 2026-03-10
Add Iconfeed

Vorschreibung:


src/cgame/cg_hud_iconfeed.c
```diff
@@ -0,0 +1,397 @@
...
+
+static const float fade_in_size      = 2.0f;
+static const int   fade_in_duration  = 200;
+static const float fade_in_alpha     = 0.01f;
+static const int   fade_out_duration = 300;
+static const float fade_out_alpha    = 0.0f;
+
+/* Negative values intentionally tighten the row by slightly overlapping icons. */
+static const int icon_gap            = -2;
+static const int icon_shift_duration = 200;
```
Aranud: Define or other static variables are upper case in code. Would be great to keep the same formatting for consistencies.

Aranud (commented):
Single line comment use /* */ instead of //. Not really an issue, but could be changed for consistencies with other code parts.

src/cgame/cg_hud_iconfeed.c
Vorschreibung: Thanks, done!

PR #3412, opened 2026-03-24
ui: fix linking on non-Linux platforms

Aciz:
`ui_menu.c` is included in cgame as well, unlike `ui_main.c`, so calling functions directly that live there from `ui_menu.c` will fail to link on most platforms. Call via function pointer in `displayContextDef_t` instead so that everything links correctly.

refs #3406


Vorschreibung:
Not needed anymore

PR #3419, opened 2026-03-29
game: Add fear kill mechanics 

Aranud:
On suicide or team switch, if the player has been hit by an enemy player which is still alive in the last 1500 ms, reward the enemy player with the kill and reward kill assist to other players as well.


Oksii:
Can this be a server toggle? Selfkill mechanics are used wildly differently in competitive play and this would unnecessarily reward players with stats. Additionally it would be a breaking change for us in stat collecting/parsing too when comparing historical results - which are actively used for player skill calculations.

x0rnn:
Also, 1500ms is too much. We added this on Hirntot via lua and decided on 400ms and also added an extra check if the next respawn is in more than 5 seconds.

Aranud:
> Can this be a server toggle? Selfkill mechanics are used wildly differently in competitive play and this would unnecessarily reward players with stats. Additionally it would be a breaking change for us in stat collecting/parsing too when comparing historical results - which are actively used for player skill calculations.

It can yes. How does the computation could be a breaking change in your case ? You can check if the kill is a MOD_FEAR death.

> Also, 1500ms is too much. We added this on Hirntot via lua and decided on 400ms and also added an extra check if the next respawn is in more than 5 seconds.

I don't mind changing the value, but could you explain how 1500ms is too much ?

I like the idea to the check if player is close from next respawn ... but 5 seconds seems too much. Any explanation why not waiting the perfect tap close to 1 second ?

x0rnn:
1.5s proved to be too much because you'd get too many false positives, e.g. you would hide behind a wall, enemy could die meanwhile, and you'd still get "scared to death". We lowered it to 400ms to only count actual in-fight scenarios.

5s could be lowered I guess.

u1ps:
> Also, 1500ms is too much. We added this on Hirntot via lua and decided on 400ms and also added an extra check if the next respawn is in more than 5 seconds.

Do i get it right selfkilling 5secs before your next spawn would not reward kill? There should not be any exception even if you selkfill 0.5 secs before spawn. This is exactly for what this feature is. Consistently reward kill to player who experiences enemy magically dying in middle of fight.

x0rnn:
Yea, but it still caused too many false positives. You'd just selfkill because your respawn was coming up, but you ended rewarding the enemy for not even doing an in-fight selfkill because you were still inside that time window.

Maybe limiting it to 300ms would be better without any check for spawntimes.

u1ps:
> Yea, but it still caused too many false positives. You'd just selfkill because your respawn was coming up, but you ended rewarding the enemy for not even doing an in-fight selfkill because you were still inside that time window.

Thats exactly how it should work. PLayers choosing to optimize spawn will reward kill even if they get hit by 1 bullet. From perspective of other player shooting they refuse to stay and fight

Aranud:
Your vision are opposite on this point. 
I quote Oksii (explanation from discord) which give a better understanding how the comp vision is : 

> It'd generally be considered a bad play, poor awareness and bad positioning in comp if you're fighting a person who you know is just pushing to selfkill and respawn - wasting bullets and potentially risking life - for nothing in return. Of course there's exceptions to the rules when it comes to gaining map control, but in practice 9/10 times it's a bad play to be engaging with these players, and even worse rewarding them positively for it would send the absolute worst message/encourage play for cheap kills on selfkillers

Let have a perceptive from attacker who suicide against enemy on defense. 
Currently, suicide player get an edge when re spawning in time by wasting enemy player bullets but get no reward (important to note the no reward).
In contrast, even if the player get killed before he suicide, it would not change the fact it will get an edge.

The only case where it could be a changer is the exact moment where attacker kill the defender and then doesn't suicide because he his in good position. This is a precise timing and discouraging it doesn't seem fair.

The pressure is on the defender to do properly his job and ensure the attacker force his respawn without gaining an edge on the field. Defender are closer to obj than attacker and they was suicide as well. Timing is the key.

If punishing early suicide is enough to not blame too much correct suicide timing, could it fit to everyone or is still too much from a pub perspective ?

Other aspect, we may reward correct suicide timing ... but it sound weird.

x0rnn:
> If punishing early suicide is enough to not blame too much correct suicide timing, could it fit to everyone or is still too much from a pub perspective ?

Can you explain better what you mean by this?

src/game/g_cmds.c
```diff
@@ -1592,7 +1614,16 @@ qboolean SetTeam(gentity_t *ent, const char *s, qboolean force, weapon_t w1, wea
 			ent->flags                        &= ~FL_GODMODE;
 			ent->client->ps.stats[STAT_HEALTH] = ent->health = 0;
 			ent->sound1to2                     = HR_NUM_HITREGIONS;           // not a headshot kill
-			player_die(ent, ent, ent, 100000, MOD_SWITCHTEAM);
+
+			if (G_FearCheck(ent))
+			{
+				player_die(ent, &g_entities[ent->client->lasthurt_client], &g_entities[ent->client->lasthurt_client],
+				           100000, MOD_FEAR);
+			}
```
Aciz, line 1622: Does this actually make sense here? I don't think players use team switches the same way as they use selfkills to avoid getting killed by opponents/missing spawn. This might also cause weird `MOD_FEAR` frags with things like auto-balancers, assuming their functionality runs through this code for switching players' team.

Aciz (commented):
Should the fear check be performed on client disconnect as well?

x0rnn:
Probably shouldn't. Shouldn't happen on team switch either.

PR #3420, opened 2026-03-29
qcommon: decouple mod localization from engine traps

Vorschreibung:
Move the tinygettext localization backend into a shared qcommon path that now builds into the engine and the client-side mod libraries.

The new backend uses the same dictionary logic for client and mod catalogs, keeps the legacy engine translation entry points intact, and removes the old default-mod gating from filesystem startup.

Switch cgame and ui to initialize and query localization directly through qcommon instead of relying on translation syscalls. That keeps the mod side compatible with engines that only provide the baseline cvar and filesystem traps, including 2.60b-compatible setups.

Update the build so FEATURE_GETTEXT also pulls the qcommon localization sources and tinygettext into cgame and ui, including the SDL-backed tinygettext path where needed.

### DIFF
--- a/cmake/ETLPlatform.cmake
+++ b/cmake/ETLPlatform.cmake
@@ -168,6 +168,16 @@ if(UNIX)
 			find_library(force_feedback_library ForceFeedback REQUIRED)
 			find_library(opengl_library OpenGL REQUIRED)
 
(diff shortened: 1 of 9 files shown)


Vorschreibung:
Makes `cgame`/`ui` to link against SDL; no way

PR #3431, opened 2026-04-12
cgame: add cg_allowSelfKillSpawnProtection cvar, fixes #3321

### LINKED ISSUE #3321: Allow `/kill` during spawn protection or ensure first wave exceeds protection window
When the first respawn wave happens in under 3 seconds, players cannot use `/kill` to self‑kill for ammo because spawn protection blocks `/kill` during invulnerability. See https://github.com/etlegacy/etlegacy/blob/master/src/game/g_cmds.c#L1267.

Players should be able to `/kill` after the first wave if they choose, or the first wave should not be shorter than the protection window. This is especially crucial for competetive scene to allow players to get extra 30 rounds of ammunition at the start.

Proposed possible fixes:
1) Config: allow `/kill` during spawn protection. Add a server cvar that allows `/kill` even while spawn protection is active.

2) Config: adjust self‑kill protection time. Add a server cvar that defines how long `/kill` is blocked after spawn (e.g., allow `/kill` after 1s even if invulnerability lasts 3s).

3) Adjust first reinforcement wave. Ensure the first wave spawn time is always **greater than or equal to** the spawn protection window (e.g., if protection is 3000 ms, first wave must be >3000 ms).

Cultti:
This PR moves PW_INVULNERABLE self kill check to client side and adds `cg_allowSelfKillSpawnProtection` cvar that allows clients to self kill during protection if they choose to do so.

### DIFF
--- a/src/cgame/cg_consolecmds.c
+++ b/src/cgame/cg_consolecmds.c
@@ -1290,6 +1290,25 @@ void CG_ForceTapOut_f(void)
 }
 
 /**
+ * @brief Client-side handler for the /kill command.
+ */
+static void CG_Kill_f(void)
+{
+	if (!cg.snap)
+	{
+		return;
+	}
+
+	if (cgs.gamestate == GS_PLAYING && !cg_allowSelfKillSpawnProtection.integer && cg.spawnInvulnerability && cg.snap->ps.powerups[PW_INVULNERABLE] > cg.time)
+	{
+		CG_CenterPrint("You are invulnerable - ^3/kill^7 is disabled.");
+		return;
+	}
+
+	trap_SendClientCommand("kill");
+}
+
+/**
  * @brief Shows a popup message.
  */
 static void CG_CPM_f(void)
@@ -3577,6 +3596,7 @@ static consoleCommand_t commands[] =
 	{ "undoSpeaker",            CG_UndoSpeaker_f             },
 	{ "cpm",                    CG_CPM_f                     },
 	{ "forcetapout",            CG_ForceTapOut_f             },
+	{ "kill",                   CG_Kill_f                    },
 	{ "timerSet",               CG_TimerSet_f                },
 	{ "timerReset",             CG_TimerReset_f              },
 	{ "resetTimer",             CG_TimerReset_f              }, // keep ETPro compatibility
@@ -3703,7 +3723,6 @@ static const char *gameCommand[] =
 	"imwa",
 	"imws",
 	//   "invite",
-	"kill",
 	"lock",
 	"mapvote",
 #ifdef FEATURE_MULTIVIEW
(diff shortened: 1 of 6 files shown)


Cultti:
Actually this does prevent self also now when revived which wasn't intended behaviour. Will fix.

src/cgame/cg_consolecmds.c
```diff
@@ -1289,6 +1289,25 @@ void CG_ForceTapOut_f(void)
...
+ * @brief Client-side handler for the /kill command.
+ */
+static void CG_Kill_f(void)
+{
+	if (!cg.snap)
+	{
+		return;
+	}
+
+	if (!cg_allowSelfKillSpawnProtection.integer && cg.spawnInvulnerability && cg.snap->ps.powerups[PW_INVULNERABLE] > cg.time)
```
Aciz: server side check also checked for cheats here (can check on client with `cgs.sv_cheats`), probably should be included here for same functionality

Cultti:
Okay now it should work as previously:
- Allows self kill during protection while gamestate not in GS_PLAYING (eg. warmup) regardless of the setting
- Doesn't prevent self kills when player was revived

src/cgame/cg_consolecmds.c
```diff
@@ -1289,6 +1289,25 @@ void CG_ForceTapOut_f(void)
 	trap_SendClientCommand("forcetapout");
 }
 
+/**
+ * @brief Client-side handler for the /kill command.
+ */
+static void CG_Kill_f(void)
+{
+	if (!cg.snap)
```
Aciz: this should probably also early exit on demo playback

could also check for `cg.snap->ps.pm_flags & PMF_FOLLOW` so you can't `/kill` when following someone

src/cgame/cg_consolecmds.c
Cultti: Yeah this PR changes the behaviour so that players can decide if they want to use this self kill protection system or not. Previously it was mantaded by the server with the `sv_cheats` cvar. See #3321 discussion.

I kept the original protection enabled for clients and now they just can opt out from it. So in theory it should function as previously if client doesn't change this setting.

src/cgame/cg_consolecmds.c
Cultti: Changed statement to
```
if (cg.demoPlayback || !cg.snap || (cg.snap->ps.pm_flags & PMF_FOLLOW))
{
	return;
}
```

src/cgame/cg_consolecmds.c
Cultti: Or do you mean that if sv_cheats is enabled we should allow self kills like we allow when setting `cg_allowSelfKillSpawnProtection` to 1?

Aciz: oh yeah nvm ignore that, doesn't matter now that you can just turn it off at will

Aranud:
Seems good, work as intended. Anything to add before merging ?

Cultti:
I think this is as good as it can get. Go for it. 👍

PR #3432, opened 2026-04-17
sv_tracker: support multiple endpoints via semicolon-separated list 

wahke:
Allows sv_tracker to accept a semicolon-separated list of addresses,
so a server can report stats to more than one tracker endpoint at
once (for example the official ETL tracker + a community tracker
like wolffiles.eu).

A single address still works exactly as before, no change for
existing configs.

Example:
  set sv_tracker "et-tracker.trackbase.net:4444;tracker.wolffiles.eu:4444"

Implementation:
- Tracker_Init() parses the semicolon list and resolves each endpoint
- Tracker_Send() dispatches to all resolved endpoints
- Tracker_Frame() checks the modified flag on sv_tracker and reinits
  when the cvar changes. Tracker_Init runs before etl_server.cfg is
  executed, so without this check values from the cfg would be
  ignored and the default kept. Now updates from etl_server.cfg,
  rcon or console take effect live.

CVAR_PROTECTED stays on sv_tracker, no flag changes.

Tested on one of our live servers. Log excerpt:

  0   Tracker: Server communication enabled (1 endpoint(s).)
  100 execing etl_server.cfg
  350 Tracker: sv_tracker changed, reinitializing...
  350 Resolving et-tracker.trackbase.net:4444
  350 et-tracker.trackbase.net:4444 resolved to 91.220.53.120:4444
  350 Resolving tracker.wolffiles.eu:4444
  350 tracker.wolffiles.eu:4444 resolved to 138.201.36.184:4444
  350 Tracker: Server communication enabled (2 endpoint(s).)

The jump from 1 to 2 endpoints is the moment etl_server.cfg loads
and the reinit picks up the new value. Also works with runtime
rcon set.


wahke:
## Why this is useful

Right now sv_tracker only accepts one address, so admins have 
to pick between reporting to the official ETL tracker or a 
community one. Most pick the official one, which is the right 
call but also means community analytics projects don't really 
get off the ground.

wolffiles.eu runs a community tracker at 
tracker.wolffiles.eu:4444 that consumes the same UDP stat 
packets ETL already emits. It renders player profiles, ELO, 
session history, weapon breakdowns, activity heatmaps, etc:

player stats

With this PR an admin can feed something like this alongside 
the official tracker using one cvar, no patched binaries or 
either/or decision needed.

## How it works

sv_tracker now accepts a semicolon-separated list of addresses. 
A single address still works exactly as before, the split only 
kicks in when the string contains a semicolon. sv_tracker stays 
CVAR_PROTECTED, no flag changes.

Example:
`set sv_tracker "et-tracker.trackbase.net:4444;tracker.wolffiles.eu:4444"`

## Cvar update handling

Tracker_Init() runs before the server config is executed, so a 
value from etl_server.cfg was previously ignored and the tracker 
kept the default. Tracker_Frame now checks the modified flag on 
sv_tracker and reinits when it changes, so updates from 
etl_server.cfg, rcon or the console all take effect live.

Log excerpt from one of our servers:

```
0   Tracker: Server communication enabled (1 endpoint(s).)
100 execing etl_server.cfg
350 Tracker: sv_tracker changed, reinitializing...
350 Resolving et-tracker.trackbase.net:4444
350 et-tracker.trackbase.net:4444 resolved to 91.220.53.120:4444
350 Resolving tracker.wolffiles.eu:4444
350 tracker.wolffiles.eu:4444 resolved to 138.201.36.184:4444
350 Tracker: Server communication enabled (2 endpoint(s).)
```

The jump from 1 to 2 endpoints is the moment etl_server.cfg is 
loaded and the reinit picks up the new value. Also works with 
runtime `rcon set sv_tracker ...`.

wahke (commented):
reworked to handle cvar updates. added a modified flag check in Tracker_Frame that reinits the tracker when sv_tracker changes, so values from etl_server.cfg are picked up on the next frame instead of being ignored in favor of the default.

tested on one of our servers, log excerpt:

0 Tracker: Server communication enabled (1 endpoint(s).)
100 execing etl_server.cfg
350 Tracker: sv_tracker changed, reinitializing...
350 Resolving et-tracker.trackbase.net:4444
350 et-tracker.trackbase.net:4444 resolved to 91.220.53.120:4444
350 Resolving tracker.wolffiles.eu:4444
350 tracker.wolffiles.eu:4444 resolved to 138.201.36.184:4444
350 Tracker: Server communication enabled (2 endpoint(s).)
also works with runtime rcon set.

src/server/sv_tracker.c
```diff
@@ -54,12 +54,98 @@ enum
...
+	           trackerAddrs[numTrackerAddrs].ip[3],
+	           BigShort(trackerAddrs[numTrackerAddrs].port));
+	numTrackerAddrs++;
+}
+
+/**
+ * @brief Parse a ';'-separated list of addresses and resolve each entry
+ * @param[in] list Raw list from sv_tracker cvar
+ */
+static void Tracker_ParseAddressList(const char *list)
```
Aranud, line 112: You should change the separator to use space which is the used string separator for cvar and reuse the existing function COM_ParseExt function. Like this, you are consistent with the existing code and cvar doesn't need to handle a different way to split string.

src/server/sv_tracker.c
```diff
@@ -54,12 +54,98 @@ enum
 } catchBot;
 qboolean catchBotNum = 0;
 
-netadr_t addr;
+#define MAX_TRACKERS 8
```
Aranud, line 66: If you track up to 8 different address, be aware that the cvar can only hold a string of 256 chars MAX. Considers using shorter URL address will help holding 8 different servers if this count is reached one day.

Aranud (changes requested):

src/server/sv_tracker.c
wahke: Done, switched to COM_ParseExt with whitespace as separator. Custom ; parser is gone — cvar now looks like sv_tracker "host1:4444 host2:4444" which feels much more in line with the rest of the engine. Builds clean locally

src/server/sv_tracker.c
wahke: Good catch, hadn't thought about that. Added a doxygen note on MAX_TRACKERS mentioning the 256 char cvar bound so anyone setting this up knows to keep hostnames short. Thanks!

PR #3434, opened 2026-04-23
etltv: add tv zone for tv server queue memory allocations, refs #229

ryzyk-krzysiek:
- zone is added only to dedicated server and if server is delayed (`sv_etltv_delay` is set)
  zone minimum default size is 24MB
  added cvar `sv_etltv_zoneMegs` for increasing the default size (setable only in commandline)
  1 player per 1 minute of delay uses around 0.2MB, so for example 12 players with 5 minute delay should use around 12-13MB (tested only on bots)
  added debugging info to `meminfo` and `zonelog` command accordingly
- rename member variable
- set client entity to tv server entity - fixes a bug where if the entities on master server are bots then players joining tv server would inherit `SVF_BOT` flag if they connect on the same server slot as bot from master server, which would make tv server think they are a bot and never send a snapshot to such client

refs #229

### DIFF
--- a/src/server/sv_init.c
+++ b/src/server/sv_init.c
@@ -1237,6 +1237,7 @@ void SV_Init(void)
 	sv_etltv_autoplay   = Cvar_Get("sv_etltv_autoplay", "0", CVAR_ARCHIVE_ND);
 	sv_etltv_clientname = Cvar_GetAndDescribe("sv_etltv_clientname", "ETLTV", CVAR_ARCHIVE_ND, "Name of the ETLTV client.");
 	sv_etltv_delay      = Cvar_GetAndDescribe("sv_etltv_delay", "0", CVAR_INIT, "Delay feed by number of seconds.");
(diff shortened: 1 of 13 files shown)


src/server/sv_init.c
```diff
@@ -1237,6 +1237,7 @@ void SV_Init(void)
 	sv_etltv_autoplay   = Cvar_Get("sv_etltv_autoplay", "0", CVAR_ARCHIVE_ND);
 	sv_etltv_clientname = Cvar_GetAndDescribe("sv_etltv_clientname", "ETLTV", CVAR_ARCHIVE_ND, "Name of the ETLTV client.");
 	sv_etltv_delay      = Cvar_GetAndDescribe("sv_etltv_delay", "0", CVAR_INIT, "Delay feed by number of seconds.");
+	sv_etltv_zoneMegs   = Cvar_GetAndDescribe("sv_etltv_zoneMegs", DEF_TVZONEMEGS_S, CVAR_INIT, "TV server queue memory zone. (in MB)");
```
Aciz: cvar description here could maybe have a brief note on why adjusting this might be necessary? Otherwise this just seems like a copy of `com_zoneMegs` for tv server for the end user without any justification for it's existence

ryzyk-krzysiek: Updated the cvar description, hope it is more clear now.

PR #3437, opened 2026-04-24
fix: prevent restricted weapon slots from ghost-blocking across map changes (#3375)

### LINKED ISSUE #3375: Latched class might persist across maps
As reported by Oksii on discord:

>[12:56]Oksii: We had the oddest bug yesterday. Not sure how or if reproducible at all. 
A player effectively was "latched" as class permanently through out map changes, map restarts, config reloads. Only until after the player swapped to spectator did it fix itself. 
>
>He selected panzer in a previous game, the match was surrendered, and despite not being panzer or any soldier class, being medic: panzer became unavailable for anyone else to use. Other soldier weapons (e.g flamer, mg) were available to anyone. 
>
>Server config used: https://github.com/Oksii/legacy-configs/blob/main/configs/legacy6.config
>[12:58]sonNy: IIRC you can fix it by having that person taking panzer again and then switching back to other class?
>[12:58]sonNy: At least I saw h2o do that fix himself
>[12:58]Oksii: Maybe. We all tried swapping to panzer/soldier, we tried restarting the map, we reloaded server cfg, >nothing resolved it until whole team went spec
>[12:59]Oksii: It persisted through out 3 maps
>[12:59]Oksii: Even just a map change should clear it, given that he was on a different class entirely for the other 2 maps

RngesusSolutions:
When a player switched weapons in limbo (e.g. Soldier Panzer -> Medic mp40/thompson), the
  stale playerWeapon field was never cleared because it is only synced
  from latchPlayerWeapon at spawn time in ClientSpawn. G_TeamCount()
  checked both fields unconditionally, so the old weapon permanently
  occupied a restricted slot  surviving map rotations via session
  persistence.

  - g_cmds.c: G_TeamCount() now only considers latchPlayerWeapon for
    dead/limbo/spectator players, ignoring the stale playerWeapon
  - g_session.c: G_WriteClientSessionData() writes latchPlayerWeapon
    into the playerWeapon session field for dead/limbo players so
    stale values are never persisted to disk

fixes https://github.com/etlegacy/etlegacy/issues/3375

  1. Player picks Panzer then both fields = Panzer
  2. Player dies, switches to Medic in limbo -> latchPlayerWeapon = MP40, but playerWeapon stays Panzer
  3. Map change -> session saves stale playerWeapon = Panzer to disk and restores it
  4. G_TeamCount() sees stale Panzer = slot blocked forever


src/game/g_cmds.c
```diff
@@ -1918,9 +1918,27 @@ int G_TeamCount(gentity_t *ent, int weap)
 
 		if (weap != -1)
 		{
-			if (level.clients[j].sess.playerWeapon != weap && level.clients[j].sess.latchPlayerWeapon != weap)
+			// For alive players (not in limbo), check the weapon they are
+			// actively carrying. For dead/limbo players, only check the
+			// weapon they intend to spawn with. Previously both fields were
+			// checked unconditionally, which caused a stale playerWeapon
+			// (only synced at spawn time in ClientSpawn) to permanently
+			// "latch" a weapon slot across map changes — see issue #3375.
```
Aranud: Remove "— see issue #3375" as it can be deprecated if migrated to another ticketing tool. Git commit and message reference are better place to hold this information.

src/game/g_session.c
```diff
@@ -76,8 +76,22 @@ void G_WriteClientSessionData(gclient_t *client, qboolean restart)
 	cJSON_AddNumberToObject(root, "spectatorClient", client->sess.spectatorClient);
 	cJSON_AddNumberToObject(root, "userSpectatorClient", client->sess.userSpectatorClient);
 	cJSON_AddNumberToObject(root, "playerType", client->sess.playerType);
-	cJSON_AddNumberToObject(root, "playerWeapon", client->sess.playerWeapon);
-	cJSON_AddNumberToObject(root, "playerWeapon2", client->sess.playerWeapon2);
+
+	// If the player is in limbo or dead, their playerWeapon is stale (only
+	// synced from latchPlayerWeapon at spawn time in ClientSpawn). Persist the
+	// latched values so that on session restore the stale active weapon doesn't
+	// ghost-occupy a restricted weapon slot. See issue #3375.
```
Aranud: Remove "— see issue #3375" as it can be deprecated if migrated to another ticketing tool. Git commit and message reference are better place to hold this information.

src/game/g_session.c
```diff
@@ -76,8 +76,22 @@ void G_WriteClientSessionData(gclient_t *client, qboolean restart)
 	cJSON_AddNumberToObject(root, "spectatorClient", client->sess.spectatorClient);
 	cJSON_AddNumberToObject(root, "userSpectatorClient", client->sess.userSpectatorClient);
 	cJSON_AddNumberToObject(root, "playerType", client->sess.playerType);
```
Aranud, line 78: What about the playerType when latched ? Doesn't it follow the same behaviour as weapons ?

Aranud (commented):
Added a few comments

PR #3448, opened 2026-05-06
cgame: fix duplicate selfkill killfeed icons

Vorschreibung:
Move the self-kill icon insertion to the obituary event scope so it runs once per death event instead of potentially multiple times for each visible popupmessages component.


Aciz:
refs #3385

PR #3472, opened 2026-05-31
Improve compatibility on modern OpenBSD, NetBSD and FreeBSD

renaudallard:
cmake: link libexecinfo on the BSDs for backtrace support

That will be sufficient to compile fully on OpenBSD, but not enough for the others.


Vorschreibung:
Thanks!

PR #3475, opened 2026-06-02
refactor: `net_uint8_t` for DirToByte, AddEvent

WofWca:
The point of `DirToByte` is to use it for `eventParm`,
which is `net_uint8_t`.

I didn't try to compile this, only ran my linter.

### DIFF
--- a/src/game/g_local.h
+++ b/src/game/g_local.h
@@ -1577,7 +1577,7 @@ void G_ClientSound(gentity_t *ent, int soundIndex);
 void G_TouchTriggers(gentity_t *ent);
 
 void G_AddPredictableEvent(gentity_t *ent, int event, int eventParm);
-void G_AddEvent(gentity_t *ent, int event, int eventParm);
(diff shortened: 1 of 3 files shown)


Aciz (approved):
FWIW, `net_uint8_t` is just a typedef for `int`, the point of is to indicate how many bits (and whether it's unsigned or not) of the field is networked. `entityState_t` and `playerState_t` contain variables that are networked, but while the underlying type is `int`, they might not consume the full 4 bytes on the wire, hence the typedefs to indicate how much of it is actually networked, and how much data it can hold.

WofWca:
Yep, and the same holds for the return type of `DirToByte`.

But yes, this commit proliferates this type beyond the net code, maybe I'm not seeing downsides to that.

Aciz:
> Yep, and the same holds for the return type of `DirToByte`.
> 
> But yes, this commit proliferates this type beyond the net code, maybe I'm not seeing downsides to that.

Yeah no this is a good change, it makes it more clear for the caller how many bits are available for the parameter.

PR #3480, opened 2026-06-26
client: add Discord Rich Presence (Windows, FEATURE_DISCORD)

jimmymeister98:
Add an optional Discord Rich Presence wrapper for the Windows client, gated behind a new FEATURE_DISCORD CMake option.

FEATURE_DISCORD defaults to ON for MSVC builds and OFF elsewhere. When the feature is disabled, or when building on non-MSVC platforms, no Discord SDK code is compiled or linked.

## Short list of changes:
- Added Discord Rich Presence for MSVC builds (others have no-ops for the added cvars -> can be added later) 
- Added checkbox under network settings to enable/disable DRP IPC
  - Disabling tears down the presence handler so there should be no performance hit
  - Rich Presence polls every 15s and enriches the current active discord session with "ET Legacy, current map, current team"

## Motivation

Discord Rich Presence makes it easier for players to show that they are playing ET Legacy and can improve project and game visibility through Discord’s activity status.

## Details
Wrapper (src/discord/cl_discord.c/.h):
- CL_DiscordInit/Frame/Shutdown, reading player state/team/map/server from cls/cl. RunCallbacks every frame; presence throttled to 15s (cls.realtime +1 idiom); startTimestamp via time(NULL) on the CA_ACTIVE transition (polling prevClsState, no edits to CL_Disconnect/CL_FirstSnapshot/CL_Connect_f).
- CA_CINEMATIC excluded from the connected gate (shows "In Menu"); team read only at CA_ACTIVE && cl.snap.valid.
- Live opt-out via cl_discordPresence (CVAR_ARCHIVE, 0-1): 1->0 full teardown (ClearPresence+Shutdown), 0->1 re-init. cl_discordAppId (CVAR_ARCHIVE) placeholder; empty -> no-op (logs once).
- Event handlers log once; joinGame/spectateGame/joinRequest NULL; autoRegister=0; no DISCORD_DISABLE_IO_THREAD.

Build:
- Vendored Discord RPC SDK under src/discord/win32-static/ (static lib, no DLL). discord_log2_shim.c provides log2() for the /MD lib under the engine's /MT static CRT (FORCE_STATIC_VCRT); MSVC-only, same gate.
- CMake define + source-append + lib link + include dir share one if(MSVC AND FEATURE_DISCORD) gate in ETLSetupFeatures.cmake.

Wiring:
- CL_DiscordInit after CL_InitCvars (CL_Init); CL_DiscordFrame after Con_RunConsole, before cls.framecount++ (CL_Frame); CL_DiscordShutdown before Com_Memset(&cls,0,...) (CL_Shutdown) — each under #ifdef FEATURE_DISCORD; include guard in client.h.
- In-game System -> NETWORK checkbox bound to cl_discordPresence.

Tested on Win 11 & macOS Tahoe 26.5 (apple metal)
The Unix style EOL and the uncrustify steps of the Coding Convention were considered.
The Discord App-ID is currently a placeholder and should be replaced. Feel free to create a new Discord developer application (If you need some help in any way i'll be happy to help). The ID is found in Line 288 in cl_cvars.c
If someone wants to add osx and linux support this is the library that was used https://github.com/discord/discord-rpc

Screenshot 2026-06-26 234355
Screenshot 2026-06-26 234323


jimmymeister98:
The two files that fail the "pre-build" action because of them not being uncrustified are from the discord-sdk so they shouldnt be uncrustified.
The SonarCloud Code Analysis Quality seem to refer to old problems predating my changes.

ensiform:
You can get binaries for x64 Linux and macos too, problem also arm64. You'd have to build new bundled version for x86 Linux support as discord never provided a build for that

jimmymeister98:
Okay, i've built the macos binary from the discord-rpc repo. 
The Discord rich-presence should now work for both intel x86_64 and arm arch on apple metal.

jimmymeister98:
Rich presence works on i386 32-bit builds on linux (validated via vm). Also added a compat patch for Discord installed via snap.

64-Bit build wont start after building (also on main) and crashes with "Setting relative mouse location fails (system not supported)" 

System: Ubuntu 25.04 (Plucky Puffin)

PR #3485, opened 2026-07-01
Fix server lag spike on flag/objective capture

RngesusSolutions:
G_UpdateSpawnPointStatePlayerCounts() was calling ClientUserinfoChanged() for every client whose resolved spawn point changed. On an objective or flag capture, that can be most of the server in a single frame. Each call rebuilds and reliably broadcasts the full CS_PLAYERS configstring to all clients, creating a burst of reliable messages in one tick.

In practice, this showed up as a server hitch that got worse as player count increased. This was a regression from the fireteam spawnpoint feature (upstream cec44f70, PR #3155); 2.82.0 only performed an assignment here.

Instead, changed clients are now marked with pers.spawnInfoChangePending and drained from G_RunFrame() through G_FlushSpawnPointInfoChanges(). That path updates only the 'sp' and 'msp' keys in the existing CS_PLAYERS configstring and limits the work to 8 clients per frame.

The keys and values written here match what ClientUserinfoChanged() writes, so existing 2.84 clients continue to support fireteam overlay spawn numbers without any client-side changes. ClientUserinfoChanged() also clears the pending flag, since its full rebuild already includes the current values.

This also initializes sess.previousUserMinorSpawnPointValue to -1 in both session init paths, G_InitGame and G_InitSessionData. Previously it was left uninitialized, which caused the first player-count pass to flag a false update for every client.

### DIFF
--- a/src/game/g_client.c
+++ b/src/game/g_client.c
@@ -2161,6 +2161,10 @@ void ClientUserinfoChanged(int clientNum)
 	}
 #endif
 
+	// the full CS_PLAYERS rebuild below includes the current 'sp'/'msp' values,
+	// so any pending targeted spawn info update (G_FlushSpawnPointInfoChanges) is supercession
+	client->pers.spawnInfoChangePending = qfalse;
+
 	trap_GetConfigstring(CS_PLAYERS + clientNum, oldname, sizeof(oldname));
 	trap_SetConfigstring(CS_PLAYERS + clientNum, configStr);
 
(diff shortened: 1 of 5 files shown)


src/game/g_client.c
```diff
@@ -2179,6 +2183,85 @@ void ClientUserinfoChanged(int clientNum)
...
+		}
+
+		client->pers.spawnInfoChangePending = qfalse;
+
+		// not fully connected yet: ClientBegin -> ClientUserinfoChanged writes
+		// the full configstring (including sp/msp) anyway
+		if (client->pers.connected != CON_CONNECTED)
+		{
+			continue;
+		}
```
Aciz, line 2244: This is just redundant, no? If a player has not connected yet, they can't have a pending spawn point change anyway, so you've already early exited at this point?

src/game/g_client.c
```diff
@@ -2179,6 +2183,85 @@ void ClientUserinfoChanged(int clientNum)
...
+
+		clientNum = client - level.clients;
+
+		trap_GetConfigstring(CS_PLAYERS + clientNum, cs, sizeof(cs));
+
+		if (!cs[0])
+		{
+			continue;
+		}
```
Aciz, line 2253: ~~Same for this, you're using `level.sortedClients` to get the client, afaik that never contains a client that would have empty `CS_PLAYERS`.~~

Edit: or more specifically, a client which would have a queued spawnpoint update would not have an empty configstring

src/game/g_team.c
```diff
@@ -2401,7 +2401,13 @@ void G_UpdateSpawnPointStatePlayerCounts()
 		{
 			client->sess.resolvedSpawnPointIndex      = resolvedSpawn.major;
 			client->sess.resolvedMinorSpawnPointIndex = resolvedMinorSpawnPt;
-			ClientUserinfoChanged(client - level.clients);
+			// don't call ClientUserinfoChanged() here: on objective/flag capture the resolved
+			// spawn point changes for most of the server at once, and rebuilding and broadcasting
+			// the full CS_PLAYERS configstring for every client in a single frame causes a
+			// noticeable server lag spike (reliable messages in thousands). Instead flag the
+			// client and let G_FlushSpawnPointInfoChanges() (G_RunFrame) push a targeted,
+			// rate-limited update of just the 'sp'/'msp' keys.
```
Aciz, line 2409: >Instead flag the client and let G_FlushSpawnPointInfoChanges() (G_RunFrame) push a targeted, rate-limited update of just the 'sp'/'msp' keys

? you're not pushing an update of "just sp/msp keys", you're pushing a full `CS_PLAYERS` update?

src/game/g_client.c
RngesusSolutions: Not quite sortedclients also is containing con_connecting clients (anyone still loading after a map change has their team restored from sess data), and G_UpdateSpawnPointStatePlayerCounts() iterates that same list filtering only by team so they can get flagged while connecting. Skipping them here just avoiding a wasted update since ClientBegin -> ClientUserinfoChanged rebuilds the full string anyway. I can modify the comment to make more clear?

src/game/g_client.c
RngesusSolutions: You're right cs_players cant be empty here this can be removed

src/game/g_team.c
RngesusSolutions: The full cs_players configstring still gets broadcast; what we limit is the mutation to the existing string’s sp/msp keys. The real savings come from the per frame rate limit, avoiding the full userinfo rebuild and letting the engine skip the broadcast entirely when the resolved values do not actually change. Should I clarify the comment i guess can be misleading?

src/game/g_client.c
Aciz: But `setspawnpt` cannot be executed until a client has finished connecting (`ent->client` becomes a valid pointer only in `ClientBegin`):

<https://github.com/etlegacy/etlegacy/blob/b05dffd61d0c41d415501bf49a18b872551165d7/src/game/g_cmds.c#L5343-L5351>

Which would mean that `spawnInfoChangePending` is always `qfalse` up until client has finished connecting, hence this is dead code. Or am I missing something here?

src/game/g_team.c
Aciz: Yeah maybe reword this, the comment kinda reads as doing some sort of "delta update" on configstrings, which isn't a thing.

src/game/g_client.c
RngesusSolutions: setspawnpt isn the only place this flag can get set. g_updatespawnpointstateplayercounts sets it too and thats actually the main thing this PR is changing. That runs from ClientSpawn() on every respawn, from the setautospawn script action, and also when spawn point state changes.

The reachable case here is a map change / restart. clientconnect --> g_readsessiondata restores sessionteam while the player is still loading, so they are still con_connecting then calculateranks puts everyone with pers.connected != con_disconnecting into sortedclients.

Since sess.resolvedspawnpointindex doesnt persist across map changes the first playercount pass can recomput it see that its different, and flag a client thats still connecting. That pass can happen from something as simple as an alreadyloaded player spawning.

g_getspawnforclient only filters by team not connection state so the stillconnecting client can still hit this path unless im wrong somewhere i dont see?

Aranud:
Since this https://github.com/etlegacy/etlegacy/pull/3488 remove from CS spawn points selections, changes in this PR are no longers required.

PR #3486, opened 2026-07-12
client: Fix cl_shownet printing wrong entity num

WofWca:
Cherry-picked from
https://github.com/ec-/Quake3e/commit/76da681398c030bfe3db5fb76411a07b0b20ab9b.

I did not test this on ETLegacy specifically,
only on Quake3e itself.

The license is compatible (GPLv2 or later).


rmarquis:
Thanks. Checked in #3602, closing here.

Issue #2066, opened 2022-07-20
Add spawnpoint number / latched spawnpoint number information to fireteam


BystryPL:
Same as class and latched class information, spawnpoint number and latched spawnpoint number could be added to fireteam. Quite useful information in comp.

PR #3488, opened 2026-07-14
mod: add ET_EBS_FIRETEAM entity, refs #2066

ryzyk-krzysiek:
- entity contains player's health, weapon, major and minor spawn points
- fix shoutcast entity writing player information from previous frame
- fix shoutcast entity unnecessary snapshot callbacks
- remove major and minor spawn points from CS_PLAYERS configstring
- changes should be backward compatible

refs #2066


src/game/g_main.c
```diff
@@ -4663,6 +4702,8 @@ void G_RunFrame(int levelTime)
 	G_LuaHook_RunFrame(levelTime);
 #endif
 
+	G_EBS_RunEntities();
```
Aciz, line 4705: Should this be before the LUA hook? If an update happens on the EBS entities on the current frame, and LUA wants to do something with them, there would be a 1 frame delay on the state of the EBS entities?

ryzyk-krzysiek: Yes, but my guess was that more often than not lua would be modifying player information than those entities and then it gets 1 frame delayed too.

Maybe I shouldn't worry and just expect people to modify EBS entities if they change player information and Im still not 100% sure, but current solution felt simpler if people don't have to care about them.

Aciz: Right, yeah probably better this way then.

PR #3490, opened 2026-07-16
dont trigger additional minimize for macos on focus lost event

MijPeter:
When I alt-tab etlegacy on my mac it's super clunky and etlegacy appears in recently opened apps.

I've made some research and I think that this is because etlegacy manually executes minimize inside custom method IN_WindowFocusLost() which fights with how sdl resolves it for macos.
Alt-tabing in macos is focus lost event from what I tested.

I've tested this solution on my PC but tbh I'm not sure if it's really correct.

### DIFF
--- a/src/sdl/sdl_input.c
+++ b/src/sdl/sdl_input.c
@@ -1718,7 +1718,9 @@ static void IN_ProcessEvents(void)
 				{
 					break;
 				}
+#ifndef __APPLE__
 				IN_WindowFocusLost();
+#endif
 			// fall through
 			case SDL_WINDOWEVENT_LEAVE:
 				Key_ClearStates();


mittermichal:
I would suggest to also test this with fullscreen, windowed and fullscreen windowed setting

MijPeter:
Thx, I just tested locally:

1. Fullscreen:
That is the setting that I was using and the setting for which alttabbing/reopening etlegacy is super clunky.
After the 'fix' it seems to work correctly.

2. Windowed
Seems to be working correctly on master, my change doesnt seem to influence it.

3. Borderless Fullscreen
Same as 2.

MijPeter:
Only change now is that in order to mute in the background user has to tick Mute when unfocused insted of Mute when minimized.

PR #3494, opened 2026-08-01
Fix hit detection: antilag rewind time and dropped usercmds under antiwarp

### LINKED ISSUE #1408: Issues with hit detection

rmarquis:
Fixes #1408

Hit registration felt delayed, and hits sometimes didn't register at all. Both symptoms trace to the interaction
between zinx's antiwarp port (et-antiwarp.c) and antilag (g_antilag.c), and both are compounded by antiwarp
silently discarding usercmds. Two commits, three distinct root causes:

1. Antiwarp fabricated serverTime fed the antilag rewind (d3147ec02)

DoClientThinks() enforces its movement budget by capping/splitting usercmds, rewriting cmd->serverTime to a
fabricated, earlier time (deltahax) before ClientThink_cmd() latches it into pers.cmd and fires weapon events.
Antilag then rewound all other clients to pers.cmd.serverTime — the fabricated time, always older than what the
shooter actually saw. Targets were tested against a past further back than the shooter's view, so hits missed or
registered "late". Under sustained antiwarping, ps.commandTime effectively freezes while level.time advances,
pushing the rewind time outside the marker window entirely.

Fix: latch the client-authored serverTime into gclient_s.realCmdServerTime before antiwarp can rewrite it, and use
it as the rewind time in G_HistoricalTrace/G_HistoricalTraceBegin. (Note: 72bed4f16 attempted this ETPub-style in
2020 but snapshotted after the rewrite, capturing the fabricated time too; reverted in b9d645fd7. This captures it
before antiwarp touches it. The whole lineage — stock WET SDK antilag, rtcwPro — shares the flaw via
pers.cmd.serverTime.)

2. Low-ping shooters got no compensation at all (d3147ec02)

G_AdjustSingleClientPosition() bailed with return qfalse ("oops, no valid stored markers") whenever the rewind time
was newer than the newest marker — the normal case for any shooter with ping below one server frame. Their shots
traced against target positions up to a full server frame (~50 ms at sv_fps 20) ahead of what they saw, with zero
backward reconciliation. i == j after the marker walk can only mean "newer than newest marker" with i == j ==
topMarker, so falling through to the existing copy branch now clamps to the newest marker; backup/restore symmetry
is preserved.

3. Antiwarp silently swallowed button presses (b9ad853ec)

DoClientThinks() discards whole usercmds in four places (>800 ms backlog drop, "future" packets, "past" packets
after a resync jump, queue overflow overwrite), losing their buttons/wbuttons along with the movement. Weapon fire
is level-triggered on cmd.buttons & BUTTON_ATTACK inside Pmove — if the entire press window is dropped, the shot
never exists server-side.

Fix: accumulate the buttons of every discarded cmd into droppedButtons/droppedWButtons and OR-merge them into the
next cmd actually processed. A swallowed click becomes a late shot instead of no shot — and thanks to fix 1, the
late shot still rewinds to what the player saw. Movement/drop semantics and the clock-cheat protection are
unchanged; only button state is forwarded, never movement.

Testing

- [x] Compile
- [ ] Playtest with a high-ping / packet-lossy client: semi-auto spam during loss bursts, g_antiwarp 16|32 debug
output, g_debugBullets/g_debugPlayerHitboxes to eyeball rewound hitboxes
- [ ] A/B against g_antiwarp 0 to confirm no regression for non-warping players
- [ ] Low-ping / LAN match to verify compensation for sub-frame pings

Known limitations

- Timing analysis of the merged press vs. stock semantics (PM_Weapon, including the semi-auto quickfire clamp at weaponTime <= 150 ms) shows the merge reproduces stock fire/no-fire outcomes in virtually all press/release timings. The one divergent artifact is a possible "ghost shot": a click released just before the quickfire window, fully dropped, can fire when the merged press lands inside the window — stock would not have fired, and the client never predicted the shot. Rare and benign; suppressing it (forwarding press/release edges instead of button levels) would reintroduce the lost clicks this fix exists to prevent.
- Shots from within a fully-dropped full-auto hold are inherently unrecoverable; the merge restores one shot rather than the full burst.
- Weapon switches (cmd.weapon) of dropped cmds are still not forwarded.
- Original etpro's reconciliation of antiwarp/antilag could not be checked: etpro source is not publicly available (closed source; antiwarp was a private handoff from zinx, a2139bb15).

### DIFF
--- a/src/game/et-antiwarp.c
+++ b/src/game/et-antiwarp.c
@@ -193,6 +193,11 @@ void DoClientThinks(gentity_t *ent)
 		serverTime = cmd->serverTime;
 		totalDelta = latestTime - cmd->serverTime;
 
+		// remember the command time the client actually authored - antilag
+		// must rewind to this time, since cmd->serverTime may be rewritten
+		// below (deltahax) to enforce the movement budget
+		ent->client->realCmdServerTime = cmd->serverTime;
+
 		if (pmove_fixed.integer || ent->client->pers.pmoveFixed)
 		{
 			serverTime = ((serverTime + pmove_msec.integer - 1) / pmove_msec.integer) * pmove_msec.integer;
--- a/src/game/et-antiwarp.c
+++ b/src/game/et-antiwarp.c
@@ -57,18 +57,19 @@ void etpro_AddUsercmd(int clientNum, usercmd_t *cmd)
 	gentity_t *ent = g_entities + clientNum;
 	int       idx  = (ent->client->cmdhead + ent->client->cmdcount) % LAG_MAX_COMMANDS;
 
-	ent->client->cmds[idx] = *cmd;
-
 	if (ent->client->cmdcount < LAG_MAX_COMMANDS)
 	{
+		ent->client->cmds[idx] = *cmd;
 		ent->client->cmdcount++;
 	}
 	else
 	{
 		// queue full, the oldest command is overwritten - rescue its button
-		// presses so they are not silently lost (refs #1408)
+		// presses before overwriting so they are not silently lost (refs #1408)
+		// (note: with a full queue, idx == cmdhead)
 		ent->client->droppedButtons  |= ent->client->cmds[ent->client->cmdhead].buttons;
 		ent->client->droppedWButtons |= ent->client->cmds[ent->client->cmdhead].wbuttons;
+		ent->client->cmds[idx]        = *cmd;
 		ent->client->cmdhead          = (ent->client->cmdhead + 1) % LAG_MAX_COMMANDS;
 	}
 }
--- a/src/game/et-antiwarp.c
+++ b/src/game/et-antiwarp.c
@@ -65,7 +65,11 @@ void etpro_AddUsercmd(int clientNum, usercmd_t *cmd)
 	}
 	else
 	{
-		ent->client->cmdhead = (ent->client->cmdhead + 1) % LAG_MAX_COMMANDS;
+		// queue full, the oldest command is overwritten - rescue its button
+		// presses so they are not silently lost (refs #1408)
+		ent->client->droppedButtons  |= ent->client->cmds[ent->client->cmdhead].buttons;
+		ent->client->droppedWButtons |= ent->client->cmds[ent->client->cmdhead].wbuttons;
+		ent->client->cmdhead          = (ent->client->cmdhead + 1) % LAG_MAX_COMMANDS;
 	}
 }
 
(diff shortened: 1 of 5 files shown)


x0rnn:
Maybe @ryzyk-krzysiek could verify this?

ryzyk-krzysiek:
I've looked into for a bit and here are my thoughts:

1. This sounds correct. I've tried testing it but triggering deltahax and shot at the correct time it's not exactly easy to do manually. Since it seems to be pretty hard for me to reproduce the issue and when I do it's a bit impossible to tell a difference and at that point game is kind of in unplayable state it seems right now it doesn't change much, but again my tests were not good enough. 
I can confirm there can be discrepancy between antilag rewind times, but since I was unable to do testing I would be satisfied with I won't vouch for it at this time, but the idea sounds correct to me (I also don't believe it would break anything but I rather be safe than sorry especially since I don't play the game anymore for few years and I myself refrained from some antilag optimizations because of it).

ETPro uses attack timestamp like in the mentioned commit https://github.com/etlegacy/etlegacy/commit/72bed4f16c3bb05b48b1a839707a574f9aad7a6f 

<details>
<summary>decomp (CLICK)</summary>
etpro_antiwarp_antilag_1
etpro_antiwarp_antilag_2
</details>

2. This seems incorrect to me. I don't think there should ever be a situation where the requested time is newer than newest marker, furthermore there should always be 2 snapshots between the requested time unless the time is too old and we run out of history. 
This is how the engine works, client always interpolates between 2 states, and even if it doesn't have them, then it doesn't mean server doesn't, only that it didn't receive it on time. This should be true even on ping 0, client is always in the past. I think `if (i == j)` should never be true unless we run out of history because of ping, so it defaults to current game state in that case.

3. I think this is mostly design issue not a bug. Should the buttons be allowed even though everything else was discarded?

Either way I have some thoughts about it.

- https://github.com/etlegacy/etlegacy/commit/b9ad853ece9dd3e77c3241e8a806108fa246fa9b#diff-66e06036bc03740f86ded48ac6a8f0d68571ca094f0c38d8fbc23f9533c84aa3R218-R222 This `totalDelta >= drop_threshold` drop, I think it could become inconsistent with antilag, we save buttons for a command older than 800ms, but do we have antilag history for more than 800ms? if not why are we trying to save the buttons and run them if we can't guarantee that the resulting shot is leading to be anything accurate that actually happened

- https://github.com/etlegacy/etlegacy/commit/b9ad853ece9dd3e77c3241e8a806108fa246fa9b#diff-66e06036bc03740f86ded48ac6a8f0d68571ca094f0c38d8fbc23f9533c84aa3R230-R231 This looks like saving buttons from a command that is not in correct state - from the future.

- https://github.com/etlegacy/etlegacy/commit/b9ad853ece9dd3e77c3241e8a806108fa246fa9b#diff-66e06036bc03740f86ded48ac6a8f0d68571ca094f0c38d8fbc23f9533c84aa3R239-R241 This feels like it shouldn't even be possible to happen since server won't run commands coming from the past already, maybe some incorrect order issue? Not sure.

Also maybe the added save for the buttons for some dubious user commands could lead to some "exploits" or just other unwanted/undefined behaviour, since they might be either old/out of order/mangled.

rmarquis:
Thanks for your thoughts. The main commit indeed aligns with our initial findings from a few years ago. The two other commits are more side issues that have been caught but I can't say I understand the implications.

> since I was unable to do testing I would be satisfied with I won't vouch for it at this time

Very understandable. I think a series of A/B testing would be required here, similarly to how we tested it a few years ago (see #1408, this allowed us to pinpoint the issue to the antiwrap in the mod code back then): one test server running etlded + Legacy (current version), one with etlded + legacy (with the above patch), and one last with etleded + Jaymod for control. Doing an eyeball comparison over a couple of evenings might be enough to validate the patch improves the situation.

PR #3499, opened 2026-08-08
Remove Prestige system

### LINKED ISSUE #3498: Remove Prestige system
The prestige system (#1155) was added as a way to encourage voluntary XP resets once all skill levels are reached, as an alternative to XP Save. In practice it never worked as intended: it offers no meaningful reward, is easily clicked by accident, and is perceived as a penalty because it resets skills back to 0. It has remained half-baked and is now considered unused leftover code rather than a genuine feature.

Remove the prestige system entirely in preparation for a better skill/upgrade system in the future.

## Scope

* Remove prestige tracking, storage, and UI display
* Remove prestige-related database fields and userinfo keys
* Remove prestige popups, collect button, debriefing notes, scoreboard icons, crosshair info option, and "most prestigious" award
* Remove prestige streak counter and related cvars/logic
* Clean up any prestige-specific Lua hooks or conditional checks

## References

* Original feature request: #1155
* Related commits include f67a9c2, 69e25cb, 8f0e07e, b8dea36, 10ada5b, and follow-up fixes

rmarquis:


src/cgame/cg_hud_io.c
Aranud, line 1: This break the version upgrade and shall be treated in new HUD  version by shifting the dynamic color to previous previous prestige value.

src/cgame/cg_local.h
```diff
@@ -2870,8 +2843,7 @@ enum
 {
 	CROSSHAIR_BAR_CLASS         = BIT(0),
 	CROSSHAIR_BAR_RANK          = BIT(1),
-	CROSSHAIR_BAR_PRESTIGE      = BIT(2),
```
Aranud, line 2873: By shifting the dynmacic color to prestige value we shall ensure the value selected by the player will be compatible once the prestige field is removed. A new HUD version bump is required.

Aranud (changes requested):

Aranud:
HUD version upgrade shall be handle according the prestige field removed for crosshair bar. 
Otherwise, other removed part are legitimates

Aranud (approved):

PR #3544, opened 2026-08-24
Add .wav for check-changes.py

Vorschreibung:


misc/check-changes.py
```diff
@@ -148,6 +149,75 @@ def check_jpeg(path: Path, errors: List[Error]):
...
+            ) = struct.unpack("<HHIIHH", fmt[:16])
+
+            if audio_format != 1:
+                errors.append(Error(path, "WAV file must be uncompressed PCM."))
+            if bits_per_sample != 16:
+                errors.append(Error(path, "WAV file must be 16-bit."))
+            if channels != 1:
+                errors.append(Error(path, "WAV file must be mono."))
+            if sample_rate != 22050:
+                errors.append(Error(path, "WAV file must be 22050 Hz."))
```
Aciz: Why? 44KHz is supported even by the vanilla engine.

Vorschreibung: @Aciz thanks!

Aciz: I mean, this is wrong too, 11KHz is also valid :D technically, *any* sample rate is valid, as the sound backend does all the resampling required to play the sound, so I don't think having a sample rate check here makes any sense at all.

Vorschreibung: Good point! UwU :3 etc.

PR #3576, opened 2026-09-13
client: scan 8 ports for LAN server list

Helldembez:
Increased the port scan range to 8 specifically for upcoming LAN. Was considering something more complicated, but this should do just fine for the amount of times it actually gets used.

Reason: When running more than 4 gameservers locally on the same server, the local server browser of clients would only show the first 4 servers in port range 27960-27963.


rafal1137:
I remember servers having `27969`
So why not increase it more and yes its documented 

obraz

Soruce:
https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

Helldembez:
Theoretically you can set any port hosting a server. So one can host at 27980 if they wanted to. The old engine just used its default port and added 4 to scan. I don't really care if its 8 or 10. The number 8 was chosen because that is what we realistically will host for now. Unless someone has ran into an issue here trying to host more than 8 servers I don't really see a reason to unnecessarily scan more ports than needed.

PR #3577, opened 2026-09-13
server: re-associate client session when its source IP changes

mittermichal:
`SV_PacketEvent` matches an incoming packet to a client slot with `NET_CompareBaseAdr`, which compares the IP only. A changed NAT **port** is already repaired ("fixing up a translated port"), but a changed **IP** is not: the packet matches no client and is silently discarded. The server keeps sending to the stale address and never receives another ack, so the unacknowledged reliable command backlog grows until it exceeds `MAX_MSGLEN` and `SV_CheckForMsgOverflow` drops the player with "Msg overflowed".

This affects clients behind CGNAT, on mobile links, or on a VPN that drops.

### Evidence

Scanning ~6000 client demos for the `cpm "<name>^7 Msg overflowed"` broadcast, cross-referenced with 31 server console logs: 49 real overflow events, 37 with a comparable before/after IP, of which 11 changed IP. In one case a player was dropped four times in a single gather night, his source port rebinding once and then his IP changing outright. Lining his own POV demos up against the server log, each drop follows 1-6 s after the last snapshot he received.

Note that IP change explains a minority of these drops - the rest only need a few seconds of ordinary packet loss at the reliable command rates gathers run (`g_teamInfoUpdateRate` is 20 on most of the logs I have, against a default of 1000). This PR addresses the address change subset, not overflow drops in general.

Incidentally, a dropped player's own demo never records the reason: `CL_WriteDemoMessage` runs after `CL_ParseServerMessage`, and `disconnect` longjmps out via `Com_Error(ERR_SERVERDISCONNECT)`, so the final message is never written, while `CL_StopRecord_f` still writes the end marker. The drop is only ever visible from another player's POV.

### The change

When no client matches on address, a second pass looks the client up by `qport` and moves the session to the new address. To limit abuse the candidate must be `CS_CONNECTED` or later, must not be a bot or democlient, must send a sequence that would pass the netchan ordering check, and may only be moved once per `SV_ADDRESS_CHANGE_INTERVAL` (5 s). Gated on `sv_allowClientIpChange`.

### This is not authentication

A spoofer needs a matching `qport` and a sequence ahead of the client's current one, both of which travel in the clear. It extends the trust already granted to port changes so that it also covers address changes. A properly safe version would require proof of the challenge, which is a protocol change.

I have defaulted the cvar to `1`, which is a guess - `0` may be the right call until that is settled. I would rather have a maintainer's view on this than have it found in review.

### Testing

Local server, 12 omni-bots, `g_teamInfoUpdateRate 20`, client changing address mid-game via `net_ip` + `net_restart`:

| scenario | result |
| --- | --- |
| `net_ip` change, `sv_allowClientIpChange 0` | `msg overflowed` drop 9.7 s after the change |
| `net_ip` change, `sv_allowClientIpChange 1` | session continues, address change logged |
| `net_port` change | unchanged, still "fixing up a translated port" |

Test server binaries built from this branch, for admins who want to try it before a merge: https://github.com/mittermichal/etlegacy/releases/tag/v2.85.0-client-ip-change1

Discussed on Discord: https://discord.com/channels/260750790203932672/1537752538663878696


mittermichal:
i should have told it to run the `uncrustify` :sweat_smile:

mittermichal:
this was after a quite long session investigating that issue and for me the focus is the competitive servers (they have modified g_teamInfoUpdateRate to 100 atm) for which I made build available here:  https://github.com/mittermichal/etlegacy/releases/tag/v2.85.0-client-ip-change1
That was priority for me and I just let claude also write PR, so feel free to close this and use it as reference. But I confirmed manually that it fixed changing `net_port` and then doing `net_restart`

Issue #1147, opened 2020-03-05
Add weapon toss sound (bouncing)


ETLegacyTeam:
Author: @IR4T4 (IR4T4)
Date: 2018-12-25
Redmine Issue: 1147, https://dev.etlegacy.com/issues/1147

---

I can't hear the weapon toss sound anymore  when body is switching to corpse. 

---

Related commit(s):
c4c1794d66db63063204e519d5e93da1fdb82df5 - game: add bouncing effect on items drop (weapon/ob..

ETLegacyTeam:
From: @Aranud (Aranud)
Date: 2018-12-30T19:02:12Z

---

Not present in vanilla, this one can be add.

ETLegacyTeam:
From: @Aranud (Aranud)
Date: 2019-09-03T17:07:49Z

---

We need a generic bouncing sound for all weapon. NQ have these for both clip and weapon by example.

rmarquis:
Two free sound (Creative Common) that might be good enough to fit:
* [GUN DROP](https://freesound.org/people/MTJohnson/sounds/426313/)
* [Dropping Sword](https://freesound.org/people/Christopherderp/sounds/342452/)

The .weap files need to be extended with a new `bounceSound` parameter, which can then be linked to a new sound event.

PR #3603, opened 2026-09-21
mod: add weapon toss sound (bouncing), refs #1147

Aranud:


Aranud:
Feedback : 

> I tried those sounds. Two things I would adjust a little would be:
Ammo/Medic pack pitch lower
MP40/Thompson (it sounds a bit like a toy currently, at least the end of the bounce sound)
The rest were fine in my opinion.

>Wouldn't that reveal position behind corners

>I'd say the sounds feel too "light". At least for SMG and rifle, they make me think more about ammunition boxes than weapons.
But idea is good in principle, though I'd probably also reduce slightly teh sound volume as to not make it too spammy.

x0rnn:
By the way, this would play both sounds (pickup and drop) at the same time?

Edit: ah, this is meant when body switches to cropse/tapout...

